PR #15, opened 2017-03-15
changed cn to uid on userDN lookup

jkaberg:
this patch changes cn to uid. please test this against openldap upon merge.


clems4ever:
Some unit tests are red but I'm ok with the change if it is a quick win for FreeIPA, it should do the job for now. Though, I'll probably customize the search at some point using the config file.

jkaberg:
@clems4ever yeah, I'm unsure if OpenLDAP can use the UID but if it's an option in the config then that would work out just fine 😄

clems4ever:
Pull request #18 is a long-term solution for what you tried to do.

Issue #14, opened 2017-03-14
No redirect, verfication ok


jkaberg:
Somehow, this doesn't work:
```
Parse configuration file: /etc/auth-server/config.yml
{ port: '80',
  ldap_url: 'ldap://ipa.xxxxx.local',
  ldap_users_dn: 'cn=users,cn=accounts,dc=xxxxx,dc=local',
  ldap_user: 'uid=xxxxx,cn=sysaccounts,cn=etc,dc=xxxxx,dc=local',
  ldap_password: 'xxxxx',
  session_secret: 'xxxxx',
  session_max_age: xxxxx,
  store_directory: '/var/lib/auth-server/store',
  logs_level: 'debug',
  notifier: { filesystem: { filename: '/var/lib/auth-server/notifications/notification.txt' } } }
Listening on 80...
info: 1st factor: Starting authentication of user "joel"
debug: 1st factor: Start bind operation against LDAP
debug: 1st factor: username=joel
debug: 1st factor: base_dn=cn=users,cn=accounts,dc=xxxxx,dc=local
info: 1st factor: LDAP binding successful
debug: 1st factor: Retrieve email from LDAP
debug: 1st factor: document={"dn":"uid=joel,cn=users,cn=accounts,dc=xxxxx,dc=local","controls":[],"mail":"joel@xxxxx.local"}
debug: 1st factor: Retrieved email is joel@xxxxx.local
info: POST 2ndfactor totp: Initiate TOTP validation for user joel
debug: POST 2ndfactor totp: Fetching secret for user joel
debug: POST 2ndfactor totp: TOTP secret is {"userid":"xxxxx","secret":{"ascii":"s!yoi0)}kooTrJ>!4qsmmFs4kF,/LK,I","hex":"xxxxx","base32":"xxxxx","otpauth_url":"otpauth://totp/SecretKey?secret=xxxxx"},"_id":"xxxxx"}
debug: POST 2ndfactor totp: TOTP validation succeeded
```

Virtual host sbox:
```
server {
	listen 192.168.1.11:443 ssl http2;

	server_name sbox.xxxx.local;

	access_log /var/log/nginx/xxxx.local/sbox.xxxx.local-access.log;
	error_log /var/log/nginx/xxxx.local/sbox.xxxx.local-error.log;

	auth_request /authentication/verify;

	auth_request_set $user $upstream_http_x_remote_user;
	proxy_set_header X-Forwarded-User $user;
[...]

PR #22, opened 2017-04-09
domain used for cookie - an issue in the example

pccasto:
Fortunately the discussion in issue #14 helped me figure this out.  Eventually... :-).

Could not get the buttons to register for TOTP or U2F.  Kept getting a message that "the email could not be sent" - which I thought indicated an issue with the file-based notification. 
 
Could get the notification.txt when doing a password change, so dug deeper.  Added and ran tcpdump on the authelia_auth instance, and was seeing the 403 error for all calls - and noticed the domain for the cookie...

Ran into two different issues while verifying this - I think credential may be cached so that log-out doesn't work, and saw some occasional odd connection problems with the ldap server.  But tested enough to determine that those do not seem related to this change.

Thanks, and Best Regards,
Paul

### DIFF
--- a/config.template.yml
+++ b/config.template.yml
@@ -72,7 +72,7 @@ access_control:
 session:
   secret: unsecure_secret
   expiration: 3600000
-  domain: example.com
+  domain: test.local
 
 
 # The directory where the DB files will be saved


clems4ever:
Hello @pccasto , thank you for the fix. Indeed, it was a mistake.

For the 2 issues you reported:
1/ Logout procedure should invalidate the session by modifiying an attribute in the session. Then, during next access to the secret the check procedure should return a 401 error code. I will test it soon.
2/ I also figured some connection problems with the LDAP. This is due to a bad implementation of the reconnection at the level of the ldapjs npm module. Actually, when connection is not used for a long time it is closed and the client goes into 'idle' state. On next query, the client should start a new connection. There is a flag that if set to true should automatically start a new connection but it works only after the second request... The first one fails...

Thank you very much again for the fix.
Clément.

pccasto:
Hi Clément,

Thanks for the project.  Glad to be able to help.

Best Regards,
Paul

PR #47, opened 2017-07-13
Use Redis as a store of user sessions for resilience

clems4ever:
Authenticated users now have their session stored in Redis in order to make the service resilient. So far, sessions were stored in memory, therefore, if one restarted the service all users were all logged off.
With Redis, user sessions can now be distributed and replicated for resilience.

Other stores can be easily implemented since everything depend on express-session.


VeloAddict:
How would one store the session in an sqlite database?

PR #64, opened 2017-08-08
Don't build Docker container from source for the example

FrozenDragoon:
This splits the example scripts into two sections:

* ```example-commit``` is the same functionality as the current branch, using npm to build from source every time.  This would allow testing of branches that haven't yet been merged, etc.
* ```example-dockerhub``` instead pulls the image from [Docker Hub](https://hub.docker.com/r/clems4ever/authelia/).  This removes the requirement for dependencies like npm, which are only used in the building process anyway, as the app is actually run in Docker.  

I figured, we already have a prebuilt container, why not use it?

This also allows the example environment to run on  Windows machines.  Currently npm throws an incompatible OS error when attempting to run the provided commands.


clems4ever:
Hello @FrozenDragoon , this split is indeed a good idea.

README.md
```diff
@@ -104,7 +104,7 @@ Add the following lines to your **/etc/hosts** to alias multiple subdomains so t
     
 Deploy the **Authelia** example with one of the following commands:
 
-Build Docker from current commit:
+Build Docker conatiner from current commit:
```
clems4ever: misspelled :)

README.md
```diff
@@ -102,12 +102,18 @@ Add the following lines to your **/etc/hosts** to alias multiple subdomains so t
 
 ### Run it!
     
-Deploy **Authelia** example with the following command:
+Deploy the **Authelia** example with one of the following commands:
```
clems4ever: I like it!

clems4ever:
You need to make the scripts executable to pass the checks.

clems4ever:
I think I will also add a test on AppVeyor for testing the example on Windows. It would also allow testing of Authelia against a MS Active Directory as a LDAP server at some point.

FrozenDragoon:
Fixed the permission issue that was causing the test to fail. 

I realize now I should have done a pull request against a dev branch, not master, sorry.

Still a novice with almost everything being used here, but I'm glad you like this suggestion.

clems4ever:
Can you please squash every CL except 0654178bd865fae2357aad264f32ee23e619ff9b so that you end up with only 2 CLs?

clems4ever:
Hello @FrozenDragoon, Good job! I'm writing just to let you know that some comments have not been taken into account before I can merge.
Can you also please squash the commit messages and rename the message from "PR #64" to something related to the purpose of the dev. I have cloned your branch and wrote an example commit message you can use if you wish.

FrozenDragoon:
@clems4ever I appologize for the delay.  I have pushed the changed commit message.  

It makes sense to squash for clarity, it was just not something I had done before.  Thank you for your example.

clems4ever:
Hello @FrozenDragoon, no problem since I'm in vacation right now. I will definitely have a look and accept when I'm back home at the beginning of September. Is that OK?

clems4ever:
Hello @FrozenDragoon, some checks have failed. It's due to new release version of typescript. Either you can try to fix the issue with the new version or set it exactly equal to 2.3.2 (as it was previously).

FrozenDragoon:
So that's why it failed. I saw that and tried to track it down, but couldn't.  I will try to get it updated today.

clems4ever:
Hello @FrozenDragoon, I fixed the compilation issue in mainstream. Please rebase.

FrozenDragoon:
Rebased and updated.  Latest test failed, I believe due to something similar as #76 .

You added new tests in the latest version, I've updated them with the new location of ```/scripts/example/dc-example.sh``` (moved to ```/scripts/example-commit/dc-example.sh```).  I squashed that commit into the main one.

I might look at moving it back to the original location so that the tests don't need to be updated and any new tests should not fail either.

PR #174, opened 2017-10-23
Fix npm package versions with package-lock.json

clems4ever:


khimaros:
For reproduction, these are clean-room builds inside of a debian/stretch container with nodejs 8.7.0 installed. Which nodejs version are you using?

clems4ever:
On my workstation
```
npm --version
5.4.0
node --version
v7.10.1
```

On Travis
```
$ node --version
v7.10.1
$ npm --version
4.2.0
```
And both work.

khimaros:
Your package-lock.json causes a failure earlier in the npm install process (chromedriver). master actually makes it further.

It seems the error occurs only with `--only=dev` (as recommended in the documentation), presumably because bluebird is not listed as a devDependency.

clems4ever:
Hmmm, this is interesting since the --only=dev is supposed to be tested in Travis. Unfortunately it uses previously downloaded packages from `npm install` command. This is clearly a bug. I will fix it soon.

khimaros:
Thank you! Either way, I would recommend rolling back the package-lock.json change, as I suspect this will have very unexpected behavior.

clems4ever:
Thank you for troubleshooting. I think package-lock.json should be in source code anyway. It did not impact any of the tests in Travis so I guess it's ok not to revert it. I prefer correcting unexpected behaviors if they appear.

PR #193, opened 2017-11-01
Add support of basic auth for single-factor protected endpoints

clems4ever:


jmccoy555:
Hi @clems4ever (looks like you've been busy)

Will hopefully give this a go shortly. Did I see there was a breaking change to Authelia's verify endpoint?? Is that now in master?

Will let you know how I get on with a few apps after some testing!!

Thanks

clems4ever:
Hello @jmccoy555 , a bit busy yes :).

The feature is cool actually, it's a first step toward more complex mechanisms like app passwords and tokens (probably leveraging OAuth2).

Yes, there is a breaking change. I will add a changelog in the readme for the next release. I had to move the */verify* endpoint to */api/verify*. It requires a little change in your nginx configuration.

Alright, I'm waiting for your feedback then! Thank you.

jmccoy555:
Hi @clems4ever 

Well the update was easy (just need to see what other new goodies there are to play with). The openHAB app is working nicely, but Transdrone isn't so far, which is odd as it works with auth_ldap directly in Nginx. Getting message='AccessDeniedError: First factor not yet validated: userid is missing.

I've added the proxy_set_header  Proxy-Authorization $http_authorization; header which I presume is required, and everything else in Nginx is the same between the two. 

Will have to do some more investigating between the two, I will let you know if I find anything. 

I'll also give active-sync a go (just for fun), but that will take a bit more adjusting in Nginx. 

p.s. OAuth2 would be great, would save messing around with Amazon etc!

p.p.s. I did have a look to try and understand the code to help with the auth_http, maybe looking to see how you have done auth_basic will help me see.

clems4ever:
Hello @jmccoy555 , I had the exact same issue today. 
It's due to the fact that **Proxy-Authorization** is not set correctly for this app. You need to ensure **Proxy-Authorization** is correctly forwarded to **/api/verify**. Authelia is checking whether the header exist to decide the auth method to be used by the **/api/verify** endpoint (either session cookie or basic auth).
My problem was I sent **Authorization** header to a protected app but nginx did not transform it into **Proxy-Authorization** before forwarding to Authelia.
So, Authelia, fell back to session cookie method instead of basic auth, leading to the exact same error message you have. I should add more logs to make it clearer.

jmccoy555:
Hi @clems4ever 

Yes I did notice that there was no mention of authorization in Autherlia's header log, so guess the issue was something related to headers and Authelia was doing as you have described; but to be honest my eyes were tied and I didn't trust them!!

I have just check by manually setting the header 

proxy_set_header Proxy-Authorization "Basic USER_AND_PASS", where USER_AND_PASS = base64(user:pass)

and it is working as expected. 

So does this mean that this is not a standard header being received by Nginx from the app, or is there another reason it is not being transformed? Is there a way to see what headers Nginx is receiving, httpbin is not helpful in this case. 

Also, do you think it would be good to get a FAQ, or working config example section on git, or is something you have planned?

clems4ever:
Hello @jmccoy555 , 
thank you for the troubleshooting. There is something I don't understand yet: does your application require authentication (as well as Authelia)? I mean, do both proxy and app need to authenticate (potentially with the same credentials) in your setup?

It seems to me that non preemptive authentication is meant to tell the user how he can authenticate but the two first steps are not really mandatory. Maybe I'm wrong.

It would be good to have a FAQ somewhere indeed. A wiki page could be a good location and a link to it in the README. The working config section is also a good idea but I think any new config would need a complete working environment to test it (I'm thinking of docker for the sake of simplicity). BTW, I'm currently implementing your stack (portal, authelia, private) with docker-compose for the test environment. This is a good example because it is not straightforward regarding the nginx configuration. I have some issues right now.

To answer your question, there was nothing planned apart from the implementation of your stack.

jmccoy555:
Hi @clems4ever 

No, maybe I've confused things slightly. Previously I have been using auth_ldap in Nginx as Authelia didn't support auth_basic. I know this works and am therefore using it and referring to what I am seeing to understand what is and isn't happening when trying to just use Authelia.

From what I have quickly read today, non preemptive authentication required the server to return a request for the authorization header, step 1 and 2 before the app will supply this information in the authorization header as step 3.

So you see this in the log for every request (auth_ldap directlty in Nginx);

[02/Nov/2017:10:28:11 +0000] "GET / HTTP/1.1" 401 188 - "-"
[02/Nov/2017:10:28:11 +0000] "GET / HTTP/1.1" 200 761 - "Basic base64user:pass"

Preemptive authentication on the other hand bypassed these first two steps and sends the basic auth header and credentials in each request. 

So you just see this;

[02/Nov/2017:10:28:11 +0000] "GET / HTTP/1.1" 200 761 - "Basic base64user:pass"

So, what I think Authelia is doing is, when recieving a request from an app that uses preemptive authentication, i.e. straight in at step 3, and supplies the authorisation header, is going hey great this is auth_basic and off you go.

Whereas when an app is using non preemptive authentication (arguably more secure as credentials are not sent unless requested), the app just issues a request to the server, which is sent to Authelia with no authorization header (because there isn't one at this stage), which by design results in session cookie auth, so you never get past step 1.

So I think for this to work there would need to be another stage in the process i.e. step 2, where Autherlia goes, 401 if you like I can accept auth_basic if you send me the authorization header, this is the returned by the app and you are at step 3 with an authorization header, or if not Authelia assumes you just want the login page and redirects you to the log in page. 

It sounds logical to me, but then this is only the first time I've even discovered there are two different approaches. 


For the wiki / FAQ could you maybe do an initial commit with a template so we can start to populate? I've certainly learnt a lot and there's lots of good info in these issues that would be helpful for others I'm sure.  

For the working config I was think of maybe Nginx config examples, with anything you needed to do at the other end in comments. Maybe this is to much, but I think some specific examples world certainly help, or maybe the docker compose could use a backend app or something. Anyway, trying not to stray off topic too much!!

clems4ever:
Hello @jmccoy555, I created the wiki: https://github.com/clems4ever/authelia/wiki. You can start playing with it if you wish. Tell me if you need any permission to create PRs on it. I will grant you the rights.

Yes, we were on the same page for nginx config. But what I meant is that a config alone might not be very helpful because it often depends on a complex architecture which can be simulated easily with docker-compose. Any way, don't hesitate to propose things if you have an idea of how it could work.

Concerning Authorization, I got your point. It is now clear to me, thank you for the explanation. This very clear in this documentation by the way: https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication. Implementing a protocol blindly is never a good idea anyway... :).
There is still something I need to figure out. Does this mechanism mean we need to choose between basic auth and authentication from a browser?! 
Here is my rationale:  I want to get access to a protected resource with a browser, I send the request and the server sends back a 401 error (this is exactly the first step of basic auth btw). At this point, nginx catches the error and redirect to Authelia login page with a 30X, thus, as a client, I only see a 30X. Therefore, if I use basic auth for accessing this endpoint, I do not get the 401 but a 30X instead and I'm screwed, right?

clems4ever:
Can you please tell me more about your environment? Which applications are trying to talk to each other?

jmccoy555:
Hi @clems4ever 

OK, I'll have a look, and let you know.

Great.

Yes, the Mozilla doc is good. I guess this could be an issue (and may explain the usual ugly basic auth login box that you are normally presented with). I didn't think about the mechanism if access from a browser when I wrote my last comment, whops! Maybe this is a limitation for apps that use non preemptive authentication. I'm not sure if this would be a big issue or not. I guess there is not really a benefit in using Authelia (other than one less thing to configure) to do this from an app anyway and it may in fact be better to keep the load just on the Nginx server, rather than both. 

The only app that I've tried that behaves like this is Transdroid, and there is an Issue raised. 

https://github.com/erickok/transdroid/issues/312

https://github.com/erickok/transdroid/issues/105

PR #245, opened 2018-07-20
Make notifications appear on top instead of inline

naggie:
Currently notifications reflow the document which causes the interface
to jump twice which can be frustrating if you're trying to click
something.

This change makes the notification appear at the top of the form as
such:

image

@clems4ever do you agree with this? If not feel free to discard this PR.

### DIFF
--- a/client/src/css/01-main.css
+++ b/client/src/css/01-main.css
@@ -19,9 +19,10 @@ body {
 /* notifications */
 .notification {
   padding: 10px;
-  margin: 10px 0px;
+  margin: 15px 0px;
   border-radius: 6px;
   display: none;
+  position: absolute;
 }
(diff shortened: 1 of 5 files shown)


clems4ever:
I think those failing tests are not related to your change, I have the same failures on my PR and I don't yet understand why. I'll have a look.

clems4ever:
The code is ok for me, there is only a missing space between the links to register devices in second factor page. Please fix it and I will merge the PR.

clems4ever:
2018-07-26-004645_1920x1080_scrot

naggie:
Ah -- it seems that I removed the important whitespace when running a whitespace remover. Fixed -- thanks

PR #246, opened 2018-07-20
Make first factor login page UI indicate loading state

naggie:
I decided against a spinner to go for a more bootstrap-native approach:

* Submit button changes to "Please wait..."
* Form fades and disables
* Form reverts to normal on authentication failure, selecting the password instead of removing it

image


naggie:
It seems I'll need to update some tests. Will do.

clems4ever:
This feature is nice! This one indeed needs an update of the tests. :)

naggie:
The failing tests have been fixed, although I think we're experiencing the same CI issue that you mentioned in the other PR.

clems4ever:
Those errors seem a bit different from the ones in other PR but let me fix the issue first before doing anything else. I'm on it now.

clems4ever:
I fixed the issue with the tests in master but it seems this PR is still failing. It's probably due to selenium trying to click on the disabled button.

clems4ever:
The code is ok. I tested the feature the effect is really nice. Please fix the tests and you're good to go!

naggie:
It seems I get different, intermittent errors when running locally:
image
image

Looking at the code, there doesn't seem to be a reason why selenium can't find the username element -- even if it's disabled due to the selector used. Even so, the timeout is set such that selenium should retry giving the JS time to restore the element....

Different tests fail every time on this branch but not master. I am investigating, suspecting a race condition.

naggie:
I've fixed it -- the problem was that as the password is now selected, some tests sent keys in such a way that a bad password was appended to instead of replaced causing subsequent tests in the same instance to fail.

I fixed it by clearing the password every time in `loginWithUserPassword`

clems4ever:
Awesome! Thank you very much for the additional fix. I'm glad you contributed to the project!
I hope you will keep doing so as there are plenty of things to improve.

PR #255, opened 2018-08-19
Move wiki documentation to Authelia repo

clems4ever:


clems4ever:
I'll do files one by one soon.

PR #281, opened 2018-10-12
Changed minimal config to provide for a working installation + docker swarm example

p-rintz:
I changed the current minimal config to include all configuration snippets that are needed for a working config, since there were some things missing that needed to be taken from other configs.

I also added an example configuration that is working with docker swarm. A reverse proxy would still need to be configured, but I dont currently have a 100% working configuration with traefik internal tools, which is why I left out my current one.

### DIFF
--- a/config.minimal.yml
+++ b/config.minimal.yml
@@ -86,8 +86,8 @@ regulation:
 
 notifier:
   # For testing purpose, notifications can be sent in a file
-  ## filesystem:
-  ##   filename: /tmp/authelia/notification.txt
+   filesystem:
+     filename: /tmp/authelia/notification.txt
 
   # Use your email account to send the notifications. You can use an app password.
   # List of valid services can be found here: https://nodemailer.com/smtp/well-known/  
@@ -98,11 +98,11 @@ notifier:
   ##   service: gmail
   
   # Use a SMTP server for sending notifications
-  smtp:
-    username: test
-    password: password
-    secure: false
-    host: 'smtp'
-    port: 1025
-    sender: admin@example.com
+  #smtp:
+  #  username: test
+  #  password: password
+  #  secure: false
+  #  host: 'smtp'
+  #  port: 1025
+  #  sender: admin@example.com
 
--- a/config.minimal.yml
+++ b/config.minimal.yml
@@ -114,7 +114,7 @@ access_control:
       - domain: 'dev.example.com'
         policy: allow
         resources:
-- '^/users/bob/.*$'      
+          - '^/users/bob/.*$'      
 
 # Configuration of the authentication regulation mechanism.
 #
--- a/config.minimal.yml
+++ b/config.minimal.yml
@@ -5,7 +5,7 @@
 authentication_backend:
   file:
     # The path to the database file. The file is at the root of the repo.
-    path: /etc/authelia/users_database.yml
+    path: /etc/authelia/volume/users_database.yml
 
 session:
   # The secret to encrypt the session cookies with.
--- a/config.minimal.yml
+++ b/config.minimal.yml
@@ -23,14 +23,64 @@ storage:
   local:
     path: /etc/authelia/volume
 
+# TOTP Issuer Name
+#
+# This will be the issuer name displayed in Google Authenticator
+# See: https://github.com/google/google-authenticator/wiki/Key-Uri-Format for more info on issuer names
 totp:
   issuer: example.com
 
+# Authentication methods
+#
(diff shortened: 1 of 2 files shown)


docker-compose.swarm.minimal.yml
```diff
@@ -0,0 +1,47 @@
+version: '3.4'
```
clems4ever: Do you think it would be possible to test this config in Travis?

config.minimal.yml
```diff
@@ -14,3 +14,157 @@ session:
   # The domain to protect.
   # Note: Authelia must also be served by that domain.
   domain: example.com
+
```
clems4ever: I think the tests are broken due to this change and not reflected in the tests.

config.minimal.yml
```diff
@@ -14,3 +14,157 @@ session:
...
+# Configuration of the storage backend used to store data and secrets.
```
clems4ever: I would reduce the amount of comments in this configuration because it was made to be simple to read for users and with the least number of options to not make people confused.
I suggest we only leave the most useful comments and not list all possibilities for an option in the comments.

config.minimal.yml
```diff
@@ -5,7 +5,7 @@
 authentication_backend:
   file:
     # The path to the database file. The file is at the root of the repo.
-    path: /etc/authelia/users_database.yml
+    path: /etc/authelia/volume/users_database.yml
```
clems4ever: I think you can avoid that.

clems4ever: This configuration will be reused in a demo mode of Authelia (to be implemented) where users won't even need to have a nginx or docker installed on the machine. In that case 'volume' makes no sense.

docker-compose.swarm.minimal.yml
p-rintz: I dont really have any experience with Travis at all, so I dont really know, sorry.

Swarm is close to the same. The "deploy" part is not available in normal non-swarm docker. Docker configs also isnt available in normal docker, afaik. I have this config running on my swarm cluster, but I have no idea if/how travis can test it. Id have to look into it later.

clems4ever: Ok, so we should mention this configuration in the documentation but put a warning saying this configuration is not automatically tested.

config.minimal.yml
p-rintz: I kept them in, since I thought youd want it to be like the other config. But that was just my assumption. Ill look at it and see which I think should definitely stay and you can then tell me if you agree.

config.minimal.yml
p-rintz: I did it like this because in docker the volume that is mounted should preferably be in a subdirectory of /etc/authelia, because that way you only need one single docker volume instead of two. (for users_database.yml and storage each)

Its of course possible to do it with two volumes and without a subdirectory.

config.minimal.yml
p-rintz: Sorry, which change exactly do you mean here? The empty line?

docker-compose.swarm.minimal.yml
p-rintz: Yes, compose will throw the following:
WARNING: Some services (authelia) use the 'deploy' key, which will be ignored. Compose does not support 'deploy' configuration - use `docker stack deploy` to deploy to a swarm.
WARNING: Some services (authelia) use the 'configs' key, which will be ignored. Compose does not support 'configs' configuration - use `docker stack deploy` to deploy to a swarm.

Sadly, I dont cant find a linter for stack configs. (Portainer has a helpful linter built in, which is what I use)

config.minimal.yml
p-rintz: Currently travis is failing at the minimal docker config:

[ENVIRONMENT] Running: docker-compose -f docker-compose.minimal.yml -f example/compose/docker-compose.base.yml -f example/compose/nginx/minimal/docker-compose.yml down
  1 passing (25s)
  4 failing

  1) Fail TOTP challenge
       successfully login as john
         "before all" hook:
     Error: ENOENT: no such file or directory, open '/tmp/authelia/notification.txt'

p-rintz: Seems it was indeed the commented out filesystem notification that made the travis check fail.

docker-compose.swarm.minimal.yml
clems4ever: ok but anyway if we can use it with compose it would be a good idea at some point to test it. Not a blocker for this PR though.

PR #289, opened 2018-10-23
Introduce "bypass", "one_factor", "two_factor" and "deny" ACL rules

clems4ever:
In this PR we deprecate the `auth_methods` and allow ACL rules to be more granular. That way a user could be granted access with one or two factor depending on the resource he wants to access and not only the domain.

[DEPRECATION] `auth_methods` is deprecated after this PR.

[BREAKING] The format of ACLs is changing to handle more use cases.
The new format allows to specify the subject of the ACL that can be either anybody, a group or a specific user instead of putting ACLs into categories as before. Adding this attribute and replacing categories (any, groups, users) by a list of ACLs allows Authelia to treat ACLs as iptables does with rules which enables more expressiveness.
With this new formalism, the first matching rules is applied so make sure your rules are properly sorted.


clems4ever:
cc @nightah

PR #308, opened 2018-12-17
update url to nginx.conf

RHITNL:
corrected the url to nginx.conf. Missing a dot


RHITNL:
still wrong url. My bad :(

PR #319, opened 2019-01-27
Themes update

BankaiNoJutsu:
Major changes include:

- fix various css issues on themes
- added grunt-concurrent build mode, builds faster.
- added time-grunt to check build time.
- added beta install script, GUI and CLI. (still linking on my fork as of now)
   - is based on node 10 (crypt3 has deprecated warnings, but works)
- Update readme in themes/
- Various minor things.

Please test :)
Thanks


clems4ever:
Hello @BankaiNoJutsu , beware that I'm currently rewriting the UI in React to ease development and make Authelia compliant with material design rules using material-components-web.
I completely rewrote the frontend of the portal and I'm also introducing scripts to simplify the development workflow.

I would recommend splitting your review so that I can more easily see what is introduced

clems4ever:
I've just pushed my work in progress in branch react-ui so that you can have a look. This work will tremendously improve the development cycle both for the frontend and backend.

Regarding theming, with react and material-component-web we'll be able to use sass partials and variables to avoid code duplication.

BankaiNoJutsu:
Hello @clems4ever , cool thank you! just had a look, looks great! Just need to dig into React a bit, as it's all new to me. 
From what I understood when building with sass you get a compiled css from the scss, therefore only one theme can be made at a time? 
However possibly there are ways to make multiple themes, and using JS to change it dynamically on a separate page like /admin for example?

Concerning my changes on current master, maybe they make no sense anymore, as the client itself is now completely different. I will close this pull, if okay with you.

clems4ever:
Hello @BankaiNoJutsu ,
I think you don't really need to know React if you want to help with theming, Sass is sufficient I think. Yes, you're right, Sass is compiled into css and so this only allows one theme at a time. This is already the same as in master since we specify a given theme during the build.
Making theme change at runtime with JS is generally a bad idea because it usually breaks a security feature offered by browsers called Content-Security-Policy (https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP). Maybe generating several CSS files and switching using HTML could work instead of JS. We would need to test. But anyway, since the user cannot yet save a preferred theme (that would require it to be stored in DB), there is no value in switching in live.

What is good with Sass is that we will avoid a lot of code duplication for the theming. You can just put all the variables you want to change in a single file and generate a new variable file for each custom theme. It's like you have a templated skeleton that you fill with a single file containing all the customizable properties.

Indeed, you're current PR will hardly fit in the new UI I think. I was about to re-introduce the dark theme but if you want to help me with that you can check how the new UI is working while I'm fixing/rewriting the remaining tests.

Do you understand the new scripts I added to simplify the development workflow?
You can basically run `./scripts/authelia-scripts start config.minimal.yml` to bootstrap a hot reloadable dev environment that allows you to see your changes in (almost) real-time as you code.

I'm glad you like Authelia and you contribute to it! Don't hesitate to tell me if you need help with something. We can even discuss in Gitter if we need a more interactive conversation.

clems4ever:
This PR is not up to date anymore, I'm closing it.

PR #320, opened 2019-01-30
Fix dead link in docs/deployment-dev.md

kevynb:
Hello,

I only fixed a dead link in the deployment-dev file because it didn't work when I clicked on it :)

Cheers!


clems4ever:
Awesome! Thank you very much.

PR #344, opened 2019-03-27
Capture IP address and Target URL in Duo 2FA request

nightah:
This change provides additional details in the Duo 2FA request including:

- Client IP
- Target URL

duo-additions

### DIFF
--- a/server/src/lib/routes/secondfactor/duo-push/Post.ts
+++ b/server/src/lib/routes/secondfactor/duo-push/Post.ts
@@ -6,6 +6,8 @@ import * as UserMessage from "../../../../../../shared/UserMessages";
 import redirect from "../redirect";
 import { Level } from "../../../authentication/Level";
 import { DuoPushConfiguration } from "../../../configuration/schema/DuoPushConfiguration";
+import GetHeader from "../../../utils/GetHeader";
+import { HEADER_X_TARGET_URL } from "../../../../../../shared/constants";
 const DuoApi = require("@duosecurity/duo_api");
 
 interface DuoResponse {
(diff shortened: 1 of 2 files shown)


server/src/lib/routes/secondfactor/duo-push/Post.ts
```diff
@@ -17,11 +17,13 @@ interface DuoResponse {
   stat: "OK" | "FAIL";
 }
 
-function triggerAuth(username: string, config: DuoPushConfiguration): Promise<DuoResponse> {
+function triggerAuth(username: string, config: DuoPushConfiguration, req: Express.Request): Promise<DuoResponse> {
   return new Promise((resolve, reject) => {
+    const clientIP = req.headers["x-forwarded-for"] || req.connection.remoteAddress;
```
clems4ever: Use GetHeader method in server/src/lib/utils.

server/src/lib/routes/secondfactor/duo-push/Post.ts
```diff
@@ -17,11 +17,13 @@ interface DuoResponse {
...
+    const targetURL = req.headers["x-target-url"];
```
clems4ever: same comment as above.

server/src/lib/routes/secondfactor/duo-push/Post.ts
clems4ever: req.ip would be better by the way, it's the single IP of the client, X-Forwarded-For can be a chain of IPs.

clems4ever: and req.ip is the IP used in the logger.

Issue #333, opened 2019-03-16
Unable to bypass on certain resources for domain


jkaberg:
With the following config I'm still presented with an login window? I was under the presumption that if I was trying to access for instance `https://sub.domain.tld/path1/api/...` I would not need to authenticate? :smile: 

Small config snippet
```
access_control:
  default_policy: deny
  rules:
    - domain: 'sub.domain.tld'
      resources:
        - '^/path1/rpc.*$'
        - '^/path2/api.*$'
        - '^/path3/api.*$'
      policy: bypass

    - domain: '*.domain.tld'
      subject: 'group:admins'
      policy: two_factor
```

clems4ever:
Hello @jkaberg, the path you provided in your example matches the second rule with police two_factor. So, AFAIC, the behavior you see is expected.

Please double check your config.

jkaberg:
@clems4ever okay, so perhaps I've misunderstood the docs/example? How can I make the resources at sub.domain.tld (the api endpoints) available for anyone without auth regardless if they're authenticated or not?

clems4ever:
Helo @jkaberg ,

clems4ever:
Hello @jkaberg ,
```
access_control:
  default_policy: deny
  rules:
  - domain: 'sub.domain.tld'
    policy: bypass
```
should be enough for any resource from `sub.domain.tld` to be allowed to everyone.

In your first example you provided `/path1/api/...` which either matches your second rule if you're an admin or the default policy if you're not.

For the latest example you require, the following should be ok
```
access_control:
  default_policy: deny
  rules:
  - domain: 'sub.domain.tld'
    resources:
    - "^/path1/.*$"
    policy: bypass
```

jkaberg:
Well that doesn't work @clems4ever, my config is exactly like what you mentioned and I'm still presented with the auth window. I guess this is a bug?

On another note, my impression when reading the following config is that (in this particular order):
- IF domain IS sub.domain.tld AND resource (path) IS x/y/z REQUIRE bypass (none)
- ELSE IF domain IS *.domain.tld AND "subject" (user/group) IS group:admins REQUIRE two_factor

Am I wrong in how I interpret this? I feel that the documentation is a bit lacking in this regard.
```
access_control:
  default_policy: deny
  rules:
    - domain: 'sub.domain.tld'
      resources:
        - '^/path1/rpc.*$'
        - '^/path2/api.*$'
        - '^/path3/api.*$'
      policy: bypass

[...]

Issue #88, opened 2017-09-21
Make Authelia and Traefik play nice


jkaberg:
Forward auth has been implemented in Traefik, which is supposed to be similar too ```ngx_http_auth_request_module``` which Authelia depends on in NGINX. 
https://github.com/containous/traefik/pull/1972
https://github.com/containous/traefik/pull/2110

Sample traefik.toml config:
```
defaultEntryPoints = ["https"]
[entryPoints]
  [entryPoints.https]
  address = ":443"
  [entryPoints.https.auth.forward]
  address = "https://auth.domain.tld/verify"
```

However Traefik stalls indefinitely when this option is enabled in Traefik and sending forward auth to Authelia. I don't know if this is related to https://github.com/containous/traefik/pull/2127, or if it's an different issue. It seem's like whatever Authelia responds isn't parsed properly by Traefik.

I'm going to use this issue to track the problem. Do you have any ideas @clems4ever?

Solving this would remove the NGINX dependency entirely in an Traefik environment when using Authelia

SerialVelocity:
It shouldn't stall but it will return a 401 instead of a 302 that traefik needs. I filed this as #85.

clems4ever:
Hello @jkaberg , I gave a try to Traefik. This is the first time I use it and I'm a bit unhappy since it only allows you to forward auth requests at the entrypoint level. That means you can either secure everything or nothing...
Btw, I'm still trying to redirect to Authelia login page but I have not made it work so far.

jkaberg:
@clems4ever yes thats true. You will need to be more precise in the Authelia config file on what you want to protect and not (add more to default group etc.). One add a feature request @ Traefik to allow certain Traefik "frontends" to be protected by Authelia via [labels](https://docs.traefik.io/configuration/backends/docker/#on-containers) (eg. ```traefik.forward.auth=https://auth.domain.tld/verify```)

Let me know how it goes, and if you need any help

clems4ever:
Hello @jkaberg , make Authelia work with Traefik is a bit more complicated than I expected unfortunately. I have not succeeded yet.

jkaberg:
@clems4ever I'm not 100% sure but I think this depends on https://github.com/containous/traefik/pull/2127

clems4ever:
@jkaberg , indeed this is the problem I had while testing...

SerialVelocity:
Traefik v1.5.0-rc1 has now been released with that PR

SerialVelocity:
[...]

PR #348, opened 2019-03-29
Build x-original-url from forwarded headers

ViViDboarder:
This is to allow broader support for proxies. In particular, this allows
support with Traefik.

This patch also includes some examples of configuration with Traefik.

This is to hopefully improve #88 as well as some comments made in #333.

### DIFF
--- a/example/traefik/config.minimal.yml
+++ b/example/traefik/config.minimal.yml
@@ -1,0 +1,61 @@
+###############################################################
+#                Authelia minimal configuration               #
+###############################################################
+
+logs_level: debug
+
+authentication_backend:
+  file:
+    path: /etc/authelia/users_database.yml
+
+session:
+  secret: unsecure_session_secret
+  domain: example.com
+
+# Configuration of the storage backend used to store data and secrets. i.e. totp data
+storage:
+  local:
+    path: /etc/authelia/storage
+
+# TOTP Issuer Name
+#
+# This will be the issuer name displayed in Google Authenticator
+# See: https://github.com/google/google-authenticator/wiki/Key-Uri-Format for more info on issuer names
+totp:
+  issuer: example.com
+
+# Access Control
+#
+# Access control is a set of rules you can use to restrict user access to certain 
+# resources.
+access_control:
+  # Default policy can either be `bypass`, `one_factor`, `two_factor` or `deny`.
+  default_policy: deny
+
+  rules:
+    - domain: traefik.example.com
+      policy: two_factor
+    - domain: who.example.com
+      policy: two_factor
+
+# Configuration of the authentication regulation mechanism.
+regulation: 
+  # Set it to 0 to disable max_retries.
+  max_retries: 3
+
+  # The user is banned if the authenticaction failed `max_retries` times in a `find_time` seconds window. 
+  find_time: 120
+
+  # The length of time before a banned user can login again.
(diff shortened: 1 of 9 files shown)


example/traefik/config.minimal.yml
```diff
@@ -0,0 +1,61 @@
+###############################################################
```
clems4ever: This config will end up in a suite with tests instead.
I'll do it for you in the next commit if you want. This will make Traefik officially supported in Authelia project!

example/traefik/users_database.yml
```diff
@@ -0,0 +1,29 @@
+###############################################################
```
clems4ever: This will end up in a suite.

server/src/lib/utils/RequestUrlGetter.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    if (HasHeader(req, Constants.HEADER_X_ORIGINAL_URL)) {
+      return GetHeader(req, Constants.HEADER_X_ORIGINAL_URL);
+    }
+
+    const proto = GetHeader(req, Constants.HEADER_X_FORWARDED_PROTO);
+    const host = GetHeader(req, Constants.HEADER_X_FORWARDED_HOST);
+    const port = GetHeader(req, Constants.HEADER_X_FORWARDED_PORT);
+    const uri = GetHeader(req, Constants.HEADER_X_FORWARDED_URI);
+
```
clems4ever: Throw here if X-Original-Url one of those is missing. Something like

```
if (!proto || !host || !port || !uri) {
    throw new Error("Missing headers holding user requested URL.");
}
```

clems4ever:
Apparently there is a way around this fix: https://github.com/clems4ever/authelia/issues/88#issuecomment-473344850.

jkaberg:
@clems4ever yes I'm using `traefik.frontend.headers.customRequestHeaders` with https://github.com/clems4ever/authelia/issues/333 hence my comment [here](https://github.com/clems4ever/authelia/issues/333#issuecomment-477538635)

clems4ever:
> @clems4ever yes I'm using `traefik.frontend.headers.customRequestHeaders` with #333 hence my comment [here](https://github.com/clems4ever/authelia/issues/333#issuecomment-477538635)

Sorry, did not get it. Clearer now. Thanks.

server/src/lib/utils/RequestUrlGetter.ts
```diff
@@ -0,0 +1,20 @@
+import Constants = require("../../../../../shared/constants");
+import Express = require("express");
+import GetHeader from "../../utils/GetHeader";
+import HasHeader from "../..//utils/HasHeader";
```
clems4ever: please remove double slash.

example/traefik/config.minimal.yml
ViViDboarder: Thanks! Do you want me to remove this then? Or should I leave it so you can move it wherever it needs to go?

clems4ever: I will do it. Thanks.

clems4ever:
I have something working in a suite. I'm trying to fix the tests now.

jkaberg:
@clems4ever looking forward to it :rocket:

clems4ever:
Superseeded by #353

PR #350, opened 2019-04-08
Fix attaching User/Groups headers for bypass strategy

nightah:
This change includes attaching the following headers when an authenticated session exists while using the bypass strategy:

- Remote-User
- Remote-Groups


clems4ever:
Sorry for the delay. LGTM.

PR #353, opened 2019-04-11
Make Traefik officially supported by Authelia.

clems4ever:
This PR uses X-Forwarded-* headers to build the URL of the target website and append it to the login page URL for Traefik to work. I also create a suite for testing a setup using Traefik.

I removed the Redirect header previously sent by Authelia /api/verify endpoint because the information can be computed by nginx directly and Traefik cannot leverage it anyway.


server/src/lib/utils/RequestUrlGetter.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class RequestUrlGetter {
+  static getOriginalUrl(req: Express.Request): string {
+
+    if (HasHeader(req, Constants.HEADER_X_ORIGINAL_URL)) {
+      return GetHeader(req, Constants.HEADER_X_ORIGINAL_URL);
+    }
+
+    const proto = GetHeader(req, Constants.HEADER_X_FORWARDED_PROTO);
+    const host = GetHeader(req, Constants.HEADER_X_FORWARDED_HOST);
```
jkaberg: Any reason not to include the port as @ViViDboarder did? https://github.com/clems4ever/authelia/blob/9304e834162b8f237a08367b08d95477ef2ed035/server/src/lib/utils/RequestUrlGetter.ts#L15

Note that this could very well differ from 80 and 443 in some instances.

server/src/lib/utils/RequestUrlGetter.ts
```diff
@@ -12,14 +12,12 @@ export class RequestUrlGetter {
 
     const proto = GetHeader(req, Constants.HEADER_X_FORWARDED_PROTO);
     const host = GetHeader(req, Constants.HEADER_X_FORWARDED_HOST);
-    const port = GetHeader(req, Constants.HEADER_X_FORWARDED_PORT);
     const uri = GetHeader(req, Constants.HEADER_X_FORWARDED_URI);
 
-    if (!proto || !host || !port) {
```
ViViDboarder: Similar point to @jkaberg, the port ought to always be present, but it's possible that the URI is empty. Checking that it's truthy seems like it could end up with false negatives.

I will admit that I hadn't quite validated everything, so my assumptions may be wrong, but is there a reason for making this change?

server/src/lib/utils/RequestUrlGetter.ts
clems4ever: Well in my current setup X-Forwarded-Host embeds the port but I guess I misconfigured nginx. I will have a look.

server/src/lib/utils/RequestUrlGetter.ts
clems4ever: You're right, I will test that out.

clems4ever: Authelia does not need X-Forwarded-Port eventually, the port is included in X-Forwarded-Host, building the url out of x-Forwarded-Port would be redundant.

Log that confirms what I'm saying:
```
debug: date='Thu Apr 11 2019 21:52:11 GMT+0200 (CEST)' method='GET', path='/api/verify' requestId='cc6e2002-df4b-4774-9d24-eee7a708648c' sessionId='zpB5cJqbdABqeKC9fhnUjvi3CnlEG_Sm' ip='192.168.240.1' message='Header x-forwarded-host is set to secure.example.com:8080'

File: server/src/lib/utils/RequestUrlGetter.ts (hunk 1 above)
Comment (clems4ever) [reply server/src/lib/utils/RequestUrlGetter.ts]: Please see my answer in below comment. X-Fowarded-Host hold the domain AND the port as you can see in the following log:
```debug: date='Thu Apr 11 2019 21:52:11 GMT+0200 (CEST)' method='GET', path='/api/verify' requestId='cc6e2002-df4b-4774-9d24-eee7a708648c' sessionId='zpB5cJqbdABqeKC9fhnUjvi3CnlEG_Sm' ip='192.168.240.1' message='Header x-forwarded-host is set to secure.example.com:8080'```
And in my setup Traefik is not on port 443 or 80 but 8080 instead.

File: server/src/lib/IdentityCheckMiddleware.ts [hunk 3]
```diff
@@ -111,8 +113,9 @@ export function post_start_validation(handler: IdentityValidable,
           vars.userDataStore);
       })
       .then((token: string) => {
-        const host = req.get("Host");
-        const link_url = util.format("https://%s/#%s?token=%s", host,
+        const scheme = GetHeader(req, Constants.HEADER_X_FORWARDED_PROTO);
```
clems4ever: This is breaking.

example/compose/nginx/portal/nginx.conf.ejs
```diff
@@ -35,10 +35,15 @@ http {
 
         # Serve the backend API for the portal.
         location /api {
-            proxy_set_header  Host $http_host;
```
clems4ever: This is breaking the links in identity validation emails. it should be replaced by X-Forwarded-Host instead.

server/src/lib/routes/verify/Get.ts
```diff
@@ -24,42 +25,43 @@ async function verifyWithSelectedMethod(req: Express.Request, res: Express.Respo
   }
 }
 
-/**
- * The Redirect header is used to set the target URL in the login portal.
- *
- * @param req The request to extract X-Original-Url from.
- * @param res The response to write Redirect header to.
- */
-function setRedirectHeader(req: Express.Request, res: Express.Response) {
```
clems4ever: This is breaking, please remove the use of **upstream_http_redirect** in nginx configuration and replace with **$scheme://$http_host$request_uri;**

clems4ever:
@jkaberg , @ViViDboarder , would you guys be ok to write a small documentation for Traefik under docs/proxies?

server/src/lib/routes/secondfactor/u2f/register/post.ts
```diff
@@ -10,12 +9,16 @@ import { ServerVariables } from "../../../../ServerVariables";
...
 import UserMessages = require("../../../../../../../shared/UserMessages");
 import { AuthenticationSession } from "../../../../../../types/AuthenticationSession";
+import GetHeader from "../../../../utils/GetHeader";
+import * as Constants from "../../../../../../../shared/constants";
 
 
 export default function (vars: ServerVariables) {
   function handler(req: express.Request, res: express.Response): BluebirdPromise<void> {
     let authSession: AuthenticationSession;
-    const appid = u2f_common.extract_app_id(req);
```
clems4ever: [BREAKING] This was using the Host header but now uses the X-Forwarded-Host header.

server/src/lib/routes/secondfactor/u2f/register_request/get.ts
```diff
@@ -8,11 +7,15 @@ import { AuthenticationSessionHandler } from "../../../../AuthenticationSessionH
 import { AuthenticationSession } from "../../../../../../types/AuthenticationSession";
 import UserMessages = require("../../../../../../../shared/UserMessages");
 import { ServerVariables } from "../../../../ServerVariables";
+import GetHeader from "../../../../utils/GetHeader";
+import * as Constants from "../../../../../../../shared/constants";
 
 export default function (vars: ServerVariables) {
   function handler(req: express.Request, res: express.Response): BluebirdPromise<void> {
     let authSession: AuthenticationSession;
-    const appid: string = u2f_common.extract_app_id(req);
```
clems4ever: [BREAKING] This was using the Host header but now uses the X-Forwarded-Host header.

server/src/lib/routes/secondfactor/u2f/sign/post.ts
```diff
@@ -11,11 +10,15 @@ import { AuthenticationSessionHandler } from "../../../../AuthenticationSessionH
 import UserMessages = require("../../../../../../../shared/UserMessages");
 import { AuthenticationSession } from "../../../../../../types/AuthenticationSession";
 import { Level } from "../../../../authentication/Level";
+import GetHeader from "../../../../utils/GetHeader";
+import * as Constants from "../../../../../../../shared/constants";
 
 export default function (vars: ServerVariables) {
   function handler(req: express.Request, res: express.Response): BluebirdPromise<void> {
     let authSession: AuthenticationSession;
-    const appId = u2f_common.extract_app_id(req);
```
clems4ever: [BREAKING] This was using the Host header but now uses the X-Forwarded-Host header.

server/src/lib/routes/secondfactor/u2f/sign_request/get.ts
```diff
@@ -9,28 +7,31 @@ import { AuthenticationSessionHandler } from "../../../../AuthenticationSessionH
 import UserMessages = require("../../../../../../../shared/UserMessages");
 import { ServerVariables } from "../../../../ServerVariables";
 import { AuthenticationSession } from "../../../../../../types/AuthenticationSession";
+import GetHeader from "../../../../utils/GetHeader";
+import * as Constants from "../../../../../../../shared/constants";
 
 export default function (vars: ServerVariables) {
   function handler(req: express.Request, res: express.Response): BluebirdPromise<void> {
     let authSession: AuthenticationSession;
-    const appId = u2f_common.extract_app_id(req);
```
clems4ever: [BREAKING] This was using the Host header but now uses the X-Forwarded-Host header.

server/src/lib/utils/RequestUrlGetter.ts
jkaberg: Thanks @clems4ever, just wanted to be sure 😄

PR #355, opened 2019-04-16
Update doc/proxies/nginx.md examples

nightah:
This is to ensure that the proxies documentation for nginx is in-line with the latest breaking changes that have been made.
The existing examples no longer will work, and will result in a blank redirection.


clems4ever:
Thank you for this contribution @nightah .

PR #370, opened 2019-05-31
Allow passwords to be specified via environment

rajha-korithrien:
This adds the following major features:
- The ability to have passwords passed in via environment variables rather than the configuration file

This feature is needed to allow the use of Kubernetes secrets when deploying Authelia in Kubernetes via a Docker image

This adds the following minor features:
- Better debugging output for domain/subdomain comparison
- additions to .gitignore for vim and idea editors
- A .dockerignore file to make the copy of the initial context to the docker daemon go faster

These features were necessary to increase developer productivity and make it simpler to track down domain/cookie issues.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git
+.cache
+**/node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ src/.baseDir.ts
 .vscode/
 
 *.swp
-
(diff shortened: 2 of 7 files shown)


clems4ever:
Hello @rajha-korithrien , thanks for the PR. It looks good and definitely helps with deployments on Kube. 
Sorry for the late reply though. I've been busy lately.

PR #375, opened 2019-06-25
Add Support for Private Certificate Authority Signed LDAPS Server Certificates

rajha-korithrien:
What this pull request does:
* Provides users the ability to specify the CA of their ldaps server when such server uses a private CA.
* Added some info and debug logging to understand how Authelia initializes the TLS system in ldapjs

Why do we need it:
* Not all ldaps servers will use a certificate signed by a publicly trusted CA. For example: in corporate environments with an existing authentication stack, or in a Kubernetes environment where the ldaps server uses a certificate signed by the cluster CA
* The logging makes it easier to track what is going on when setting up an ldaps backend.

Risks:
The version of Node was moved from the 8.x branch to the 10.x branch which uses an improved TLS stack. The resulting Docker image has been tested in:
* ldaps with publicly trusted CA (which was working before and continues to work)
* ldaps with a private CA (which was not working before and is working now)
* ldap non-TLS configurations (which was working before and continues to work).

Currently there is an issue with Node where certain self-signed certificates do not work which is addressed in nodejs/node#22906. This impacts the use of private CA Active Directory certificates, and the patch 22906 was tested to work with Authelia as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.7.0-alpine
+FROM node:10.16.0-alpine
 
 WORKDIR /usr/src
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10.16.0-alpine
+FROM node:8.7.0-alpine
 
 WORKDIR /usr/src
 
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -9,7 +9,7 @@
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "requires": {
-        "@babel/highlight": "7.0.0"
+        "@babel/highlight": "^7.0.0"
       }
     },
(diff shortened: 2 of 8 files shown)


rajha-korithrien:
There are 3 failures in the tests, but I don't understand how my changes caused them. I didn't add ldaps tests, and the failures seem to be in the kubernetes suite, which this pull request does not modify.

clems4ever:
Hello @rajha-korithrien , the tests related to dockerhub have been removed in master. A rebase should solve those. For kube, I'll try to run the tests on master and if they are green I will try to run them locally on your branch.

clems4ever:
Tests are green in master. Please rebase and I'll help you out finding what's going on with Kube tests.

clems4ever:
It might be related to the upgrade of node in the docker image btw. Did you test the image?

rajha-korithrien:
@clems4ever I have tested the image, it is currently operating in a development environment I am working on. It works fine in that environment. I do admit that I have only tested it via ldaps and not ldap.

rajha-korithrien:
I did the suggested rebase to master, and set the Node version back to 8.7.0. This version of Node has some known errors processing certain forms of TLS certificates (specifically those with blank CN, but will work for many TLS certs).

All tests pass now :-)

clems4ever:
Thanks for your rebase and fix. Let's merge it as it is and wait for the next major coming soon to do more improvements on top of it. I think the problem of hanging session will naturally disappear.
Sound good?

PR #389, opened 2019-08-29
Theming support is back ;)

BankaiNoJutsu:
Hey Clément! :)

I'm back for at least some visual stuff, as I lack the pure coding skills...

There are three themes as of now:
- the original one
- black on white
- white on black

$theme-used variable can be changed in client/src/assets/scss/_variables.scss to choose the theme.
Sadly i was not able to make the themes dynamic, so they need to be specified before the build process.

Please have a look and let me know if something is to change.
Also I saw you have been working on a go version, will check it out too.

Cheers again for this wonderful piece of software.


davama:
Look forward to testing this out!

It'd be awesome if this were dynamic from the config.yml file

BankaiNoJutsu:
Should be doable, we just need to parse and push the theme variable from the yml conf to the service i guess. However not sure how to proceed

clems4ever:
@BankaiNoJutsu , I'm really sorry but this PR is unlikely to make it through as I'm about to finish rewriting the UI once again to simplify it and use the official material-ui library for React to have a better looking and a smoother user experience.
By the way, theming is given by the library itself so that it's also easier to manage for future themes development. Moreover, the theme will be switchable at runtime if needed in the future.

Returning to your PR, I'm really sorry but this time is the last time I refactor the frontend so deeply. This refactoring is kind of an enabler I would say. It'll come up in a couple of days, I'm at the end of it. I'd be very happy to have your feedback (as well as @davama ) on the default theme and the user experience then. If you're not yet on Gitter, I invite you to join and have a chat!

Sorry again...

PR #427, opened 2019-11-03
Add "{CRYPT}" detail to deployment-dev

Peaches491:
Lost a couple hours to this subtle detail of the `users_database.yml` format. I was missing the `{CRYPT}` at the beginning of the hashed password line. 

Alternatively, the `hash-password` tool could prepend this automatically, but I'm pretty new to Authelia, and so I thought a README edit was better first PR 😜


clems4ever:
Thank you for this update @Peaches491. Please note that I will remove the necessity to prepend "{CRYPT}" in v4 because it's not mandatory at all for file-based users databases. I just wanted to mirror the behavior from OpenLDAP at the time.

PR #429, opened 2019-11-04
Build a statically linked version of Authelia.

clems4ever:


theoparis:
Why is this closed? I was looking for a static build of authelia and ended up copying the build command from this PR...

PR #430, opened 2019-11-05
Enable Multiarch docker builds

nightah:
This change provides Authelia builds for the following architectures:
- amd64
- arm32v7
- arm64v8

The arm32 variant has been tested as I own the respective hardware, I'd recommend you test the arm64 variant with your hardware @clems4ever. It's also available at `nightah/authelia` just pull the latest tag to ensure the manifests work too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,62 @@
 language: go
+
 required: sudo
+
 go:
-- '1.13'
+  - '1.13'
+
 services:
-- docker
-- ntp
-- xvfb
-addons:
-  chrome: stable
-  apt:
-    sources:
-    - google-chrome
-    packages:
-    - libgif-dev
-    - google-chrome-stable
+  - docker
+  - ntp
+  - xvfb
 
 before_script:
-- curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
-- nvm install v11 && nvm use v11 && npm i
-script:
-- "source bootstrap.sh"
-- "authelia-scripts ci"
-after_success:
-- "authelia-scripts docker publish"
+  - export PATH=./cmd/authelia-scripts/:/tmp:$PATH
 
-# TODO(c.michaud): publish built artifact on Github.
+jobs:
+  include:
+    - stage: test
+      addons:
+        chrome: stable
+        apt:
+          sources:
+            - google-chrome
+          packages:
+            - libgif-dev
+            - google-chrome-stable
+      script:
+        - curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
+        - nvm install v11 && nvm use v11 && npm i
+        - source bootstrap.sh
+        - authelia-scripts ci
+      # TODO(c.michaud): publish built artifact on Github.
+    - &build-images
+      stage: build images
+      env:
+        - ARCH=amd64
+      install: skip
+      script:
+        - while sleep 9m; do echo '===== Prevent build from terminating  ====='; done &
+        - authelia-scripts docker build --arch=$ARCH
+        - kill %1
+      after_success:
+        - authelia-scripts docker push-image --arch=$ARCH
+    - <<: *build-images
+      env:
+        - ARCH=arm32v7
+    - <<: *build-images
+      env:
+        - ARCH=arm64v8
+    - stage: deploy manifests
+      env:
+        - DOCKER_CLI_EXPERIMENTAL=enabled
+      install: skip
+      script:
+        - authelia-scripts docker push-manifest
 
 notifications:
   email:
     recipients:
-    - clement.michaud34@gmail.com
+      - clement.michaud34@gmail.com
     on_success: change
     on_failure: always
--- a/cmd/authelia-scripts/cmd_docker.go
+++ b/cmd/authelia-scripts/cmd_docker.go
@@ -5,13 +5,67 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 )
 
-func DockerBuildOfficialImage() error {
+var arch string
+var supportedArch = []string{"amd64", "arm32v7", "arm64v8"}
+var defaultArch = "amd64"
+
+func init() {
+	DockerBuildCmd.PersistentFlags().StringVar(&arch, "arch", defaultArch, "target architecture among: "+strings.Join(supportedArch, ", "))
+	DockerPushCmd.PersistentFlags().StringVar(&arch, "arch", defaultArch, "target architecture among: "+strings.Join(supportedArch, ", "))
+
+}
+
+func checkArchIsSupported(arch string) {
+	for _, a := range supportedArch {
+		if arch == a {
+			return
+		}
+	}
+	log.Fatal("Architecture is not supported. Please select one of " + strings.Join(supportedArch, ", ") + ".")
+}
+
+func dockerBuildOfficialImage(arch string) error {
 	docker := &Docker{}
-	return docker.Build(IntermediateDockerImageName, ".")
+	// Set default Architecture Dockerfile to amd64
+	dockerfile := "Dockerfile"
+
+	// If not the default value
+	if arch != defaultArch {
+		dockerfile = fmt.Sprintf("%s.%s", dockerfile, arch)
+	}
+
+	if arch == "arm32v7" {
+		err := CommandWithStdout("docker", "run", "--rm", "--privileged", "multiarch/qemu-user-static", "--reset", "-p", "yes").Run()
+
+		if err != nil {
+			panic(err)
+		}
+
+		err = CommandWithStdout("bash", "-c", "wget https://github.com/multiarch/qemu-user-static/releases/download/v4.1.0-1/qemu-arm-static -O ./qemu-arm-static && chmod +x ./qemu-arm-static").Run()
+
+		if err != nil {
+			panic(err)
+		}
+	} else if arch == "arm64v8" {
+		err := CommandWithStdout("docker", "run", "--rm", "--privileged", "multiarch/qemu-user-static", "--reset", "-p", "yes").Run()
+
+		if err != nil {
+			panic(err)
+		}
+
+		err = CommandWithStdout("bash", "-c", "wget https://github.com/multiarch/qemu-user-static/releases/download/v4.1.0-1/qemu-aarch64-static -O ./qemu-aarch64-static && chmod +x ./qemu-aarch64-static").Run()
+
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	return docker.Build(IntermediateDockerImageName, dockerfile, ".")
 }
 
 // DockerBuildCmd Command for building docker image of Authelia.
@@ -19,7 +73,8 @@ var DockerBuildCmd = &cobra.Command{
 	Use:   "build",
 	Short: "Build the docker image of Authelia",
 	Run: func(cmd *cobra.Command, args []string) {
-		err := DockerBuildOfficialImage()
+		checkArchIsSupported(arch)
+		err := dockerBuildOfficialImage(arch)
 
 		if err != nil {
 			log.Fatal(err)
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -1,34 +1,57 @@
...
+      - docker-ce
+      - libgif-dev
+      - google-chrome-stable
 
 before_script:
-- curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
-- nvm install v11 && nvm use v11 && npm i
+  - curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
+  - nvm install v11 && nvm use v11 && npm i
+  - docker run --rm --privileged multiarch/qemu-user-static:register --reset -p yes
```
clems4ever: Please comment what this command is supposed to do. I'm not quite sure by just reading it. Don't hesitate to also paste the link where you found it if it's the case.

cmd/authelia-scripts/cmd_docker.go
```diff
@@ -43,6 +44,15 @@ var DockerPushCmd = &cobra.Command{
 	},
 }
 
+// DockerManifestCmd Command for pushing Authelia docker manifest to Dockerhub
+var DockerManifestCmd = &cobra.Command{
```
clems4ever: I think this should be included in the push command.

cmd/authelia-scripts/cmd_docker.go
```diff
@@ -11,7 +11,8 @@ import (
 
 func DockerBuildOfficialImage() error {
 	docker := &Docker{}
-	return docker.Build(IntermediateDockerImageName, ".")
+	Dockerfile := os.Getenv("DOCKERFILE")
```
clems4ever: the Dockerfile could be deduced from an -arch flag provided by the user. No need for an env variable.

cmd/authelia-scripts/cmd_docker.go
```diff
@@ -83,21 +93,60 @@ func deploy(docker *Docker, tag string) {
...
+
+	if err != nil {
+		panic(err)
+	}
+}
+
 func publishDockerImage() {
 	docker := &Docker{}
 
+	ARCH := os.Getenv("ARCH")
```
clems4ever: I'd prefer having a flag for this option so that it is documented in the usage of the command. The other environment variables are env vars because they explicitly come from Travis and do not need to be known by the user to run the command. This is not the case or this arch customization, and the users should be aware of it from the usage to avoid running the command without the flag and get an erroneous name of their docker image.

You can use amd64 as the default value if the flag is not provided.

cmd/authelia-scripts/docker.go
```diff
@@ -4,8 +4,8 @@ package main
 type Docker struct{}
 
 // Build build a docker image
-func (d *Docker) Build(tag string, target string) error {
-	return CommandWithStdout("docker", "build", "-t", tag, target).Run()
+func (d *Docker) Build(tag string, dockerfile string, target string) error {
```
clems4ever: why not parameterizing the command with an architecture instead of the name of a dockerfile? That would probably make more sense for a user of the command. Btw, this arch customization should be a documented flag of the build command. That way users can see it in the usage of the command.

cmd/authelia-scripts/docker.go
```diff
@@ -22,3 +22,12 @@ func (d *Docker) Login(username, password string) error {
 func (d *Docker) Push(tag string) error {
 	return CommandWithStdout("docker", "push", tag).Run()
 }
+
+// Manifest push a docker manifest to dockerhub.
+func (d *Docker) Manifest(tag string, amd64tag string, arm32v7tag string, arm64v8tag string) error {
+	return CommandWithStdout("docker", "manifest", "push", "--purge", tag).Run()
```
clems4ever: doing a return at every line is a bug. Only the first command will be executed. The rest is unreachable code.

cmd/authelia-scripts/main.go
```diff
@@ -42,7 +42,7 @@ var Commands = []AutheliaCommandDefinition{
 	AutheliaCommandDefinition{
 		Name:        "docker",
 		Short:       "Commands related to building and publishing docker image",
-		SubCommands: CobraCommands{DockerBuildCmd, DockerPushCmd},
+		SubCommands: CobraCommands{DockerBuildCmd, DockerPushCmd, DockerManifestCmd},
```
clems4ever: As said previously, this should probably not be a specific command but be included in the push command instead.

Dockerfile
```diff
@@ -10,7 +10,7 @@ WORKDIR /go/src/app
 COPY . .
 
 # CGO_ENABLED=1 is mandatory for building go-sqlite3
-RUN cd cmd/authelia && GOOS=linux GOARCH=amd64 CGO_ENABLED=1 go build -o authelia
+RUN cd cmd/authelia && GOOS=linux GOARCH=amd64 CGO_ENABLED=1 go build -tags netgo -ldflags '-w' -o authelia
```
clems4ever: how is this related to enabling multiarch?

Dockerfile.arm64
```diff
@@ -0,0 +1,46 @@
...
+RUN apk --no-cache add gcc musl-dev
+
+WORKDIR /go/src/app
+COPY . .
+
+# CGO_ENABLED=1 is mandatory for building go-sqlite3
+RUN cd cmd/authelia && GOOS=linux GOARCH=arm64 CGO_ENABLED=1 go build -tags netgo -ldflags '-w' -o authelia
+
+
+# ========================================
```
clems4ever: this frontend section could be extracted in its own Dockerfile to avoid duplication. This also means changing the build command accordingly but it's not a big deal IMO.

https://stackoverflow.com/questions/53659993/docker-multi-stage-how-to-split-up-into-multiple-dockerfiles

.travis.yml
nightah: [multiarch/qemu-user-static](https://github.com/multiarch/qemu-user-static) is to enable an execution of different multi-architecture containers by QEMU and binfmt_misc.

cmd/authelia-scripts/cmd_docker.go
nightah: The problem is that the push is executed in each build stage within Travis, as they're all in different hosts.

The manifest needs all of the images to be present before it can be created hence why the deploy step in the current Travis config is only executed if all builds are successful.

cmd/authelia-scripts/cmd_docker.go
nightah: I thought the purpose of these scripts were just to help with your Travis setup.

I didn't really think or know that you intended for users to utilise these too, am I understanding correctly that is what the actual intent is?

cmd/authelia-scripts/docker.go
nightah: I think this ties into my comment further above regarding the fact that I thought these were just helpers for the Travis deployment pipeline.

In the multistage Travis build we are specifying the Architecture and the respective Dockerfile for said Arch.

Dockerfile
nightah: Tagged netgo to make sure we use Go built-in net package and not the system’s one.
ldflags -w just disables debug letting the file be smaller.

Dockerfile.arm64
nightah: There's no saving here, as this has to be done three times as each build step is on it's own respective host.

Unless we decided to do all the build steps in a single run, which could actually cause other timeout issues with Travis, then it should stay as is.

cmd/authelia-scripts/cmd_docker.go
clems4ever: In my mind the only script not meant to be executed by users is `authelia-scripts ci`. The other ones are here for users to have a built version of Authelia or have a local docker image or run integration tests and unit tests locally. In my mind the interface of all those scripts should be user-friendly and the ci one could be relaxed.

Now I understand the fact that env variables is the way to go in Travis so I guess the best solution would be to define the env variable in Travis and forward them as flag(s) to the CLI. I think the right level of abstraction for that script would be the architecture now. Would it work to have a flag -arch and build the Dockerfile name from that arch? Something like Dockerfile.$arch and then the user (likely me for now) could do `authelia-scripts docker build -arch=amd64`, `authelia-scripts docker push-image -arch=amd64` and `authelia-scripts docker push-manifest`

In the ci script we could probably run something like `authelia-scripts docker build -arch=$ARCH` so that you can export ARCH as the env variable in Travis and make it consumed by the script.

Does it make sense?

cmd/authelia-scripts/cmd_docker.go
clems4ever: As discussed in previous thread, this script is for users and so must have a human friendly interface even if it can (and is) be called from the CI. An optional flag is a human friendly way of configuring the script and documenting the feature to users. Moreover it does not cost anything compared to the configuration by env variable. Regarding your concern about not setting the arch, I'd say that if a user omit it let's use amd64 as the default value.

Does it sound reasonable?

cmd/authelia-scripts/docker.go
clems4ever: You are right, I misread this part. Dockerfile name is the right level of abstraction in this helper function. My bad. Nothing to check regarding Travis though.

clems4ever:
Very nice work so far! Thanks for the contribution.

cmd/authelia-scripts/docker.go
```diff
@@ -22,3 +22,12 @@ func (d *Docker) Login(username, password string) error {
...
 	return CommandWithStdout("docker", "push", tag).Run()
 }
+
+// Manifest push a docker manifest to dockerhub.
+func (d *Docker) Manifest(tag string, amd64tag string, arm32v7tag string, arm64v8tag string) error {
+	CommandWithStdout("docker", "manifest", "push", "--purge", tag).Run()
+	CommandWithStdout("docker", "manifest", "create", tag, amd64tag, arm32v7tag, arm64v8tag).Run()
+	CommandWithStdout("docker", "manifest", "annotate", tag, arm32v7tag, "--os", "linux", "--arch", "arm").Run()
+	CommandWithStdout("docker", "manifest", "annotate", tag, arm64v8tag, "--os", "linux", "--arch", "arm64", "--variant", "v8").Run()
+	return CommandWithStdout("docker", "manifest", "push", "--purge", tag).Run()
```
clems4ever: Why do we push twice?

cmd/authelia-scripts/main.go
clems4ever: confirmed.

cmd/authelia-scripts/docker.go
```diff
@@ -22,3 +22,12 @@ func (d *Docker) Login(username, password string) error {
 func (d *Docker) Push(tag string) error {
 	return CommandWithStdout("docker", "push", tag).Run()
 }
+
+// Manifest push a docker manifest to dockerhub.
+func (d *Docker) Manifest(tag string, amd64tag string, arm32v7tag string, arm64v8tag string) error {
+	CommandWithStdout("docker", "manifest", "push", "--purge", tag).Run()
```
clems4ever: Add panic statement if there is an error in any of the command. It will stop the execution of next commands on failure to prevent reaching a bad state.

cmd/authelia-scripts/docker.go
nightah: The —purge flag will check to see if a manifest exists locally and remove if so.

It’s a safety net. Second push is where the actual built manifest gets published.

clems4ever: Add a comment on the first line then please because it was not obvious.

cmd/authelia-scripts/cmd_docker.go
nightah: Per discussion in Gitter I think this will still present issues if a person tries to build different Architectures because of the other steps scattered throughout the different Dockerfiles and `.travis.yml`

I'll update now with a solution so the interface is clear for users and the additional logic is utilised when called via Travis.

cmd/authelia-scripts/docker.go
nightah: On second thought this will not be required because it hasn't got cached images/manifests due to Travis always spinning up a new host, I'll remove it.

cmd/authelia-scripts/cmd_docker.go
nightah: The `ARCH` is only utilised when looking to publish images to Docker, which as far as I can see is only within your Travis build pipeline, because it would not execute if the respective Travis variables are not present.

The `ARCH` variable isn't really what's going to cause issues for you as a user, it's the fact that we're now explicitly passing the `Dockerfile` into the helper function for builds.
I've now specified a default value in this helper file which will revert to using `Dockerfile` being the amd64 variant if the `DOCKERFILE` variable isn't present (outside of Travis builds).

This should allow you to continue as per usual and support the broader multistage builds for Travis too.

clems4ever:
Thank you @nightah

PR #437, opened 2019-11-09
Remove unnecessary tags from DockerHub after publishing a manifest

nightah:
Per the title, this utilises Docker's v2 REST API to remove tags that have been published into a manifest.


nightah:
Was actually the following: https://stackoverflow.com/questions/44209644/how-do-i-delete-a-docker-image-from-docker-hub-via-command-line

PR #447, opened 2019-11-17
Removing the conversion from JSON to struct and back

### LINKED ISSUE #426: U2F Authentication - seg violation / nil pointer deref
Getting the following debug logs. I think it's only occurring with U2F keys. I was testing the master build (so this is for your information only) from docker hub on my k8s. It allows me to register a key, but then seems to crash when trying to authenticate.

time="2019-11-02T22:44:48Z" level=trace msg="Request hit" method=GET path=/api/state remote_ip=10.32.0.3
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa7c562]

goroutine 22 [running]:
github.com/clems4ever/authelia/middlewares.(*AutheliaCtx).Error(0x0, 0xe69e00, 0xc0002b1d80, 0xcd3f0a, 0x10)
        /go/src/app/middlewares/authelia_context.go:50 +0xc2
github.com/clems4ever/authelia/middlewares.AutheliaMiddleware.func1.1(0xc0003f0000)
        /go/src/app/middlewares/authelia_context.go:40 +0x15f
github.com/fasthttp/router.(*Router).Handler(0xc0003660a0, 0xc0003f0000)
        /go/pkg/mod/github.com/fasthttp/router@v0.5.2/router.go:326 +0xfb3
github.com/clems4ever/authelia/middlewares.LogRequestMiddleware.func1(0xc0003f0000)
        /go/src/app/middlewares/log_request.go:14 +0xa2
github.com/valyala/fasthttp.(*Server).serveConn(0xc0003ca400, 0xe7ee80, 0xc0000a8208, 0x0, 0x0)
        /go/pkg/mod/github.com/valyala/fasthttp@v1.6.0/server.go:2018 +0x6ab
github.com/valyala/fasthttp.(*workerPool).workerFunc(0xc0003aa8c0, 0xc0002c8aa0)
        /go/pkg/mod/github.com/valyala/fasthttp@v1.6.0/workerpool.go:223 +0xc0
github.com/valyala/fasthttp.(*workerPool).getCh.func1(0xc0003aa8c0, 0xc0002c8aa0, 0xba75a0, 0xc0002c8aa0)
        /go/pkg/mod/github.com/valyala/fasthttp@v1.6.0/workerpool.go:195 +0x35
created by github.com/valyala/fasthttp.(*workerPool).getCh
        /go/pkg/mod/github.com/valyala/fasthttp@v1.6.0/workerpool.go:194 +0x101

Tedyst:
This should fix #426. Any feedback would be appreciated!

### DIFF
--- a/middlewares/authelia_context.go
+++ b/middlewares/authelia_context.go
@@ -24,7 +24,7 @@ func NewAutheliaCtx(ctx *fasthttp.RequestCtx, configuration schema.Configuration
 
 	userSession, err := providers.SessionProvider.GetSession(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("Unable to retrieve user session: %s", err.Error())
+		return autheliaCtx, fmt.Errorf("Unable to retrieve user session: %s", err)
 	}
 
 	autheliaCtx.userSession = userSession
--- a/session/provider.go
+++ b/session/provider.go
@@ -1,7 +1,6 @@
 package session
 
 import (
-	"encoding/json"
 	"time"
 
 	"github.com/clems4ever/authelia/configuration/schema"
@@ -35,21 +34,14 @@ func (p *Provider) GetSession(ctx *fasthttp.RequestCtx) (UserSession, error) {
 		return NewDefaultUserSession(), err
 	}
 
-	userSessionJSON, ok := store.Get(userSessionStorerKey).([]byte)
+	userSession, ok := store.Get(userSessionStorerKey).(UserSession)
 
 	// If userSession is not yet defined we create the new session with default values
 	// and save it in the store.
 	if !ok {
-		userSession := NewDefaultUserSession()
+		userSession = NewDefaultUserSession()
 		store.Set(userSessionStorerKey, userSession)
 		return userSession, nil
-	}
-
-	var userSession UserSession
-	err = json.Unmarshal(userSessionJSON, &userSession)
-
-	if err != nil {
-		return NewDefaultUserSession(), err
 	}
 
 	return userSession, nil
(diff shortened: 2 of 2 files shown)


clems4ever:
Thanks for the PR. It seems to indeed fix 2 bugs. LGTM.

clems4ever:
> I removed the conversion from json to struct and from struct to json, thus removing the bug altogether.

Nice!

Tedyst:
Ok this is pretty weird. I'll try to see why that test fails.

clems4ever:
the same test failed twice in a row so it's definitely not a randomness. Do you know how to run and test a suite?

I should probably display Authelia logs in Travis when a test fails so that it's easier to figure out the actual issue. I'll see what I can do in a future changelist.

Tedyst:
I am running that test right now. For me it just crashes when trying to login. IDK why.

Tedyst:
> I am running that test right now. For me it just crashes when trying to login. IDK why.

For clarification, the test crashes, not authelia. Authelia seems to work perfectly doing the test manually and searching elements in page by class.

clems4ever:
Now I remember why I serialized in JSON... This test is actually using Redis and I think your fix broke it because it tries to store a blob of binary data. In all other cases it's working because the session middleware uses memory but when Redis is used it cannot work.

clems4ever:
It's weird that there is no error reported by the session middleware. It's probably because it tries to store the struct pointer instead of the data...

Tedyst:
Then maybe when using redis we construct the `ellptic.Curve` manually? Cause we can't marshall it.

clems4ever:
I think it's worthless to spend more time on that issue because I fixed both bugs in https://github.com/clems4ever/authelia/pull/446/files. The first problem as been fixed like you did in your first commit. The second by storing serializable versions of the U2F public key.

clems4ever:
> Then maybe when using redis we construct the ellptic.Curve manually? Cause we can't marshall it.

This is what I did yesterday in the PR I've just linked.

Tedyst:
Then I'm going to close the pull request.

> > Then maybe when using redis we construct the ellptic.Curve manually? Cause we can't marshall it.
> 
> This is what I did yesterday in the PR I've just linked.

clems4ever:
My PR should be merged in few minutes. You'll be able to test again afterwards, it should be fine.

PR #458, opened 2019-11-26
Helm chart

ishioni:
Let's start working on this


helm/Chart.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+name: authelia
+version: 0.1.1
+appVersion: 3.16.2
+description: A full-featured authentication server
```
clems4ever: I like that :)

helm/templates/configmap.yaml
```diff
@@ -0,0 +1,65 @@
...
+      ldap:
+{{ toYaml .Values.authelia.ldap.config | indent 8 }}
+      {{- else }}
+      file:
+        path: /var/lib/authelia/users_database.yml
+      {{- end }}
+
+    access_control:
+      default_policy: {{ .Values.authelia.accessControl.defaultPolicy }}
+      rules:
```
clems4ever: For the future, I had in mind to maybe provide an extension to kube ingress configuration to define the level of authentication required by the ingress directly in the ingress config instead of Authelia config. Just an idea as I ready the chart. What do you think?

helm/templates/deployment.yaml
```diff
@@ -0,0 +1,91 @@
...
+          {{- if not .Values.mongodb.enabled }}
+          - name: authelia-store
+            mountPath: /var/lib/authelia
+          {{- end }}
+          {{- if not .Values.authelia.ldap.enable }}
+          - name: authelia-users
+            mountPath: /var/lib/authelia/users_database.yml
+            subPath: users_database.yml
+          {{- end }}
+          - name: authelia-temp
```
clems4ever: Maybe I'm missing something but why do we need to define a temporary directory here?

helm/templates/pvc-config.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if not .Values.authelia.ldap.enabled -}}
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: {{ template "authelia.fullname" . }}-etc
```
clems4ever: the name could be more explicit than `-etc` I guess. We are defining this claim for the users database if I'm not mistaken so I'd rather call it something like `{{ template "authelia.fullname" . }}-userdb`. What do you think?

helm/values.yaml
```diff
@@ -0,0 +1,256 @@
...
+      #     - dev
+
+  # Default policy can either be `bypass`, `one_factor`, `two_factor` or `deny`.
+  # It is the policy applied to any resource if there is no policy to be applied
+  # to the user.
+  accessControl:
+    defaultPolicy: deny
+    rules:
+      # Rules applied to everyone
+      - domain: public.example.com
```
clems4ever: As I said earlier, I think this part could probably be integrated in the ingress configuration to improve integration with Kubernetes. That would make a lot of sense to define everything related to the endpoint in a single place.

helm/values.yaml
```diff
@@ -0,0 +1,256 @@
...
+      groups_filter: (&(member={dn})(objectclass=groupOfNames))
+      # The attribute holding the name of the group
+      group_name_attribute: cn
+      # The attribute holding the mail address of the user
+      mail_attribute: mail
+      # The username and password of the admin user.
+      user: cn=admin,dc=example,dc=com
+      password: password
+
+    # If LDAP is disabled, you can specify a users database file, however
```
clems4ever: I know understand your concerns about the read-only characteristic of the file. Let's takle that later in v4.

helm/templates/configmap.yaml
ishioni: I'm not sure I follow, but are you talking about consuming annotations in ingresses? If so, then yeah, that's a nice idea, but on the general people are starting to move away from annotating and into CRDs where this kind of stuff should belong. Case in point : traefik 2.0 or istio

helm/templates/deployment.yaml
ishioni: If you look at values.yaml, in the notifier, it's empty by default, and authelia treats that as leaving fake emails as text files in /tmp/authelia. If it's not an emptydir, that directory is read-only.

helm/templates/pvc-config.yaml
ishioni: Actually I'm placing the user database in /var/lib/authelia as of now. But that can change, and I'd rather call it -config. This is purely internal anyway, the user never touches or changes this

helm/values.yaml
ishioni: I mean it's a neat idea, but rules can be quite complicated, and annotations are supposed to be one line each. A CRD would be a better (but more complicated!) approach

helm/templates/configmap.yaml
clems4ever: Good idea.

helm/templates/deployment.yaml
clems4ever: aaaah it's for fake emails, did not quite get it when I read. Ok, it makes sense.

helm/values.yaml
clems4ever: Yep, CRD is definitely something we can consider in the future.

clems4ever:
> I still need to write a proper Readme.md with available variables as other charts. So it's not quite ready yet

Yep, this is what I was about to ask and a link from the existing documentation to that readme.

helm/templates/pvc-config.yaml
clems4ever: I'm still a bit confused because there is no usage of this claim in the rest of the Chart unlike the "-store" claim. Is it expected?

helm/templates/deployment.yaml
clems4ever: makes sense

clems4ever:
I'll give the chart a try on my own cluster by following your documentation, hopefully by the end of the week. That's gonna be a good feedback for you since I'm a newbie in Helm chart deployment. I understand the concept and I did it once if I remember correctly but I don't remember anything :).

helm/templates/pvc-config.yaml
ishioni: Yeah, this was made when I thought I could do a cool trick with the userlists. It's gone now.

ishioni:
Did I mention i hate writing these Readme.md files with a passion? I'll try to get it finished this weekend

helm/templates/deployment.yaml
```diff
@@ -0,0 +1,95 @@
...
+            {{- toYaml .Values.resources | nindent 12 }}
+          volumeMounts:
+          - name: authelia-config
+            mountPath: /etc/authelia/config.yml
+            subPath: config.yml
+          {{- if not .Values.mongodb.enabled }}
+          - name: authelia-store
+            mountPath: /var/lib/authelia
+          {{- end }}
+          {{- if not .Values.authelia.ldap.enable }}
```
primetoxinz: ldap.enable -> ldap.enabled

helm/templates/deployment.yaml
```diff
@@ -0,0 +1,95 @@
...
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "authelia.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- if not .Values.authelia.ldap.enable }}
```
primetoxinz: ldap.enable -> ldap.enabled

helm/templates/deployment.yaml
```diff
@@ -0,0 +1,95 @@
...
+      volumes:
+      - name: authelia-config
+        configMap:
+          name: {{ template "authelia.fullname" . }}-config
+      {{- if not .Values.mongodb.enabled }}
+      - name: authelia-store
+        persistentVolumeClaim:
+          claimName: {{ if .Values.persistence.existingClaim }}{{ .Values.persistence.existingClaim }}{{- else }}{{ template "authelia.fullname" . }}-store{{- end }}
+      {{- end }}
+      {{- if not .Values.authelia.ldap.enable }}
```
primetoxinz: ldap.enable -> ldap.enabled

ishioni: Nice catch! Fixed

nightah:
What is still outstanding to move this along?

@clems4ever at what point do you plan to orphan v3? I also did notice the task for creating a helm chart for v4 is this something you're working on actively?

If not, @ishioni given #569 has now been merged do you think you have the desire/bandwidth to look at adapting these changes for v4?

clems4ever:
> @clems4ever at what point do you plan to orphan v3? I also did notice the task for creating a helm chart for v4 is this something you're working on actively?

I'm not investing any time on v3 anymore. I'm starting working on the Helm chart for v4 this evening by taking this PR as a starting point.

ishioni:
@nightah I wanted to include a Readme.md comparable to the helm public charts. However, I can now see that helm itself doesn't really want new charts anyway, and they're moving to a more decentralized model, where hub.helm.sh is pulling in external repos. The Readme should still be necessary, but I think the onus of creating a helm repo would now be on the authelia project itself

As for making a v4 chart, unfortunately I now have more personal commitments and will be unable to contribute for at least a year. Sorry about that :(

nightah:
@ishioni I'm going to close this off in the interim because from a project standpoint v3 is basically orphaned. All development; features, bugfixes and security patches will be aimed at v4+.

I know that @clems4ever was looking at the Helm chart for v4 based on your contribution, perhaps it's best to wait for this to use alongside v4.

Thanks again for your efforts here and we hope to see you contribute to Authelia in future when you have free time.

Ornias1993:
@nightah Please tell me where i'm wrong, but isn't this closed a little too easy?
I don't see any significant config syntax changes between 3 and 4, which means this chart might be 80% usefull and would just need a serieus update and a branch change to v4.
(untill someone else takes over and finishes it)

@clems4ever have you started to work on updating this and if so, could you at least publish what you have and what you haven't done yet?

PR #461, opened 2019-11-28
Publish additional minor and major tags on DockerHub

### LINKED ISSUE #456: Create multiple Docker image tags for major, minor and patches
Many images in Dockerhub have 3 different tags for a given version. Let's take an example, let say the version of myapp is v1.2.3.

Then, three tags are published: myapp:1 myapp:1.2 and myapp:1.2.3.
Those images are updated according to the new released versions. For instance a patch should target myapp:1 and myapp:1.2 and create myapp:1.2.4 while a minor change should only target myapp:1 and create two new tags.

nightah:
Per description.

Closes #456.

### DIFF
--- a/cmd/authelia-scripts/cmd_docker.go
+++ b/cmd/authelia-scripts/cmd_docker.go
@@ -15,6 +15,12 @@ import (
 var arch string
 var supportedArch = []string{"amd64", "arm32v7", "arm64v8"}
 var defaultArch = "amd64"
+var travisBranch = os.Getenv("TRAVIS_BRANCH")
+var travisPullRequest = os.Getenv("TRAVIS_PULL_REQUEST")
+var travisTag = os.Getenv("TRAVIS_TAG")
+var dockerTags = regexp.MustCompile(`(?P<Minor>(?P<Major>v\d+)\.\d+)\.\d+.*`)
+var ignoredSuffixes = regexp.MustCompile("alpha|beta")
+var tags = dockerTags.FindStringSubmatch(travisTag)
 
 func init() {
 	DockerBuildCmd.PersistentFlags().StringVar(&arch, "arch", defaultArch, "target architecture among: "+strings.Join(supportedArch, ", "))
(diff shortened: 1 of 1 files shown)


cmd/authelia-scripts/cmd_docker.go
```diff
@@ -177,17 +182,21 @@ func deployManifest(docker *Docker, tag string, amd64tag string, arm32v7tag stri
...
 
-	travisBranch := os.Getenv("TRAVIS_BRANCH")
-	travisPullRequest := os.Getenv("TRAVIS_PULL_REQUEST")
-	travisTag := os.Getenv("TRAVIS_TAG")
-
 	if travisBranch == "master" && travisPullRequest == "false" {
 		login(docker)
 		deploy(docker, "master-"+arch)
 	} else if travisTag != "" {
+		patchTag := publishedTags.FindStringSubmatch(travisTag)[1]
```
clems4ever: Test the array size before doing unchecked array accesses. The best way would be to call publishedTags.FindStringSubmatch(travisTag) only once and reuse the result both for checking the length and accessing the elements. If the size of the array is wrong => log.Fatal

cmd/authelia-scripts/cmd_docker.go
```diff
@@ -196,18 +205,19 @@ func publishDockerImage(arch string) {
...
-	travisBranch := os.Getenv("TRAVIS_BRANCH")
-	travisPullRequest := os.Getenv("TRAVIS_PULL_REQUEST")
-	travisTag := os.Getenv("TRAVIS_TAG")
-	ignoredSuffixes := regexp.MustCompile("alpha|beta")
-
 	if travisBranch == "master" && travisPullRequest == "false" {
 		login(docker)
 		deployManifest(docker, "master", "master-amd64", "master-arm32v7", "master-arm64v8")
 	} else if travisTag != "" {
+		patchTag := publishedTags.FindStringSubmatch(travisTag)[1]
```
clems4ever: Same comment as above.

clems4ever:
Thanks for the PR!

PR #462, opened 2019-11-28
Publish binary artifacts to GitHub

### LINKED ISSUE #457: Publish binary version of Authelia in Github artifacts

nightah:
You need to configure your `GITHUB_API_KEY` in [Travis](https://stackoverflow.com/questions/23277391/how-to-publish-to-github-pages-from-travis-ci).

Closes #457.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,6 @@ jobs:
     - stage: test
       script:
         - authelia-scripts --log-level debug ci
-      # TODO(c.michaud): publish built artifact on Github.
     - &build-images
       stage: build images
       env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -50,6 +49,17 @@ jobs:
         - kill %1
       after_success:
         - authelia-scripts docker push-image --arch=$ARCH
+        - if [[ $TRAVIS_TAG ]]; then docker create --name authelia-binary clems4ever/authelia:$TRAVIS_TAG-$ARCH && docker cp authelia-binary:/usr/app/authelia ./authelia-linux-$ARCH; fi
```
clems4ever: It's not gonna be sufficient, we need to tar.gz the binary along with the compiled frontend until we have a better solution for packaging the frontend.

.travis.yml
```diff
@@ -50,6 +49,17 @@ jobs:
...
+      before_deploy:
+        - echo Deploying Authelia $TRAVIS_TAG-$ARCH to GitHub Releases
+      deploy:
+        provider: releases
+        api_key: '$GITHUB_API_KEY'
```
clems4ever: I will create the API key and set it in Travis.

clems4ever: Done

.travis.yml
nightah: You're absolutely right here, that's fine I'll grab the `public_html` folder and tar it up with the binary and adjust accordingly.

.travis.yml
nightah: Great.

PR #468, opened 2019-12-01
Optimise deploy steps for Github releases

nightah:


nightah:
Rebased, failure seems unrelated.

PR #470, opened 2019-12-04
Rewrite Authelia frontend to prepare for administration pages

clems4ever:
I also took the opportunity to rewrite all suites in Go.


.travis.yml
```diff
@@ -19,30 +19,79 @@ addons:
       - libgif-dev
       - google-chrome-stable
 
-install: # Install ChromeDriver (64bits; replace 64 with 32 for 32bits).
-  - wget -N https://chromedriver.storage.googleapis.com/78.0.3904.70/chromedriver_linux64.zip -P ~/
-  - unzip ~/chromedriver_linux64.zip -d ~/
-  - rm ~/chromedriver_linux64.zip
-  - sudo mv -f ~/chromedriver /usr/local/share/
```
nightah: Could this not just become `sudo mv -f ~/chromedriver /usr/local/bin/` and then perform the `chmod -x`?
That should be in the `$PATH` env var right, means you can get rid of the symbolic link too.

.travis.yml
```diff
@@ -19,30 +19,79 @@ addons:
...
+      env:
+        - SUITE_NAME=BypassAll
+      before_script:
+        # Install chrome driver
+        # TODO(c.michaud): this could be done in authelia-scripts instead for devs to not do the install themselves.
+        #                  or even provide a docker image with a selenium server.
+        - wget -N https://chromedriver.storage.googleapis.com/78.0.3904.70/chromedriver_linux64.zip -P ~/
+        - unzip ~/chromedriver_linux64.zip -d ~/
+        - rm ~/chromedriver_linux64.zip
+        - sudo mv -f ~/chromedriver /usr/local/share/
```
nightah: Could this not just become `sudo mv -f ~/chromedriver /usr/local/bin/` and then perform the `chmod -x`?
That should be in the `$PATH` env var right, means you can get rid of the symbolic link too.

PR #486, opened 2019-12-07
Fix README.md publication and tag cleanup for DockerHub

nightah:


clems4ever:
#489 should fix the issue with the tests

clems4ever:
Please rebase, it should fix your issue.

Issue #466, opened 2019-11-30
nginx config files


Iwan-Authelia:
Hi,

I've been testing with Authelia and have some issues getting the nginx config correct.

What I am trying to do is the following;

1. Hosting Authelia on a publicly accessible server
2. When 2FA is successful calling a proxy-pass to an internal webserver which hosts the actual website.

As there are multiple examples for nginx configs here I have tried all or at least most of them but so far I have been unable to get it all working.

I've attached a zip file with my configs (Authelia and nginx, anonymised)
in etc/nginx/sites-available/reverse-proxy.conf is a working config for a not-protected proxy-pass. The purpose is to protect this by Authelia.

[authelia-test-config.zip](https://github.com/clems4ever/authelia/files/3906766/authelia-test-config.zip)

I post it here for two reasons, first, I'm stuck and hope for some help :) Second, I hope to help others getting this up and running by sharing some working configs.

nightah:
I've included an extract of my nginx config below.

This includes, the Authelia application hosted in Docker as the container named `authelia` and published at `auth.example.com` an example of Nextcloud hosted in Docker as the container named `nextcloud` and published at `nextcloud.example.com` and the respective additional nginx configuration.

`authelia.conf` is included on every host as part of the template, however, I only include `auth.conf` for endpoints that need protection and have been configured in my Authelia `config.yml`. The `proxy.conf` is included whenever reverse proxying.

**auth.example.com**
```nginx
server {
    server_name auth.example.com;
    listen 80;
    return 301 https://$server_name$request_uri;
}

server {
    server_name auth.example.com;
    listen 443 ssl http2;
    include /config/nginx/ssl.conf;

    location / {
        set $upstream_authelia http://authelia:8080;
        proxy_pass $upstream_authelia;
        include /config/nginx/proxy.conf;
    }
}
```

**nextcloud.example.com**
```nginx
server {
    server_name nextcloud.example.com;
    listen 80;
    return 301 https://$server_name$request_uri;
}

server {
    server_name nextcloud.example.com;
    listen 443 ssl http2;
    include /config/nginx/ssl.conf;
    include /config/nginx/authelia.conf;

    location / {
        set $upstream_nextcloud https://nextcloud;
        proxy_pass $upstream_nextcloud;
        include /config/nginx/auth.conf;
        include /config/nginx/proxy.conf;
    }
}
```

**authelia.conf**
```nginx
location /authelia {
    internal;
[...]

PR #487, opened 2019-12-08
authelia-from-scratch.md

Iwan-Authelia:
How to setup Authelia from scratch on Bare Metal.


docs/authelia-from-scratch.md
```diff
@@ -0,0 +1,448 @@
+# Installing **Authelia** on bare metal from scratch
```
clems4ever: Please specify it's for debian in the title.

docs/authelia-from-scratch.md
```diff
@@ -0,0 +1,448 @@
...
+            }
+    }
+    
+    server {
+            server_name authelia.example.com;
+            listen 443 ssl http2;
+            include ssl.conf;
+    
+            location / {
+                    add_header X-Forwarded-Host authelia.example.com;
```
clems4ever: missing add_header X-Forwarded-Proto $scheme;

docs/authelia-from-scratch.md
```diff
@@ -0,0 +1,448 @@
...
+    ssl_session_timeout 24h;
+    ssl_session_cache shared:SSL:50m;
+    ssl_session_tickets off;
+    EOF
+
+### (re)Start NGINX
+    # ln sites-available/authelia.conf sites-enabled/
+    # systemctl restart nginx
+
+### Install Docker
```
clems4ever: You can add a section where the user can download the artifact from github and verify the shasum instead and mark the following sections as "optional" up to "Setup Authelia". This is easier and does not require to install additional packages.

Those steps are only required if one needs to build for a given non-built target.

docs/authelia-from-scratch.md
```diff
@@ -0,0 +1,448 @@
...
+    # mkdir /etc/authelia/
+    # cat <<EOF > /etc/authelia/config.yml
+    ###############################################################
+    #                   Authelia configuration                    #
+    ###############################################################
+    
+    port: 9091
+    
+    logs_level: info
+    jwt_secret: a_secret
```
clems4ever: This should absolutely be changed too. Why not use $jwt_secret as for other secrets ?

docs/authelia-from-scratch.md
```diff
@@ -0,0 +1,448 @@
...
+        mail_attribute: mail
+        user: cn=admin,dc=example,dc=com
+        password: $slapd-admin-password
+    
+    access_control:
+      default_policy: two_factor
+    
+    session:
+      name: authelia_session
+      secret: unsecure_session_secret
```
clems4ever: same remark as above, why not use $session_secret as for other secrets?

docs/authelia-from-scratch.md
Iwan-Authelia: Done :)

docs/authelia-from-scratch.md
Iwan-Authelia: Done :)

docs/authelia-from-scratch.md
Iwan-Authelia: Good idea! Done :)

docs/authelia-from-scratch.md
```diff
@@ -0,0 +1,461 @@
...
+### Install Debian the latest version of Debian 10
+    Debian Network install from a minimal CD: https://www.debian.org/CD/netinst/
+    
+### Uninstall apparmor and install your favourite editor if not already installed
+    # apt-get remove apparmor
+    # apt-get install $editor gcc g++ make slapd ldap-utils redis-server curl mariadb-server nginx git apt-transport-https ca-certificates gnupg2 software-properties-common ntpdate
+
+**NOTE** You will be prompted to set the $slapd-admin-password
+
+#### Install GO;
```
clems4ever: Do you really need that at this point if the user does not build from source?

docs/authelia-from-scratch.md
```diff
@@ -0,0 +1,461 @@
...
+**NOTE** First copy the /etc/letsencrypt/live/example.com directory from the location where you have letsencrypt running
+
+**NOTE** All installation steps below are to be executed as root or as a normal user using 'sudo'
+
+### Install Debian the latest version of Debian 10
+    Debian Network install from a minimal CD: https://www.debian.org/CD/netinst/
+    
+### Uninstall apparmor and install your favourite editor if not already installed
+    # apt-get remove apparmor
+    # apt-get install $editor gcc g++ make slapd ldap-utils redis-server curl mariadb-server nginx git apt-transport-https ca-certificates gnupg2 software-properties-common ntpdate
```
clems4ever: Are gcc, g++ and make required if the user does not build from source?

docs/authelia-from-scratch.md
```diff
@@ -0,0 +1,461 @@
...
+The following passwords are needed;
+
+    slapd-admin-password=<slapd-admin-password>
+    redis-password=<redis-password>
+    mariadb-password=<mariadb-password>
+    user-password=<user-password>
+    jwtsecret=<jwt-secret>
+    sessionsecret=<session-secret>
+    
+These packages are optional and only needed if you are planning to compile from source;
```
clems4ever: Good!

docs/authelia-from-scratch.md
clems4ever: Maybe we can mark it as optional in that case. What do you think?

docs/authelia-from-scratch.md
```diff
@@ -0,0 +1,461 @@
...
+                    proxy_pass $upstream_target;
+                    include auth.conf;
+                    include proxy.conf;
+                    #proxy_ssl_certificate /etc/nginx/ssl/xxx.example.com.cert;
+                    #proxy_ssl_certificate_key /etc/nginx/ssl/xxx.example.com.key;
+            }
+    }
+    EOF
+
+**NOTE** If protected.example.com needs a different SSL certificate than landing.example.com you can un-comment the two proxy_ssl_certificate lines and point them to the correct certificates. It is even possible to have protected.example.com on a completely different domain-name but that is not guaranteed to work without issues.
```
clems4ever: Having protected.example.com on a completely different domain than Authelia does not work since the cookie produced by Authelia won't be sent when visiting the site. So even if the user is authenticated, he will be redirected to Authelia portal.

I think you can safely remove the line or just tell that it does not work. It will avoid people thinking it could work, as it does not for sure.

docs/authelia-from-scratch.md
clems4ever: Add a mark like '(for building from source only)' in the title to clearly identify what is really optional.

clems4ever: You can do the same for all optional sections.

docs/authelia-from-scratch.md
Iwan-Authelia: It actually works, I have a proof of concept :)

docs/authelia-from-scratch.md
Iwan-Authelia: Not really, I'll add the to the optional list

docs/authelia-from-scratch.md
Iwan-Authelia: It's already on the optional list :)

clems4ever:
Can you rename authelia-from-scratch.md into authelia-on-baremetal.md?
Do we also include some pointers from the existing docs?

docs/authelia-from-scratch.md
nightah: @clems4ever different SSL certs doesn't necessarily mean different domains.

@Iwan-Authelia `protected.example.com` and `landing.example.com` are subdomains on the same domain still, so if your proof of concept is in reference to something similar then the statement by Clèment is still correct.

davama:
@Iwan-Authelia 
this is a cool doc!
Currently going through (testing) it myself.

Minor detail:
systemd file, why not add `WorkingDirectory=/opt/authelia/dist` instead of having a shell script?

my .02 cents

Regards

docs/authelia-on-baremetal.md
```diff
@@ -0,0 +1,462 @@
...
+    # cd /etc/nginx
+    # cat <<EOF > auth.conf
+    # Basic Authelia Config
+    auth_request /authelia;
+    auth_request_set $target_url $scheme://$http_host$request_uri;
+    auth_request_set $user $upstream_http_remote_user;
+    auth_request_set $groups $upstream_http_remote_groups;
+    proxy_set_header X-Forwarded-User $user;
+    proxy_set_header X-Forwarded-Groups $groups;
+    error_page 401 =302 https://authelia.example.com/#/?rd=$target_url;
```
davama: I have been testing your doc and i could not get redirection to work.
Looking around I found this: https://github.com/authelia/authelia/issues/466#issuecomment-566786122

```
error_page 401 =302 https://authelia.example.com/?rd=$target_url;
```
After changing my auth.conf line to the above did redirection work.

Still playing with it but just FYI

Regards

nightah:
@Iwan-Authelia do you have the time/bandwidth to deal with the requested changes and then we can look to include this in `https://docs.authelia.com`?

Iwan-Authelia:
@nightah Sorry, I have never used github before and fail to understand how this system works.

I would be happy for anyone else to finish this.

nightah:
Closing this off for now, might look to include it in the community section of the docs time permitting though.

Thanks @Iwan-Authelia.

PR #506, opened 2019-12-18
Add Buildkite CI tooling

nightah:
This commit is in prep for setting up Buildkite as the future CI tool for Authelia.

@clems4ever a couple things to be done on the Authelia DockerHub Org because I don't seem to have the required access but we can discuss that on Gitter.

Credits to the LSIO team for their base-image and overlay where a lot of this has come from.
I'll add a `README.md` for how to run the container too which will also give credits to the LSIO team.


ci/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+   apk add --no-cache \
+     bash \
+     ca-certificates \
+     coreutils \
+     chromium \
+     chromium-chromedriver \
+     curl \
+     docker-compose \
+     git \
+     hub@edget \
```
clems4ever: why is it edget instead of edge?

ci/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+   tar xfz s6-overlay.tar.gz -C / && \
+ echo "**** Patch CVE-2019-5021 ****" && \
+   sed -i -e 's/^root::/root:!:/' /etc/shadow && \
+ echo "**** Create abc user and make our folders ****" && \
+   useradd -u 911 -U -d /buildkite -s /bin/false abc && \
+   usermod -G wheel abc && \
+   sed -i 's/# %wheel/%wheel/g' /etc/sudoers && \
+ echo "**** Install Buildkite ****" && \
+   mkdir -p /buildkite/builds /buildkite/hooks /buildkite/plugins && \
+   curl -Lfs -o /usr/local/bin/ssh-env-config.sh https://raw.githubusercontent.com/buildkite/docker-ssh-env-config/master/ssh-env-config.sh && \
```
clems4ever: please fix the version by using a tag or branch instead of master. This will prevent breaking in the future.

ci/root/etc/docker/daemon.json
```diff
@@ -0,0 +1,3 @@
+{
+  "registry-mirrors": ["http://registrycache:5000"]
```
clems4ever: How is registrycache service setup? I don't see any reference except here in that PR.

ci/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+     tzdata \
+     yarn@edge && \
+ echo "**** Add s6 overlay ****" && \
+   cd /tmp && \
+   curl -Lfs -o s6-overlay.tar.gz "https://github.com/just-containers/s6-overlay/releases/download/${OVERLAY_VERSION}/s6-overlay-${ARCH}.tar.gz" && \
+   tar xfz s6-overlay.tar.gz -C / && \
+ echo "**** Patch CVE-2019-5021 ****" && \
+   sed -i -e 's/^root::/root:!:/' /etc/shadow && \
+ echo "**** Create abc user and make our folders ****" && \
+   useradd -u 911 -U -d /buildkite -s /bin/false abc && \
```
clems4ever: put a more explicit username than abc. Something like "builder", "buildkite", "executor" or a composition would fit.

ci/Dockerfile
nightah: Because of this:
https://github.com/clems4ever/authelia/blob/4bd491c3ce9b4cf6e87837b24c2be48aa033daa0/ci/Dockerfile#L26-L27
some of the packages are in different edge repos so `@edge` referers to edge/community and `@edget` refers to edge/testing.

[Hub](https://pkgs.alpinelinux.org/packages?name=hub&branch=edge) specifically resides in edge/testing.

ci/Dockerfile
nightah: This is the same as upstream's configuration for how they build their own docker container:
https://github.com/buildkite/agent/blob/f51dc916c5279d72e541c5b47b48159668ac6dea/packaging/docker/alpine-linux/Dockerfile#L24-L26

I also have notifications on when the upstream repo performs any commits to be able to track and monitor this.

ci/root/etc/docker/daemon.json
nightah: The registrycache is another separate container that needs to be run, if the local docker service cannot hit the registry cache it will automatically go to DockerHub as if the cache did not exist, so it's transparent.

I will provide a `docker-compose.yml` which includes strings for both the cache and the respective nodes.
There's also some files you will need to map in to the buildkite node containers which I will address in the `README.md`.

nightah: I've just added an example `docker-compose.yml` and a `README.md` on usage.

ci/Dockerfile
clems4ever: Human monitoring is not reliable. I'm ok to try like that but we'll fix the version it if breaks too often.

ci/Dockerfile
clems4ever: perfect, thanks.

ci/README.md
```diff
@@ -0,0 +1,99 @@
+[logo]: https://github.com/clems4ever/authelia/raw/master/docs/images/authelia-title.png "Authelia"
+[![alt text][logo]](https://github.com/AntilaX-3/)
```
clems4ever: Why is the logo linked to this URL? Can you please either refer to authelia homepage or just remove it?

nightah: Yep remnants from my other repo I copied the `README.md` from, fixed now though.

PR #507, opened 2019-12-18
Utilise Buildkite for Authelia CI/CD

nightah:
This change looks to replace Travis with Buildkite for Authelia's CI/CD pipeline.

It's worth calling out this PR is not dependent on #506, mainly due to the fact that I still have some workers in place. However, my recommendation is that we merge #506 first and ensure that we have our respective workers setup then we clean the Buildkite Org and then look to merge this change in.

Examples of a non-tagged and tagged commit respectively:

**non-tagged**
https://buildkite.com/authelia/authelia-dev/builds/319
https://buildkite.com/authelia/authelia-dev/builds/323

**tagged**
https://buildkite.com/authelia/authelia-dev/builds/320
https://buildkite.com/authelia/authelia-dev/builds/324

**dockerhub**
https://hub.docker.com/r/nightah/authelia

I've left the commit history for now, happy to squash prior to finalising and pulling in to master.
Let's catch up on Gitter to finalise the steps to get this over the line.


.buildkite/steps/e2etests.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -eu
+
+for SUITE_NAME in BypassAll Docker DuoPush HighAvailability Kubernetes LDAP Mariadb NetworkACL Postgres ShortTimeouts Standalone Traefik;
```
clems4ever: You can have the list of suites dynamically generated from authelia-scripts with command `authelia-scripts suites list`

Dockerfile
```diff
@@ -42,7 +42,7 @@ WORKDIR /node/src/app
 COPY web .
 
 # Install the dependencies and build
-RUN npm ci && npm run build
+RUN yarn install && yarn build
```
clems4ever: What is the rationale for moving from npm to yarn?

bootstrap.sh
```diff
@@ -1,16 +1,29 @@
...
 if [ -z "$OLD_PS1" ]; then
   OLD_PS1="$PS1"
   export PS1="(authelia) $PS1"
 fi
 
-export USER_ID=$(id -u)
-export GROUP_ID=$(id -g)
-export CI=false
+if [ $(id -u) = 0 ]; then
+  export USER_ID=1000
```
clems4ever: Put a warning log here saying that the uid of the user cannot be root and defaults to 1000.

bootstrap.sh
```diff
@@ -1,16 +1,29 @@
...
+else
+  export USER_ID=$(id -u)
+fi
+
+if [ $(id -g) = 0 ]; then
+  export GROUP_ID=1000
```
clems4ever: Put a warning log here saying that the gid of the user cannot be root and defaults to 1000.

example/compose/authelia/Dockerfile.backend
```diff
@@ -1,6 +1,9 @@
 FROM golang:1.13-alpine
 
-RUN apk --no-cache add gcc musl-dev
+#TODO(nightah): Remove when turning off Travis
+ARG TRAVIS
+RUN if [ "$TRAVIS" == "true" ]; then apk --no-cache add gcc musl-dev; fi
```
clems4ever: Why don't we need to install gcc and the libc in buildkite setup?

Dockerfile
nightah: Everything I've read suggested that yarn is faster than npm and that it's the natural successor of npm. Happy to change it back if we're overly concerned.

example/compose/authelia/Dockerfile.backend
nightah: Because we aren't building the binaries again in Buildkite, they're shared from the build step to speed up the tests.

https://github.com/authelia/authelia/blob/37c6c4383e18af2894b3c4e4dc6e8c4bcda0885b/.buildkite/pipeline.yml#L6-L8

nightah:
@clems4ever I think this now should be ready to merge, let me know if there's anything else otherwise I'll squash and merge.

Dockerfile
clems4ever: it's fine, just wanted to confirm.

example/compose/authelia/Dockerfile.backend
clems4ever: Ah yes, got it!

PR #509, opened 2019-12-19
Fix PostgreSQL Update Second Factor Method Pref

james-d-elliott:
- column name is second_factor_method, not method


james-d-elliott:
Hold off on this, I should have tested it. It doesn't fix it, oddly.

james-d-elliott:
Fixed. It was an issue with the VARCHAR length not being long enough for the "mobile_push" value when you pick the DUO method.

clems4ever:
Thank you for the patch @james-d-elliott !

PR #514, opened 2019-12-20
Implement SMTP StartTLS and Adaptive Auth

### LINKED ISSUE #464: Authelia v4 is unable to send email via Office365 SMTP
Hi,

I've used the same config and ldap as for V3 and got this error;

ERRO[0037] Error while checking password for user iwan: Cannot find user DN of user iwan. Cause: LDAP Result Code 201 "Filter Compile Error": ldap: filter does not start with an '('  method=POST path=/api/firstfactor remote_ip=127.0.0.1 

Ldap config;

authentication_backend:
  ldap:
    url: ldap://127.0.0.1
    base_dn: dc=example,dc=com
    additional_users_dn: ou=people
    users_filter: uid={0}
    additional_groups_dn: ou=groups
    groups_filter: (&(member={dn})(objectclass=groupOfNames))
    group_name_attribute: cn
    mail_attribute: mail
    user: cn=admin,dc=example,dc=com
    password: xxxx

root@auth:/usr/src# ldapwhoami -vvv -h 127.0.0.1 -D "uid=iwan,ou=people,dc=example,dc=com" -x -W
ldap_initialize( ldap://127.0.0.1 )
Enter LDAP Password:
dn:uid=iwan,ou=people,dc=example,dc=com
Result: Success (0)
root@auth:/usr/src#

james-d-elliott:
- If the StartTLS extension is advertised we automatically StartTLS
- Uses the secure config key to determine if we should verify the cert
- Attempt auth when the config has a SMTP password and the server supports the AUTH extension
- Check the mechanisms supported by the server and use PLAIN, LOGIN, or CRAM-MD5 depending on which is supported
- Changed secure key to use boolean values instead of strings
- Arranged SMTP notifier properties/vars to be in the same order

Fixes #464 

I've built and published an image on docker hub: [jamesdelliott/authelia:smtp-adaptive-auth](https://hub.docker.com/layers/jamesdelliott/authelia/smtp-adaptive-auth/images/sha256-d07bc0e6a4068d398f9d2b8d905d251fe062a07d4fd244cf857a180fb3bfd4d0)

Requires Testing:
- [x] SMTP Server that doesn't Support StartTLS (should succeed)
- [x] SMTP Server that does support StartTLS (should succeed)
- [x] SMTP Server that doesn't support AUTH (should succeed)
- [x] SMTP Server that only supports AUTH CRAM-MD5 (should succeed)
- [x] SMTP Server that only supports AUTH PLAIN (should succeed)
- [x] SMTP Server that only supports AUTH NTLM (should fail with error)
- [x] SMTP Server that only supports AUTH LOGIN (should succeed)
- [x] SMTP Server that supports AUTH PLAIN and AUTH LOGIN (should use AUTH PLAIN)
- [x] SMTP Server with bad cert and secure set to False (should succeed)
- [x] SMTP Server with bad cert and secure set to True (should fail)
- [x] SMTP Server with bad cert and secure commented out, or unset (should succeed)
- [x] SMTP Server with good cert and secure set to true (should succeed)
- [x] SMTP Server with good cert and secure commented out, or unset (should succeed, wont verify cert)
- [x] SMTP Server that only offers AUTH LOGIN after STARTTLS

Things I plan to add to this PR:
- [x] Debug Logging for STARTTLS
- [x] Debug Logging for AUTH
- [x] Rebase / Squash
- [x] Finalize Spec

## Final Specification Tests

Tests unrelated to password/auth (should be possible in test suite):
- [x] Server that doesn't support STARTLS (should succeed)
- [x] Server that does support STARTLS (should succeed)
- [x] Server with self-signed cert and secure set to False (should succeed)
- [x] Server with self-signed cert and secure set to True (should fail)
- [x] Server with self-signed cert and secure commented/unset (should succeed)
- [x] Server with good cert and secure set to True (should succeed)
- [x] Server with good cert and secure commented/unset (should succeed)

Tests WITH password set in config (should be possible in test suite):
- [x] Server that doesn't support AUTH (should fail)
- [x] Server that only supports an AUTH mechanism NOT supported by Authelia (should fail)
- [x] Server that only supports AUTH PLAIN (should succeed)
- [x] Server that only supports AUTH LOGIN (should succeed)
- [x] Server that only offers AUTH LOGIN after STARTTLS (should succeed)

Tests WITHOUT password set in config (should be possible in test suite):
- [x] Server that doesn't support AUTH (should succeed)
- [x] Server that has optional AUTH (should succeed)
- [x] Server that has required AUTH (should fail)

Manual tests (probably not possible in test suite):
- [x] Server that supports AUTH PLAIN and AUTH LOGIN (should use AUTH PLAIN)

### DIFF
--- a/internal/notification/smtp_notifier.go
+++ b/internal/notification/smtp_notifier.go
@@ -1,50 +1,110 @@
 package notification
 
 import (
+	"crypto/tls"
+	"errors"
 	"fmt"
 	"net/smtp"
+	"strings"
 
 	"github.com/authelia/authelia/internal/configuration/schema"
+	"github.com/authelia/authelia/internal/utils"
+	log "github.com/sirupsen/logrus"
 )
 
 // SMTPNotifier a notifier to send emails to SMTP servers.
 type SMTPNotifier struct {
-	address  string
-	sender   string
 	username string
 	password string
+	sender   string
 	host     string
 	port     int
+	secure   bool
+	address  string
 }
 
 // NewSMTPNotifier create an SMTPNotifier targeting a given address.
 func NewSMTPNotifier(configuration schema.SMTPNotifierConfiguration) *SMTPNotifier {
 	return &SMTPNotifier{
-		host:     configuration.Host,
-		port:     configuration.Port,
-		address:  fmt.Sprintf("%s:%d", configuration.Host, configuration.Port),
-		sender:   configuration.Sender,
 		username: configuration.Username,
 		password: configuration.Password,
+		sender:   configuration.Sender,
+		host:     configuration.Host,
+		port:     configuration.Port,
+		secure:   configuration.Secure,
+		address:  fmt.Sprintf("%s:%d", configuration.Host, configuration.Port),
 	}
 }
 
-func (n *SMTPNotifier) authenticatedSend(recipient string, msg string) error {
-	auth := smtp.PlainAuth("", n.username, n.password, n.host)
-	err := smtp.SendMail(fmt.Sprintf("%s:%d", n.host, n.port), auth, n.sender,
-		[]string{recipient}, []byte(msg))
-	if err != nil {
-		return err
-	}
-	return nil
-}
+// Send send a identity verification link to a user.
+func (n *SMTPNotifier) Send(recipient string, subject string, body string) error {
+	msg := "From: " + n.sender + "\n" +
+		"To: " + recipient + "\n" +
+		"Subject: " + subject + "\n" +
+		"Content-Type: text/html\n" +
+		"MIME-version: 1.0;\nContent-Type: text/html; charset=\"UTF-8\";\n\n" +
+		body
 
-func (n *SMTPNotifier) unauthenticatedSend(recipient string, msg string) error {
-	// Connect to the remote SMTP server.
 	c, err := smtp.Dial(n.address)
 
 	if err != nil {
 		return err
+	}
+
+	// Do StartTLS if available (some servers only provide the auth extnesion after, and encrpytion is preferred)
+	starttls, _ := c.Extension("STARTTLS")
+	if starttls {
+		tlsconfig := &tls.Config{
+			InsecureSkipVerify: !n.secure,
+			ServerName:         n.host,
+		}
+		log.Debugf("SMTP server supports STARTTLS (InsecureSkipVerify: %t, ServerName: %s), attempting", tlsconfig.InsecureSkipVerify, tlsconfig.ServerName)
+		err := c.StartTLS(tlsconfig)
+		if err != nil {
+			return err
+		} else {
+			log.Debug("SMTP STARTTLS completed without error")
+		}
+	} else {
+		log.Debug("SMTP server does not support STARTTLS, skipping")
+	}
+
+	// Attempt AUTH if password is specified only
+	if n.password != "" {
+
+		// Check the server supports AUTH, and get the mechanisms
+		authExtension, m := c.Extension("AUTH")
+		if authExtension {
+			log.Debugf("Config has SMTP password and server supports AUTH with the following mechanisms: %s.", m)
+			mechanisms := strings.Split(m, " ")
+			var auth smtp.Auth
+
+			// Adaptively select the AUTH mechanism to use based on what the server advertised
+			if utils.IsStringInSlice("PLAIN", mechanisms) {
+				auth = smtp.PlainAuth("", n.username, n.password, n.host)
+				log.Debug("SMTP server supports AUTH PLAIN, attempting...")
+			} else if utils.IsStringInSlice("LOGIN", mechanisms) {
+				auth = LoginAuth(n.username, n.password)
+				log.Debug("SMTP server supports AUTH LOGIN, attempting...")
+			}
+
+			// Throw error since AUTH extension is not supported
+			if auth == nil {
+				return fmt.Errorf("SMTP server does not advertise a AUTH mechanism that Authelia supports. Advertised mechanisms: %s.", m)
+			}
+
+			// Authenticate
+			err := c.Auth(auth)
+			if err != nil {
+				return err
+			} else {
+				log.Debug("SMTP AUTH completed successfully.")
+			}
+		} else {
+			return errors.New("SMTP server does not advertise the AUTH extension but a password was specified. Either disable auth (don't specify a password/comment the password), or specify an SMTP host and port that supports AUTH PLAIN or AUTH LOGIN.")
+		}
+	} else {
+		log.Debug("SMTP config has no password specified for use with AUTH, skipping.")
 	}
 
 	// Set the sender and recipient first
@@ -76,19 +136,4 @@ func (n *SMTPNotifier) unauthenticatedSend(recipient string, msg string) error {
 		return err
 	}
 	return nil
-}
-
-// Send send a identity verification link to a user.
-func (n *SMTPNotifier) Send(recipient string, subject string, body string) error {
-	msg := "From: " + n.sender + "\n" +
-		"To: " + recipient + "\n" +
-		"Subject: " + subject + "\n" +
-		"Content-Type: text/html\n" +
-		"MIME-version: 1.0;\nContent-Type: text/html; charset=\"UTF-8\";\n\n" +
-		body
-
-	if n.password != "" {
-		return n.authenticatedSend(recipient, msg)
-	}
-	return n.unauthenticatedSend(recipient, msg)
 }
--- a/internal/notification/smtp_notifier.go
+++ b/internal/notification/smtp_notifier.go
@@ -71,6 +71,9 @@ func (n *SMTPNotifier) Send(recipient string, subject string, body string) error
 
 	// Attempt AUTH if password is specified only
 	if n.password != "" {
+		if !starttls {
+			log.Warn("Authentication is being attempted over an insecure connection. Using a SMTP server that supports STARTTLS is recommended, especially if the server is not on your local network (username and pasword are being transmitted in plain-text).")
+		}
 
 		// Check the server supports AUTH, and get the mechanisms
 		authExtension, m := c.Extension("AUTH")
@@ -90,7 +93,7 @@ func (n *SMTPNotifier) Send(recipient string, subject string, body string) error
 
 			// Throw error since AUTH extension is not supported
 			if auth == nil {
-				return fmt.Errorf("SMTP server does not advertise a AUTH mechanism that Authelia supports. Advertised mechanisms: %s.", m)
+				return fmt.Errorf("SMTP server does not advertise a AUTH mechanism that Authelia supports (PLAIN or LOGIN). Advertised mechanisms: %s.", m)
 			}
 
 			// Authenticate
--- a/internal/configuration/schema/notifier.go
+++ b/internal/configuration/schema/notifier.go
@@ -17,10 +17,10 @@ type EmailNotifierConfiguration struct {
 type SMTPNotifierConfiguration struct {
 	Username string `yaml:"username"`
 	Password string `yaml:"password"`
-	Secure   string `yaml:"secure"`
+	Sender   string `yaml:"sender"`
 	Host     string `yaml:"host"`
 	Port     int    `yaml:"port"`
-	Sender   string `yaml:"sender"`
+	Secure   bool   `yaml:"secure"`
 }
 
 // NotifierConfiguration representes the configuration of the notifier to use when sending notifications to users.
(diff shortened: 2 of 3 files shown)


james-d-elliott:
Question. In a situation where the config defines a password, but the server does not support AUTH, do we want to throw an error or just skip AUTH and attempt to send the mail?

james-d-elliott:
Not sure how to debug the CI errors. Though nothing changed should cause issues with that suite.

nightah:
> Not sure how to debug the CI errors. Though nothing changed should cause issues with that suite. The fact the original commit passed the failed test suggests it was a false negative.

The issue was unrelated and seems to be a problem with Travis, I'll kick it back off in a second.

clems4ever:
Thanks @james-d-elliott for the contribution.

> Questions.
> 
> 1. In a situation where the config defines a password, but the server does not support AUTH, do we want to throw an error or just skip AUTH and attempt to send the mail like this: https://github.com/james-d-elliott/authelia/blob/feature/notification-smtp-adaptive-auth-and-starttls/internal/notification/smtp_notifier.go#L99-L101

I think we can throw an error. It might help to capture a misconfiguration in a SMTP server set up for Authelia.

> 2. Also in a situation where the SMTP server does not advertise auth mechanisms that Authelia supports do we want to skip AUTH or just return an error like this: https://github.com/james-d-elliott/authelia/blob/feature/notification-smtp-adaptive-auth-and-starttls/internal/notification/smtp_notifier.go#L90-L92

I was a bit concerned about the automatic upgrade from no auth to AUTH because I always wonder if the upgrade has truly been done. I don't really like what is implicit. I'd be glad to have an error but this means introducing a new configuration flag then. What do you think?

clems4ever:
Do you guys know how to configure a SMTP service in all those cases to introduce those use cases as integration tests? That would ensure non-regression for the future.
For instance, do you know how to configure postfix in all those cases?

nightah:
> Do you guys know how to configure a SMTP service in all those cases to introduce those use cases as integration tests? That would ensure non-regression for the future.
> For instance, do you know how to configure postfix in all those cases?

I can help with this part, NTLM obviously will require AD/Exchange but the other cases we can handle.

james-d-elliott:
I think for point 1 and 2 an error should be thrown. If authentication is configured you expect the server to support it. The way I've done STARTTLS also avoids the situation of servers that only offer auth mechanisms after STARTTLS. I think it's sufficient to set both 1 and 2 as failure conditions and is probably preferable, also I think that's what you're suggesting. 

As far as setup of a postfix server it should be easy. With the 4 auth mechanisms implemented it covers most configurations (anonymous, PLAIN, CRAM-MD5, LOGIN). I only tested NTLM to verify the expected behavior of specifically when an unsupported mech was the only one available.

james-d-elliott:
Some fake smtp servers to look at:

https://github.com/kirviq/dumbster
https://www.npmjs.com/package/fake-smtp-server

james-d-elliott:
Unless there are objections, I think this would provide the most portable and lightweight method of testing: https://nodemailer.com/extras/smtp-server/

nightah:
> Also some points on the decision of the order of auth mechanism preference. PLAIN is supported and maintained by the golang library and we know it works, LOGIN is custom, and CRAM-MD5 is untested.

To be honest I've never seen CRAM-MD5 out in the wild, I think my preference would be to just support PLAIN and LOGIN as I do not believe there's an explicit case to support CRAM-MD5.

I don't have an issue with the suggested method for testing, I believe the only reason @clems4ever suggested Postfix was because you're likely to run in to a lot of mail servers/relays that utilise Postfix.

nightah:
https://hub.docker.com/r/catatnight/postfix while not updated could even be used for the testing.

@clems4ever if you still want to stick to postfix and provide our own configuration we could simply just do something very similar to this: https://github.com/catatnight/docker-postfix/blob/master/assets/install.sh

james-d-elliott:
I'm fine with ditching CRAM-MD5, I've not seen many services that don't use PLAIN or LOGIN (or both). Even MS Exchange supports LOGIN after STARTTLS, and NTLM is hardly worth testing at all - I just used it to test a failure (unknown AUTH mech). I mainly implemented it because it was part of the core library. I also have no real preference on test suite direction, though I'm willing to help once one is chosen.

If we're wanting to test STARTTLS as part of the test suite with Secure = True, it will probably be complicated - it may be worth skipping. We would either need to add certs to the CI/CD test VM's trust store or we'll have to do something with the tlsconfig (crypto/tls.Config).

nightah:
> If we're wanting to test STARTTLS as part of the test suite with Secure = True, it will probably be complicated - it may be worth skipping. We would either need to add certs to the CI/CD test VM's trust store or we'll have to do something with the tlsconfig (crypto/tls.Config).

We can include the certs in the `internal/suites/x` path along with other configuration relevant to the new suite. Or we could just generate a self-signed cert alongside the service.

james-d-elliott:
I will probably have the following done in the next 24 hours:

1. [x] Moving internal/utils/smtp_login_auth.go to internal/notifier/smtp_login_auth.go
2. [x] Throwing an error in both of the following circumstances if the user configured a password:
 a. Server doesn't advertise AUTH extension.
 b. Server doesn't advertise AUTH mechanism that Authelia supports.
3. [ ] Rebasing with authelia/authelia:master so the merge is easy.
4. [ ] A retest of all functionality.

I'm pretty sure that will make it merge/release ready, though I'm not sure if you want a test suite before merging or not. If you do I may need some assistance with that. I will at least make a list of things to test.

@clems4ever can you confirm the following:

1. In a situation where smtp password is set, we want to force authentication. So if the server specified does not offer AUTH or the mechanisms it offers are not compatible the sending of the email in this situation will fail (user should configure correctly).
2. Do we want to leave the CRAM-MD5 mechanism in there. I have tested it and it works as I expected without changes. It's not hugely popular as mentioned by @nightah but since it works it may be worth keeping - though it may be harder to test than LOGIN or PLAIN, and most SMTP servers offer at least one of these (ms-exchange even does after STARTTLS).

clems4ever:
Great job @james-d-elliott ! And thank you @nightah for the support.

I'm ok with 1. and 2. I'm also ok to merge the feature when it's ready and we'll add the integration tests in another PR. I'll show you how to do it so that next time you're autonomous on that part.

Regarding the solution to test the feature, I'd rather use postfix if you guys can help me build the various configurations to run the tests against (I'm a noob in postfix config), otherwise the nodemailer solution would fit too. I did not know that they had a complete SMTP server implementation. So to summarize, I'd prefer postfix because it would also provide examples on how to setup a SMTP server in a target environment but I need assistance to have the configs running in a docker container otherwise I'll fallback to nodemailer.

For certs, if we don't test with verify=true, that's fine for the beginning. It's a point which has been skipped in other tests as well anyway.

@james-d-elliott , let us know when we can have a final review of your work before merging.

internal/notification/smtp_notifier.go
```diff
@@ -3,50 +3,106 @@ package notification
...
+				log.Debug("SMTP server supports AUTH PLAIN, attempting")
+			} else if utils.IsStringInSlice("LOGIN", mechanisms) {
+				auth = utils.LoginAuth(n.username, n.password)
+				log.Debug("SMTP server supports AUTH LOGIN, attempting")
+			} else if utils.IsStringInSlice("CRAM-MD5", mechanisms) {
+				auth = smtp.CRAMMD5Auth(n.username, n.password)
+				log.Debug("SMTP server supports AUTH CRAM-MD5, attempting")
+			}
+			if auth == nil {
+				return errors.New("SMTP server does not advertise a AUTH mechanism that Authelia Supports")
```
clems4ever: Idea for the future: we should probably introduce a hook at Authelia startup to perform early checks of configuration: "is SMTP server correctly configured?", "is db correctly configured?", "are credentials against service X ok?"
This way users could notice failures early during set up.

internal/notification/smtp_notifier.go
```diff
@@ -3,50 +3,106 @@ package notification
...
+			}
+			err := c.Auth(auth)
+			if err != nil {
+				return err
+			} else {
+				log.Debug("SMTP AUTH completed without error")
+			}
+		} else {
+			log.Debug("SMTP config has a password provided but server does not support AUTH, skipping")
```
clems4ever: As suggested, there should be an error thrown here because user provides a password which is not consumed by the server.

internal/utils/smtp.go
```diff
@@ -0,0 +1,34 @@
+package utils
```
clems4ever: Please move this file to the notifier package as suggested.

internal/notification/smtp_notifier.go
james-d-elliott: Done.

internal/utils/smtp.go
james-d-elliott: Done.

james-d-elliott:
I've rebased and squashed everything. I only have to create a new list of tests, test it all, and ensure there are no obvious mistakes. You're fine to do other commits obviously.

internal/notification/smtp_notifier.go
james-d-elliott: This is very doable, and would probably force me to get more familiar with the code base.

james-d-elliott: I will look into this. As you said future I think for the purposes of the PR you intend to resolve it prior to merging; but I'll leave this open for you to decide.

This part will probably involve two things, implement each stage of the send mail process into a function which returns an error or nil (i.e. STARTTLS, AUTH). I'm going to quickly see if I can implement this split of functionality now without breaking anything.

james-d-elliott: I had a look at this, but I think it may require a bit more thought and full implementation instead of just splitting the functionality.

james-d-elliott:
I believe this is ready for a merge. I have added a full list of tests I believe are suitable for these changes, tested them, and had a thorough read of the code.

I identified one issue and have force pushed a fix to it. I put in a check before AUTH PLAIN to check if STARTTLS occurred before attempting AUTH, this was for testing for possible future functionality which was not intended for my PR.

internal/notification/smtp_notifier.go
clems4ever: Definitely not a quick win, this is why I mentioned "for the future" meaning later in another review. The PR as it is is fine for me, the goal is reached. Thanks.

internal/notification/smtp_notifier.go
```diff
@@ -1,52 +1,112 @@
...
+		} else {
+			log.Debug("SMTP STARTTLS completed without error")
+		}
+	} else {
+		log.Debug("SMTP server does not support STARTTLS, skipping")
+	}
+
+	// Attempt AUTH if password is specified only
+	if n.password != "" {
+
```
clems4ever: Add a warning when authentication takes place without STARTTLS. This should raise a security red flag to the administrator.
We could print something like "Authentication occurred over an insecure connection. Using STARTTLS is recommended to keep credentials safe."

james-d-elliott: Good idea.

james-d-elliott:
> .

This is all done. I plan to keep helping to improve it. The code base is very understandable and well thought out, and I really like Authelia. I have a working startup check for reference, but not overly happy with it. I'll probably make a separate draft PR so you can take a look.

I'm curious what you mean about full TLS support, do you mean SMTPS instead of STARTTLS? Or some additional configuration?

nightah:
SMTPS has more or less been deprecated for a long time now. STARTTLS is the de-facto default in the mailserver world for default port 25 and submission typically supports STARTTLS too.

Perhaps @clems4ever means TLS across all of Authelia including the webserver and ldap implementations?

clems4ever:
> I'm curious what you mean about full TLS support, do you mean SMTPS instead of STARTTLS? Or some additional configuration?

I indeed meant connection over TLS from beginning to end. Not only an upgrade for credentials exchange but I did not know that SMTPS has been deprecated though and I misunderstood the STARTTLS command. I was thinking it was only upgrading to TLS for credential exchange but it actually upgrade the entire connection so that's all fine, you can forget what I wrote. Sorry.

clems4ever:
> This is all done. I plan to keep helping to improve it. The code base is very understandable and well thought out, and I really like Authelia. I have a working startup check for reference, but not overly happy with it. I'll probably make a separate draft PR so you can take a look.

Very nice! Regarding your attempt I think you should isolate the checks in a method of the configuration package so that it's easier to unit test in isolation. The aim is to have multiple kind of checks done eventually (connection tests, user hashes validity checks, ldap binding checks, etc).

clems4ever:
Also regarding this PR, since we won't support SMTPS which is totally acceptable given it has been deprecated, I think we should have an option to force STARTTLS to avoid downgrade attacks.

clems4ever:
I've created an issue for what I propose: https://github.com/authelia/authelia/issues/524.
Let me know what you think about it.

PR #519, opened 2019-12-23
Rename org from clems4ever to authelia

nightah:
Also fix references from config.yml to configuration.yml


clems4ever:
Thank you @nightah !

PR #535, opened 2020-01-05
Cross compile Go binary natively from amd64

nightah:
Reduce reliance on QEMU in order to speed up the pipeline.
Initial tests with a system under full load shows about 2-3x faster build stages, which was our single longest step.


clems4ever:
Do we still need the binary that is downloaded by authelia-scripts to run the container on the target arch using virtualization?

nightah:
Just for the sake of posterity, the QEMU binaries are still required due to the `apk` command we run in the final `arm32v7/arm64v8` Authelia containers.

PR #545, opened 2020-01-11
Remove Travis and promote Buildkite

nightah:
Given that the node issue in Buildkite has been resolved and the pipeline has been optimised significantly since it's initial implementation we are probably in a position to consider cutting over.

The pipeline seems on average to perform about 3x faster than Travis due to parallelism, dependency definitions/non-blocking and the 5 step concurrency limitation in Travis, as observed here:
[Travis Build](https://travis-ci.org/authelia/authelia/builds/635348803?utm_source=github_status&utm_medium=notification)
[Buildkite Build](https://buildkite.com/authelia/authelia/builds/107)

The pipeline also does not rebuild a PR if it has an existing build against the specified commits.
This means branches within the authelia repo will not be re-built when a PR is submitted, the commit statuses from the original build inside the branch will carry across to the PR.

This behaviour would cause significant blocking delays with Travis if there was active development in branches within the main repo.

This will also need an update to the GitHub policies to require the Buildkite commit statues as opposed to Travis.

@clems4ever I haven't put explicit notifications inside the Buildkite pipeline, however you can set this up in the buildkite interface and star the branch you want to be notified on (presumably master).
If we need/want to have more complex notifications setup, happy to add that in just let me know.


cmd/authelia-scripts/docker.go
```diff
@@ -61,3 +61,8 @@ func (d *Docker) CleanTag(tag string) error {
 func (d *Docker) PublishReadme() error {
 	return utils.CommandWithStdout("bash", "-c", `token=$(curl -fs --retry 3 -H "Content-Type: application/json" -X "POST" -d '{"username": "'$DOCKER_USERNAME'", "password": "'$DOCKER_PASSWORD'"}' https://hub.docker.com/v2/users/login/ | jq -r .token) && jq -n --arg msg "$(cat README.md | sed -r 's/(\<img\ src\=\")(\.\/)/\1https:\/\/github.com\/authelia\/authelia\/raw\/master\//' | sed 's/\.\//https:\/\/github.com\/authelia\/authelia\/blob\/master\//g')" '{"registry":"registry-1.docker.io","full_description": $msg }' | curl -fs --retry 3 -o /dev/null -L -X "PATCH" -H "Content-Type: application/json" -H "Authorization: JWT $token" -d @- https://hub.docker.com/v2/repositories/authelia/authelia/`).Run()
 }
+
+// UpdateMicroBadger updates MicroBadger metadata based on dockerhub.
+func (d *Docker) UpdateMicroBadger() error {
+	return utils.CommandWithStdout("curl", "-fs", "--retry", "3", "-X", "POST", "https://hooks.microbadger.com/images/authelia/authelia/6b8tWohGJpS4CbbPCgUHxVe_uY4=").Run()
```
clems4ever: Can anyone publish here without credentials?

clems4ever: OK got it, it's only a trigger, sorry.

PR #548, opened 2020-01-11
Remove time from binary artifact to have reproducible builds

clems4ever:


nightah:
@clems4ever I feel like having the git tag could still be useful to us because people will be able to download the binary for `master` builds without tags and the fact that we also have the `master` Docker container.

I'm about to submit a PR to this branch, if you agree just merge it in here first otherwise feel free to delete the other branch and close the PR.

clems4ever:
I'm ok to keep the tag. Let's go ahead!

nightah:
Haha merged in the wrong order so the other commit didn't get to master.
I'll fix it up!

PR #554, opened 2020-01-17
Fixes Remember Me functionality

james-d-elliott:
Adjust the remember me duration to 1 year. Was originally set to 0 which expires at the end of the user agents session. Usually this is when the browser is closed, tab is closed.

This functionality can still likely be reached by not selecting remember me, and having the expiration set to 0.

The behavior after this commit will be as follows (tested by me):
  Remember Me unchecked: duration set as expiration config.
  Remember me checked: duration set to 1 year.

### DIFF
--- a/internal/handlers/handler_firstfactor.go
+++ b/internal/handlers/handler_firstfactor.go
@@ -76,9 +76,9 @@ func FirstFactorPost(ctx *middlewares.AutheliaCtx) {
 		return
 	}
 
-	// and avoid the cookie to expire if "Remember me" was ticked.
+	// set the cookie to expire in 1 year if "Remember me" was ticked.
 	if *bodyJSON.KeepMeLoggedIn {
-		err = ctx.Providers.SessionProvider.UpdateExpiration(ctx.RequestCtx, time.Duration(0))
+		err = ctx.Providers.SessionProvider.UpdateExpiration(ctx.RequestCtx, time.Duration(31556952 * time.Second))
 		if err != nil {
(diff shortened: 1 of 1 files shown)


james-d-elliott:
That failure is almost 100% due to my PR, though not having much luck figuring out why.

nightah:
I'm having a look now to see what's going on with the integration tests, I wonder if originally it was a false positive.
I'll let you know what I find out shortly.

nightah:
Okay so @clems4ever I haven't had any luck getting a non-headless Chrome environment working to see what's going on. Could you perhaps have a look at this one when you get a chance please?

The test failure is reproducible, though the binary and code does definitely result in the correct outcome as documented, the "Remember Me" option now results in a cookie which has a 1 year session timer.

clems4ever:
Checking now.

james-d-elliott:
This looks like it's due to the inactivity timeout. It's set to 5 seconds in that suite, and the scenario is waiting 9 seconds. I'm still investigating.

james-d-elliott:
I guess I was wrong, but that change needs to occur anyway.

clems4ever:
I found the issue and made a fix for that: https://github.com/authelia/authelia/pull/557

clems4ever:
@james-d-elliott , please revert the latest fix, you won't need it once my PR is merged. Let me merge it first and then you'll be able to rebase.

james-d-elliott:
Will do. Ignore my comment on the other PR @clems4ever .

PR #555, opened 2020-01-17
Build docker image upfront in CI and use it in integration tests.

clems4ever:
Previously, the development workflow was broken because the container
generated from Dockerfile.CI was used in dev environments but the binary
was not pre-built as it is on buildkite. I propose to just remove that
image and use the "to be published" image instead in integration tests.

This will have several advantages:
- Fix the dev workflow.
- Remove CI arch from authelia-scripts build command
- Optimize CI time in buildkite since we'll cache a way small artifact
- We don't build authelia more than once for earch arch.


.buildkite/pipeline.yml
```diff
@@ -3,16 +3,21 @@ steps:
     command: "authelia-scripts --log-level debug ci"
     key: "build"
 
+  - wait
```
clems4ever: you don't need to wait here IMO

.buildkite/pipeline.yml
```diff
@@ -3,16 +3,21 @@ steps:
...
+
   - label: ":docker: Image Builds"
     command: ".buildkite/steps/buildimages.sh | buildkite-agent pipeline upload"
     agents:
       build: "true"
+    depends_on: ~
```
clems4ever: What does this line mean?

PR #557, opened 2020-01-17
Use keep me logged in bool

clems4ever:
This must be merged prior to https://github.com/authelia/authelia/pull/554 to fix the tests.

This PR uses keep me logged boolean stored in the user session instead of relying on the expiration value stored in the cookie. That way we will be able to set an expiration to 1 year instead of 0.

This PR also removes user session caching in AutheliaCtx to make sure that user session is correctly wiped when inactivity timeout is reached.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,6 +2,8 @@ steps:
   - label: ":hammer_and_wrench: Unit Test"
     command: "authelia-scripts --log-level debug ci"
 
+  - wait
+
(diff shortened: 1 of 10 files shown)


james-d-elliott:
Ah I see. The KeepMeLoggedIn was not set in the session.

We've both handled one of the issues in differing ways. This should work as an alternative to your changes in verifyFromSessionCookie(). https://github.com/authelia/authelia/pull/554/commits/79a6805ab077fc58a945e129c540280d75bd08e2

PR #569, opened 2020-01-21
Use viper to read secrets from env variables

clems4ever:
In order to prepare Authelia to be deployed with a Helm chart, it needed to be able to read passwords and secrets from env variables in order to use the Kubernetes vault. This PR makes it possible.

Today, we must list the variables bound to env variables because of a missing feature in viper but in the long term we should be able to automatically resolve them according to the predefined schema.


J7mbo:
Hi. Does this provide support for any environment variable being used in `config.yml`, if passed in say under `envrionment:` with `docker-compose`?

james-d-elliott:
Hi @J7mbo I'd recommend reading this. We're deprecating this feature. https://docs.authelia.com/configuration/secrets.html

Basically you need to mount a file in the container, and provide it one of those secret ENV's with _FILE at the end under environment with the path to the file containing the secret.

PR #580, opened 2020-01-25
[MISC] Update docs to include updated proxy configuration

nightah:
Includes updated documentation for:
- nginx
- Traefik 1.x
- Traefik 2.x

### DIFF
--- a/docs/proxies/traefik1.x.md
+++ b/docs/proxies/traefik1.x.md
@@ -1,0 +1,90 @@
+# Nginx
+
+[Traefik 1.x] is a reverse proxy supported by **Authelia**.
+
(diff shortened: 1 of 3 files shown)


docs/proxies/traefik1.x.md
```diff
@@ -0,0 +1,90 @@
+# Nginx
```
clems4ever: Should be traefik

docs/proxies/traefik2.x.md
```diff
@@ -0,0 +1,94 @@
+# Nginx
```
clems4ever: Should be Traefik2

Issue #585, opened 2020-01-26
NPE when unable to connect to SMTP server (v4)


aroberts:
Testing email-sending in a fresh v4.2.0 deploy, and stumbled across this:

```
time="2020-01-26T00:54:18Z" level=debug msg="LDAP client starts a session over raw TCP"                                                       
time="2020-01-26T00:54:18Z" level=debug msg="Sending an email to user [redacted] ([redacted]@gmail.com) to confirm identity for registering a device." method=POST path=/api/reset-password/identity/start remote_ip=192.168.1.1                                                                                
time="2020-01-26T00:54:18Z" level=debug msg="Notifier SMTP client attempting connection to smtp.gmail.com:587"                                
panic: runtime error: invalid memory address or nil pointer dereference                                                                       
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xb84bc6]                                                                        
                                                                                                                                              
goroutine 27 [running]:                                                                                                                       
net/smtp.(*Client).Close(...)                                                                                                                 
 net/smtp/smtp.go:76                                                                                                                          
github.com/authelia/authelia/internal/notification.(*SMTPNotifier).Send(0xc000310a00, 0xc0001501c0, 0x14, 0xe2b11d, 0x1e, 0xc00038e000, 0x399a
, 0xc000056250, 0x3)                                                                                                                                         
 github.com/authelia/authelia@/internal/notification/smtp_notifier.go:215 +0x546                                                              
github.com/authelia/authelia/internal/middlewares.IdentityVerificationStart.func1(0xc000144000)                                               
 github.com/authelia/authelia@/internal/middlewares/identity_verification.go:70 +0xa13                                                        
github.com/authelia/authelia/internal/middlewares.AutheliaMiddleware.func1.1(0xc0000d3900)                                                    
 github.com/authelia/authelia@/internal/middlewares/authelia_context.go:46 +0x180                                                             
github.com/fasthttp/router.(*Router).Handler(0xc0004ab3b0, 0xc0000d3900)                                                                      
 github.com/fasthttp/router@v0.5.2/router.go:326 +0xfb3                                                                                       
github.com/authelia/authelia/internal/middlewares.LogRequestMiddleware.func1(0xc0000d3900)                                                    
 github.com/authelia/authelia@/internal/middlewares/log_request.go:14 +0x108                                                                  
[...]

PR #589, opened 2020-01-27
[FIX] SMTP Notifier Unhandled Error Conditions

james-d-elliott:
- Only attempt to close the connection once it's established.
- Defer the client Quit/Close so that it always executes at the end.

Draft: I'm still a bit concerned that this may have issues if the socket is broken during execution. This should only occur in very rare instances when there is a network interruption or if the SMTP server is not configured correctly. So I'm leaving it as a draft for now.

This relates to #585. What was happening is the connection completely failed to initialize, and we were trying to close a connection that was not opened at all. This is my fault.

Note: When this is ready I'll amend my commit with a fixes note (I completely forgot doing this too early can make an issue look utterly horrible). I'll also mark it ready to merge.

### DIFF
--- a/internal/notification/smtp_notifier.go
+++ b/internal/notification/smtp_notifier.go
@@ -14,7 +14,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// SMTPNotifier a notifier to send emails to SMTP servers
+// SMTPNotifier a notifier to send emails to SMTP servers.
 type SMTPNotifier struct {
 	username          string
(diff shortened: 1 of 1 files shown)


nightah:
Don't stress too badly if you forget to include notes in the commit, the last safety net is that we can always perform a squash commit and rename with the according tag.

james-d-elliott:
As per the user reporting the issue, it's fixed with this commit. I think my initial concerns are overly cautious.

Issue #579, opened 2020-01-25
Document secret configuration via environment variables


aroberts:
This was fixed as #310, but the env vars are no longer read after the port to go.

nightah:
The code wasn't backported to v4 until recently https://github.com/authelia/authelia/commit/c95c7210d8140ffd37c5d70233e42cf9010faa4b.

I think the issue you are probably seeing is that the variables are now prefixed with `AUTHELIA`.

Please give that a try and let me know if you still have issues.

aroberts:
Ah, I did indeed miss that. I'm running a 4.2 instance, I'll verify tonight and close if I get it working

aroberts:
Confirmed that I can pass secrets via envvars. It looks like the relevant code is here:

https://github.com/authelia/authelia/blob/bb7781fd2b2ea563a6705423660a80e8542c5b13/internal/configuration/reader.go

I'm leaving this ticket open as a request for documentation. Thanks for your help!

PR #606, opened 2020-02-01
[MISC] Document usage of env variables for setting secrets.

clems4ever:
This PR fixes: https://github.com/authelia/authelia/issues/579


config.template.yml
```diff
@@ -11,6 +11,7 @@ logs_level: debug
 
 # The secret used to generate JWT tokens when validating user identity by
 # email confirmation.
+# This secret could also be set using the env variables AUTHELIA_JWT_SECRET
```
nightah: could -> can

clems4ever: done

PR #631, opened 2020-02-11
[MISC] Fail with an error message when X-Forwarded-* headers are miss…

clems4ever:
…ing.

### DIFF
--- a/internal/handlers/handler_register_u2f_step1_test.go
+++ b/internal/handlers/handler_register_u2f_step1_test.go
@@ -1,0 +1,98 @@
+package handlers
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/authelia/authelia/internal/middlewares"
+	"github.com/authelia/authelia/internal/mocks"
+	"github.com/dgrijalva/jwt-go"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type HandlerRegisterU2FStep1Suite struct {
+	suite.Suite
+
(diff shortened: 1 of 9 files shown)


internal/handlers/handler_register_u2f_step1_test.go
```diff
@@ -0,0 +1,98 @@
...
+		FindIdentityVerificationToken(gomock.Eq(token)).
+		Return(true, nil)
+
+	s.mock.StorageProviderMock.EXPECT().
+		RemoveIdentityVerificationToken(gomock.Eq(token)).
+		Return(nil)
+
+	SecondFactorU2FIdentityFinish(s.mock.Ctx)
+
+	// Return 200 KO
```
james-d-elliott: Insignificant Typo.

internal/handlers/handler_register_u2f_step1_test.go
```diff
@@ -0,0 +1,98 @@
...
```
james-d-elliott: Insignificant Typo. Should be "// Return 200 OK".

internal/handlers/handler_register_u2f_step1_test.go
james-d-elliott: Resolving this in favor of the more specific one. There are a few instances of this.

internal/handlers/handler_register_u2f_step1_test.go
james-d-elliott: There are actually quite a few, probably just a copy paste.

clems4ever: I've just removed them.

PR #642, opened 2020-02-19
[Docker] Include docker-compose.yml examples to run Authelia

### LINKED ISSUE #480: Create and document self-sufficient docker-compose setups
Using suites is too complicated for non dev people so we need to provide self-sufficient docker-compose setups to deploy the stack in seconds in order to demo authelia.

Having a "lite" and "complete" flavors would be great.

- Create docker-compose bundle for highly available setup and document it
- Create docker-compose bundle for standalone setup and document it

### LINKED ISSUE #610: Unable to get standalone demo working
Thank you for open sourcing this great project! I tried setting up the [standalone suite](https://github.com/authelia/authelia/blob/master/docs/getting-started.md) but unfortunately failed to do so at several stages:

- The [listed commands](https://github.com/authelia/authelia/blob/master/docs/getting-started.md#getting-started) do not work for me because, if I understand correctly, `authelia scripts` is not `go install`'ed which means that I have to do `cmd/authelia-scripts/authelia-scripts suites setup Standalone`. If this is intended behavior maybe the docs could be updated here?
- The Dockerfiles fail to build ([a](https://github.com/authelia/authelia/blob/master/example/compose/authelia/Dockerfile.backend), [b](https://github.com/authelia/authelia/blob/master/example/compose/authelia/Dockerfile.frontend#L7)) because `addgroup` does not accept `--guid` but `-G` instead. I was still unable to get this working with updating the the commands so I (for local testing) removed the user features from the Dockerfiles.
- And finally `docker-compose` did not come up but instead timed out after 300 seconds. I have a pretty beefy system and never have seen `docker-compose up` time out like that. I believe the problem is that `npm ci` and `npm run start` are used [at the entrypoint](https://github.com/authelia/authelia/blob/master/example/compose/authelia/resources/entrypoint-frontend.sh). Both processes, especially when dealing with webpack, are very time intense and cause this issue in my understanding.

What is the best way of checking out a demo of this project?

I was also wondering if there was more documentation on the architecture of Authelia (e.g. "Why Authelia stores users in LDAP"), API documentation, the portal.

I understand if this is not something you / the maintainers want to spend their time on, but it would be cool to get the project working and give it a ride :)

### LINKED ISSUE #681: Issue running standalone server on osx 10.15.3 (catalina)
Using latest source files from this time (3/2/20).  Using source bootstrap.sh then 
$ authelia-scripts suites setup Standalone

Several notes on gitter page...working with Amir on this, he suggested I file an issue.

Building nginx-backend
Step 1/3 : FROM nginx:alpine
---> 48c8a7c47625
Step 2/3 : ADD html /usr/share/nginx/html
---> c5036c8fb47d
Step 3/3 : ADD nginx.conf /etc/nginx/nginx.conf
---> 6270b1f089a2
Successfully built 6270b1f089a2
Successfully tagged authelia_nginx-backend:latest
Creating authelia_authelia-frontend_1 ... done
Creating authelia_authelia-backend_1 ... done
Creating authelia_nginx-portal_1 ... done
Creating authelia_smtp_1 ... done
Creating authelia_nginx-backend_1 ... done
.........................................Timeout of 300s reached... Killing process...

here is log from frontend

Allens-Mac:authelia allenharper$ docker logs -f authelia_authelia-frontend_1
yarn install --frozen-lockfile
yarn install v1.22.0
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@2.0.7: The platform "linux" is incompatible with this module.
info "fsevents@2.0.7" is an optional dependency and failed compatibility check. Excluding it from installation.
info fsevents@1.2.11: The platform "linux" is incompatible with this module.
info "fsevents@1.2.11" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
warning " > react-ga@2.7.0" has unmet peer dependency "prop-types@^15.6.0".
warning " > react-loading@2.0.3" has unmet peer dependency "prop-types@^15.6.0".
error An unexpected error occurred: "ENOENT: no such file or directory, lstat '/app/node_modules/postcss-flexbugs-fixes/bugs/'".
info If you think this is a bug, please open a bug report with the information provided in "/app/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.

### LINKED ISSUE #783: Unable to get standalone demo working
Hi, i'm docker newby so I don't know what is going on. When I try to start the suite i get this error:
addgroup: gid '999' in use
ERROR: Service 'authelia-frontend' failed to build: The command '/bin/sh -c deluser node &&     addgroup --gid ${GROUP_ID} dev &&     adduser --uid ${USER_ID} -G dev -D dev' returned a non-zero code: 1
gid 999 belongs to docker by the way

the complete errors are 
addgroup: gid '999' in use
ERROR: Service 'authelia-frontend' failed to build: The command '/bin/sh -c deluser node &&     addgroup --gid ${GROUP_ID} dev &&     adduser --uid ${USER_ID} -G dev -D dev' returned a non-zero code: 1
ERRO[0002] Failure during environment deployment.       
Removing network authelia_authelianet
INFO[0003] Environment has been cleaned!                
FATA[0003] exit status 1                                
exit status 1
INFO[0004] Tear down environment for suite Standalone... 
Removing network authelia_authelianet
WARNING: Network authelia_authelianet not found.
INFO[0000] remove .suite: no such file or directory     
INFO[0000] Environment has been cleaned!                
FATA[0005] exit status 1                                
exit status 1
The SO  is a ubuntu18

nightah:
This provides two flavours of compose bundles for users to test Authelia.

- Local
This bundle is intended for scenarios where the server will not be exposed to the internet, domains will be defined in the local hosts file and self-signed certificates will be utilised.
This bundle includes a script to generate all the required configuration and is intended as the quickest method to demo Authelia.

- Lite
This bundle is intended for scenarios where the server will be exposed to the internet, domains and DNS will need to be setup accordingly and certificates will be generated through LetsEncrypt.
The Lite element refers to minimal external dependencies; File based user storage, SQLite based configuration storage. In this configuration, the service will not scale well.

A future PR will also include the Full bundle which is intended to be a fully scalable and highly-available setup.

Closes #480.
Closes #610.
Closes #681.
Closes #783.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,10 +79,24 @@ Docker or on top of [Kubernetes].
 
 ## Getting Started
 
-You can start off with
+You can start utilising Authelia with the provided `docker-compose` bundles:
 
-    git clone https://github.com/authelia/authelia.git && cd authelia
-    source bootstrap.sh
+##### Local
+The Local compose bundle is intended for scenarios where the server will not be exposed to the internet, domains will be defined in the local hosts file and self-signed certificates will be utilised.  
+
+##### Lite
+The Lite compose bundle is intended for scenarios where the server will be exposed to the internet, domains and DNS will need to be setup accordingly and certificates will be generated through LetsEncrypt.
+The Lite element refers to minimal external dependencies; File based user storage, SQLite based configuration storage. In this configuration, the service will not scale well.
+
+##### Full
+The Full compose bundle is intended for scenarios where the server will be exposed to the internet, domains and DNS will need to be setup accordingly and certificates will be generated through LetsEncrypt.
+The Full element refers to a scalable setup which includes external dependencies; LDAP based user storage, Database based configuration storage (MariaDB or Postgres). 
+
+- `git clone https://github.com/authelia/authelia.git`
+- `cd authelia/compose/{local,lite,full}` based on the bundle you plan to run
+- If running Local or Lite modify the `users_database.yml` the default username and password is `authelia`
+- Modify the `configuration.yml` and `docker-compose.yml` with your respective domains and secrets
+- `docker-compose up -d`
 
 If you want to go further, please read [Getting Started](https://docs.authelia.com/getting-started).
 
--- a/README.md
+++ b/README.md
@@ -82,8 +82,9 @@ Docker or on top of [Kubernetes].
 You can start utilising Authelia with the provided `docker-compose` bundles:
 
 ##### Local
-The Local compose bundle is intended for scenarios where the server will not be exposed to the internet, domains will be defined in the local hosts file and self-signed certificates will be utilised.
-This bundle includes a script to generate all the required configuration and is intended as the quickest method to demo Authelia.
+The Local compose bundle is intended to test Authelia without worrying about configuration.
+It's meant to be used for scenarios where the server is not be exposed to the internet.
+Domains will be defined in the local hosts file and self-signed certificates will be utilised.
 
 ##### Lite
 The Lite compose bundle is intended for scenarios where the server will be exposed to the internet, domains and DNS will need to be setup accordingly and certificates will be generated through LetsEncrypt.
--- a/README.md
+++ b/README.md
@@ -82,7 +82,8 @@ Docker or on top of [Kubernetes].
 You can start utilising Authelia with the provided `docker-compose` bundles:
 
 ##### Local
-The Local compose bundle is intended for scenarios where the server will not be exposed to the internet, domains will be defined in the local hosts file and self-signed certificates will be utilised.  
+The Local compose bundle is intended for scenarios where the server will not be exposed to the internet, domains will be defined in the local hosts file and self-signed certificates will be utilised.
+This bundle includes a script to generate all the required configuration and is intended as the quickest method to demo Authelia.
 
 ##### Lite
 The Lite compose bundle is intended for scenarios where the server will be exposed to the internet, domains and DNS will need to be setup accordingly and certificates will be generated through LetsEncrypt.
@@ -90,13 +91,7 @@ The Lite element refers to minimal external dependencies; File based user storag
 
 ##### Full
 The Full compose bundle is intended for scenarios where the server will be exposed to the internet, domains and DNS will need to be setup accordingly and certificates will be generated through LetsEncrypt.
-The Full element refers to a scalable setup which includes external dependencies; LDAP based user storage, Database based configuration storage (MariaDB or Postgres). 
-
-- `git clone https://github.com/authelia/authelia.git`
-- `cd authelia/compose/{local,lite,full}` based on the bundle you plan to run
-- If running Local or Lite modify the `users_database.yml` the default username and password is `authelia`
-- Modify the `configuration.yml` and `docker-compose.yml` with your respective domains and secrets
-- `docker-compose up -d`
+The Full element refers to a scalable setup which includes external dependencies; LDAP based user storage, Database based configuration storage (MariaDB, MySQL or Postgres). 
 
 If you want to go further, please read [Getting Started](https://docs.authelia.com/getting-started).
 
--- a/README.md
+++ b/README.md
@@ -81,20 +81,18 @@ Docker or on top of [Kubernetes].
 
 You can start utilising Authelia with the provided `docker-compose` bundles:
 
-##### Local
+##### [Local](https://docs.authelia.com/getting-started)
 The Local compose bundle is intended to test Authelia without worrying about configuration.
 It's meant to be used for scenarios where the server is not be exposed to the internet.
 Domains will be defined in the local hosts file and self-signed certificates will be utilised.
 
-##### Lite
+##### [Lite](https://docs.authelia.com/deployment/deployment-lite)
 The Lite compose bundle is intended for scenarios where the server will be exposed to the internet, domains and DNS will need to be setup accordingly and certificates will be generated through LetsEncrypt.
 The Lite element refers to minimal external dependencies; File based user storage, SQLite based configuration storage. In this configuration, the service will not scale well.
 
-##### Full
+##### [Full](https://docs.authelia.com/deployment/deployment-ha)
 The Full compose bundle is intended for scenarios where the server will be exposed to the internet, domains and DNS will need to be setup accordingly and certificates will be generated through LetsEncrypt.
 The Full element refers to a scalable setup which includes external dependencies; LDAP based user storage, Database based configuration storage (MariaDB, MySQL or Postgres). 
-
-If you want to go further, please read [Getting Started](https://docs.authelia.com/getting-started).
 
 ## Deployment
 
--- a/compose/lite/configuration.yml
+++ b/compose/lite/configuration.yml
@@ -1,0 +1,65 @@
+###############################################################
+#                   Authelia configuration                    #
+###############################################################
+
+host: 0.0.0.0
+port: 9091
+log_level: debug
+# This secret can also be set using the env variables AUTHELIA_JWT_SECRET
+jwt_secret: a_very_important_secret
+default_redirection_url: https://public.example.com
+totp:
+  issuer: authelia.com
+
+#duo_api:
+#  hostname: api-123456789.example.com
+#  integration_key: ABCDEF
+#  # This secret can also be set using the env variables AUTHELIA_DUO_API_SECRET_KEY
+#  secret_key: 1234567890abcdefghifjkl
+
+authentication_backend:
+  file:
+    path: /etc/authelia/users_database.yml
+
+access_control:
+  default_policy: deny
+  rules:
+    # Rules applied to everyone
+    - domain: public.example.com
+      policy: bypass
+    - domain: traefik.example.com
+      policy: one_factor
+    - domain: secure.example.com
+      policy: two_factor
+
+session:
+  name: authelia_session
+  # This secret can also be set using the env variables AUTHELIA_SESSION_SECRET
+  secret: unsecure_session_secret
+  expiration: 3600 # 1 hour
+  inactivity: 300 # 5 minutes
+  domain: example.com # Should match whatever your root protected domain is
+
+  redis:
+    host: redis
+    port: 6379
+    # This secret can also be set using the env variables AUTHELIA_SESSION_REDIS_PASSWORD
+    password: authelia
+
+regulation:
+  max_retries: 3
+  find_time: 120
+  ban_time: 300
+
+storage:
+  local:
+    path: /var/lib/authelia/db.sqlite3
+
+notifier:
+  smtp:
+    username: test
+    # This secret can also be set using the env variables AUTHELIA_NOTIFIER_SMTP_PASSWORD
+    password: password
+    host: mail.example.com
+    port: 25
+    sender: admin@example.com
--- a/compose/lite/docker-compose.yml
+++ b/compose/lite/docker-compose.yml
@@ -1,0 +1,108 @@
+version: '3.7'
+
+networks:
+  net:
+    driver: bridge
+
+services:
+  authelia:
+    image: authelia/authelia
+    container_name: authelia
+    volumes:
+      - ./authelia:/var/lib/authelia
+      - ./configuration.yml:/etc/authelia/configuration.yml:ro
+      - ./users_database.yml:/etc/authelia/users_database.yml:ro
+    networks:
+      - net
+    labels:
+      - 'traefik.enable=true'
+      - 'traefik.http.routers.authelia.rule=Host(`auth.example.com`)'
+      - 'traefik.http.routers.authelia.entrypoints=https'
+      - 'traefik.http.routers.authelia.tls=true'
+      - 'traefik.http.routers.authelia.tls.certresolver=letsencrypt'
+      - 'traefik.http.middlewares.authelia.forwardauth.address=http://authelia:9091/api/verify?rd=https://auth.example.com'
+      - 'traefik.http.middlewares.authelia.forwardauth.trustForwardHeader=true'
+    expose:
+      - 9091
+    restart: unless-stopped
+    environment:
+      - TZ=Australia/Melbourne
+
+  redis:
+    image: redis:alpine
+    container_name: redis
+    volumes:
+      - ./redis:/data
+    networks:
+      - net
+    expose:
+      - 6379
+    restart: unless-stopped
+    environment:
+      - TZ=Australia/Melbourne
+
+  traefik:
+    image: traefik:v2.1.3
+    container_name: traefik
+    volumes:
+      - ./traefik/acme.json:/acme.json
+      - /var/run/docker.sock:/var/run/docker.sock
+    networks:
+      - net
+    labels:
+      - 'traefik.enable=true'
+      - 'traefik.http.routers.api.rule=Host(`traefik.example.com`)'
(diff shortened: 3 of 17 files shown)


compose/lite/configuration.yml
```diff
@@ -0,0 +1,91 @@
...
+
+# This secret can also be set using the env variables AUTHELIA_JWT_SECRET
+jwt_secret: a_very_important_secret
+
+default_redirection_url: https://auth.example.com
+
+totp:
+  issuer: authelia.com
+
+duo_api:
```
clems4ever: This block should be commented

compose/lite/configuration.yml
```diff
@@ -0,0 +1,91 @@
...
+
+host: 0.0.0.0
+port: 9091
+
+logs_level: debug
+
+# This secret can also be set using the env variables AUTHELIA_JWT_SECRET
+jwt_secret: a_very_important_secret
+
+default_redirection_url: https://auth.example.com
```
clems4ever: I would name it public.example.com instead since this auth is bypassed for this domain. It would make it clearer.

compose/lite/configuration.yml
```diff
@@ -0,0 +1,91 @@
...
+
+  redis:
+    host: redis
+    port: 6379
+    # This secret can also be set using the env variables AUTHELIA_SESSION_REDIS_PASSWORD
+    password: authelia
+
+regulation:
+  max_retries: 3
+
```
clems4ever: you can remove the empty lines between the options.

compose/lite/configuration.yml
```diff
@@ -0,0 +1,91 @@
...
+notifier:
+  smtp:
+    username: test
+    # This secret can also be set using the env variables AUTHELIA_NOTIFIER_SMTP_PASSWORD
+    password: password
+    host: 127.0.0.1
+    port: 1025
+    sender: admin@example.com
+    ## disable_require_tls: false
```
clems4ever: Remove the commented options, they don't bring anything. For advanced use cases, users will have the documentation to review the entire set of options.

compose/lite/docker-compose.yml
```diff
@@ -0,0 +1,89 @@
...
+      - ./configuration.yml:/etc/authelia/configuration.yml:ro
+      - ./users_database.yml:/etc/authelia/users_database.yml:ro
+    networks:
+      - net
+    labels:
+      - 'traefik.enable=true'
+      - 'traefik.http.routers.authelia.rule=Host(`auth.example.com`)'
+      - 'traefik.http.routers.authelia.entrypoints=https'
+      - 'traefik.http.routers.authelia.tls=true'
+      - 'traefik.http.routers.authelia.tls.certresolver=letsencrypt'
```
clems4ever: Will this line make traefik request a cert to letsencrypt? I'd rather provide a self-signed cert to not rely on the fact that users have a DNS already set up (This because I think it's a barrier for testing Authelia, the minimum should be required to bootstrap the minimal setup).

compose/lite/docker-compose.yml
```diff
@@ -0,0 +1,89 @@
...
+      - ./traefik/acme.json:/acme.json
+      - /var/run/docker.sock:/var/run/docker.sock
+    networks:
+      - net
+    labels:
+      - 'traefik.http.routers.api.rule=Host(`traefik.example.com`)'
+      - 'traefik.http.routers.api.entrypoints=https'
+      - 'traefik.http.routers.api.service=api@internal'
+      - 'traefik.http.routers.api.tls=true'
+      - 'traefik.http.routers.api.tls.certresolver=letsencrypt'
```
clems4ever: same remark as previously, provide a self-signed cert.

compose/lite/users_database.yml
```diff
@@ -0,0 +1,14 @@
+###############################################################
+#                         Users Database                      #
+###############################################################
+
+# This file can be used if you do not have an LDAP set up.
+
+# List of users
+users:
+  authelia:
+    password: "$6$rounds=50000$BpLnfgDsc2WD8F2q$Zis.ixdg9s/UOJYrs56b5QEZFiZECu0qZVNsIYxBaNJ7ucIL.nlxVCT5tqh8KHG8X4tlwCFm5r6NTOZZ5qRFN/"
```
clems4ever: Add the password in clear text in comment.

compose/lite/configuration.yml
```diff
@@ -0,0 +1,91 @@
...
+  find_time: 120
+
+  ban_time: 300
+
+storage:
+  local:
+    path: /var/lib/authelia/db.sqlite3
+
+notifier:
+  smtp:
```
james-d-elliott: We might want to utilize the file based notifier for this example unless we're providing an SMTP server with it.

compose/lite/configuration.yml
james-d-elliott: We may want to add another container for public.example.com so there is all 3 examples using the traefik middleware (authelia). I've not used it but this may be a good container to use: https://hub.docker.com/r/yeasy/simple-web/

compose/lite/docker-compose.yml
nightah: That would mean that the user has to modify their hosts file like part of the bootstrap process, I don't think that's just as simple as "run this docker-compose".
Whereas including the LetsEncrypt configuration means that by virtue of exposing port 80 in the compose stack it will fetch the required certificates.

I would also say it should be a safe assumption to make that if someone plans to test this that they likely already have some domains setup and likely already Traefik or nginx too.

compose/lite/configuration.yml
nightah: I had originally considered this, my thought process was that the general consensus is that most people will have access to an mail server, most which are well documented as to how you would utilise these with your own mail clients/services and also the Gmail example which is partly why I included that commented.

As Clèment has suggested users that really want to customise can look at the config examples and make whatever determination that is necessary.

compose/lite/docker-compose.yml
clems4ever: Do you mean that the certificate will be generated even if the machine is not publicly available?

I think it won't. If traefik leverages the certbot with http method, it means that let's encrypts needs to complete the challenge with public access to the machine. In that case it means 2 things: opening the machine to the public and creating a DNS entry. It might be a barrier for people starting from scratch on a laptop. The /etc/host method has only one simple prerequisite instead of two that's why I like it better. If we go that way we might keep the cert generation options but comment them.

What do you think?

clems4ever: After second thought, I think we have 2 different objectives. You want to expose the maximum features in the minimal snippet so that people can start from a seed and adapt to their context but they MUST adapt (DNS and public exposition, mail server, etc...)

On my side, I'm currently working on the getting started and I'm trying to provide the easiest way to test authelia locally so the less pre-requisites the better. Now that I'm aware of this, I think your PR solves your objective and I will work on mine independently based on your docker-compose. We'll see if we can find a way to factorize afterwards if possible.

compose/lite/configuration.yml
james-d-elliott: Yeah that makes sense.

compose/lite/docker-compose.yml
nightah: > Do you mean that the certificate will be generated even if the machine is not publicly available?

The only way to make this work would be to use DNS based challenges as opposed to HTTP or TLS-SNI.

Having said that to make the webserver work as it would be intended via internet access the service would need to be exposed anyway, so I don't think we're really solving much by moving to DNS based challenges alone.

docs/deployment/deployment-local.md
```diff
@@ -0,0 +1,73 @@
+---
```
clems4ever: As explained in my comment in the getting started file below, I think you should not even bother explaining anything about reverse proxy, SQL, Redis, etc.... In my mind, this bundle is dedicated to the getting started which is here to just deploy an instance of Authelia with the minimum explanation so that the user can get glimpse of what Authelia looks like for real in less than a minute. The user can simply test and know whether this is what fits their needs. It's a pure showcase. And if they are happy they can proceed with local where all details will be provided.

docs/getting-started.md
```diff
@@ -6,7 +6,29 @@ nav_order: 2
 
 # Getting Started
 
-**Authelia** can be tested in a matter of seconds with Docker and docker-compose.
+## Docker Compose
```
clems4ever: IMO this getting started should contain what is in docs/deployment/deployment-local.md. docs/deployment/deployment-local.md should not exist. Only lite, ha and kube should exist in the deployments section.
And at the end of the getting started, I would add links to those documents in deployments section.

In my mind, the profile of people I picture who will use this documentation is someone knowing nothing about Authelia and potentially nothing about CS either. This person just wants to see how Authelia looks like not based on README pictures and how to use it for real. The basic use case that I can picture is the protection of a ownCloud or a personal website but the user does not necessarily know what is DNS, reverse proxies, ports. She just ran docker run owncloud that she read in a blog post or something.

So here, in that Getting Started, I'd rather take the user by the hand and show what Authelia looks like for real in the browser by taking the user to protected websites (one single-factor and one two-factor as you did is great) and eventually show how to customize one ACL rule. This will give the best overview that will, or not, ring a bell to the user. Consequently, she will, or not, proceed with lite or HA deployment. That's why I think those should be mentioned at the end and we should provide a link to their related deployment documentation.

Do you see what I mean?

README.md
```diff
@@ -79,10 +79,24 @@ Docker or on top of [Kubernetes].
 
 ## Getting Started
 
-You can start off with
+You can start utilising Authelia with the provided `docker-compose` bundles:
 
-    git clone https://github.com/authelia/authelia.git && cd authelia
-    source bootstrap.sh
+##### Local
```
clems4ever: In order to avoid maintaining twice the same content, I'd rather redirect the user to the Getting Started documentation directly. If we really want to keep some explanations here, which would make sense, that would be something like:

```
- `git clone https://github.com/authelia/authelia.git`
- `cd authelia/compose/local`
- `docker-compose up -d`

Use user `john` with password `password` to visit `https://traefik.example.com` protected with two factors or `https://singlefactor.example.com` protected with only one factor.
```

docs/getting-started.md
```diff
@@ -6,10 +6,58 @@ nav_order: 2
...
+
+You can now visit the following locations; replace example.com with the domain you specified in the setup script:
+- https://public.example.com - Bypasses Authelia
+- https://traefik.example.com - Secured with Authelia one-factor authentication
+- https://secure.example.com - Secured with Authelia two-factor authentication (see note below)
+
+To visit https://secure.example.com you will need to register an OTP device, the email/link to generate your QR code will be in stored in: `./authelia/notification.txt`.
+Upon registering, you can grab this link easily by running the following command: `grep -Eo '"https://.*" ' ./authelia/notification.txt`.
+
+## Reverse Proxy
```
clems4ever:
```suggestion
## Deployment
```

docs/getting-started.md
```diff
@@ -6,10 +6,58 @@ nav_order: 2
...
+
+Documentation for deploying a reverse proxy collaborating with Authelia is available
```
clems4ever: Information about the reverse proxies will be given in the deployment documentation so we don't care in the getting started. We should simply redirect the user to the deployment doc. For instance this section could contain:

`So you're convinced that Authelia is what you need. You can head to the deployment documentation.
Some recipes have been crafted for helping with the bootstrap of your environment. You can choose between a lite deployment which is deployment advised for a single server setup. However, this setup just does not scale. If you want a full environment that can scale out, use the HA or Kubernetes deployment documentation.`

docs/getting-started.md
```diff
@@ -6,10 +6,58 @@ nav_order: 2
...
+[here](./supported-proxies/index.md).
+
+## FAQ
+
+### Can you give more details on why this is not suitable for production environments?
```
clems4ever: This question should be in the FAQ of lite bundle only. You can remove it from the getting started.

docs/getting-started.md
```diff
@@ -6,10 +6,58 @@ nav_order: 2
...
+
+This documentation gives instructions that will make **Authelia** non
+resilient to failures and non scalable by preventing you from running multiple
+instances of the application. This means that **Authelia** won't be able to distribute
+the load across multiple servers and it will prevent failover in case of a
+crash or an hardware issue. Moreover, users will be logged out every time
+Authelia restarts.
+
+## Development workflow
```
clems4ever: This section should be in https://docs.authelia.com/contributing/.

docs/getting-started.md
```diff
@@ -20,7 +68,7 @@ Then, start the *Standalone* [suite].
 A [suite] is kind of a virtual environment for running Authelia in a complete ecosystem.
 If you want more details please read the related [documentation](./contributing/suites.md).
 
-## Test it!
+### Test it!
```
clems4ever: Just remove that section, it became useless.

docs/getting-started.md
clems4ever: Given that there was only one question in the FAQ, just remove it.

README.md
```diff
@@ -79,10 +79,19 @@ Docker or on top of [Kubernetes].
...
+The Local compose bundle is intended for scenarios where the server will not be exposed to the internet, domains will be defined in the local hosts file and self-signed certificates will be utilised.
```
clems4ever:
```suggestion
The Local compose bundle is intended to test Authelia without worrying about configuration. It's meant to be used for scenarios where the server is not exposed to the internet. Domains will be defined in the local hosts file and self-signed certificates will be utilised.

File: docs/getting-started.md [hunk 18]
```diff
@@ -6,10 +6,58 @@ nav_order: 2
...
+Connection details to Redis are optional. If not provided, sessions will be stored in
+memory instead. This has the inconvenience of logging out users every time Authelia restarts.
+
+## Steps
+
+These commands are intended to be run sequentially:
+
+- `git clone https://github.com/authelia/authelia.git`
+- `cd authelia/compose/local`
+- `sudo ./setup.sh`
```
clems4ever: Add a comment explaining why you need sudo here.

README.md
```diff
@@ -79,10 +79,20 @@ Docker or on top of [Kubernetes].
 
 ## Getting Started
 
-You can start off with
+You can start utilising Authelia with the provided `docker-compose` bundles:
 
-    git clone https://github.com/authelia/authelia.git && cd authelia
-    source bootstrap.sh
+##### Local
```
clems4ever: Add the corresponding link to the documentation for each section.

README.md
```diff
@@ -79,10 +79,20 @@ Docker or on top of [Kubernetes].
...
+
+##### Lite
+The Lite compose bundle is intended for scenarios where the server will be exposed to the internet, domains and DNS will need to be setup accordingly and certificates will be generated through LetsEncrypt.
+The Lite element refers to minimal external dependencies; File based user storage, SQLite based configuration storage. In this configuration, the service will not scale well.
+
+##### Full
+The Full compose bundle is intended for scenarios where the server will be exposed to the internet, domains and DNS will need to be setup accordingly and certificates will be generated through LetsEncrypt.
+The Full element refers to a scalable setup which includes external dependencies; LDAP based user storage, Database based configuration storage (MariaDB, MySQL or Postgres). 
 
 If you want to go further, please read [Getting Started](https://docs.authelia.com/getting-started).
```
clems4ever: You can remove this line since the link will be in each section.

PR #649, opened 2020-02-26
[BUGFIX] Fix development workflow.

clems4ever:
Suites cannot be run locally anymore following the move of the example
directory into the suites directory.


internal/suites/example/compose/authelia/Dockerfile.backend
```diff
@@ -6,7 +6,7 @@ ARG GROUP_ID
 # gcc and musl-dev are required for building go-sqlite3
 RUN addgroup --gid ${GROUP_ID} dev && \
     adduser --uid ${USER_ID} -G dev -D dev && \
-    apk --no-cache add gcc musl-dev
+    apk --no-cache add gcc musl-dev git
```
clems4ever: This is needed because reflex does not work anymore without it apparently.

nightah: Interesting and I was going to ask why this was in here 😁.

This likely would have also been causing issues for #610 right?

PR #653, opened 2020-02-27
[FEATURE] Redis DB Index Selection

james-d-elliott:
- Allow users to specify the DB number
- This is so users who use their redis for multiple purposes can have clear demarcation between their data

This is a simple change, works as expected. I haven't updated docs since we're moving to GitHub Pages but will look to make a commit there too. I added to unit tests as you'll see to verify the behavior of both no configuration and a value other than zero.

### DIFF
--- a/config.template.yml
+++ b/config.template.yml
@@ -223,6 +223,8 @@ session:
     port: 6379
     # This secret can also be set using the env variables AUTHELIA_SESSION_REDIS_PASSWORD
     password: authelia
+    # This is the Redis DB Index https://redis.io/commands/select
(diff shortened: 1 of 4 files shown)


nightah:
We should be documenting an example in the config template even if it’s commented.

What is the default behaviour if no DbNumber is provided? Does everything default to 0?

james-d-elliott:
Yeah, the default value for it is 0, see my assert.Equal(t, 0, pConfig.DbNumber) which is the first line change in the tests. You're right about the config, let me do that quickly.

PR #658, opened 2020-02-28
[Buildkite] Enable automatic retries for failed github artifact step

nightah:
This is to handle failures which may occur when attempting to upload assets, per: https://buildkite.com/authelia/authelia/builds/465#537f931f-efc3-4f7b-9527-c927c1425a52.

### DIFF
--- a/.buildkite/deployment.yml
+++ b/.buildkite/deployment.yml
@@ -15,6 +15,8 @@ steps:
 
   - label: ":github: Deploy Artifacts"
     command: ".buildkite/steps/ghartifacts.sh"
+    retry:
+      automatic: true
     agents:
       upload: "fast"
     key: "artifacts"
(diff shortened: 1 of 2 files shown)


.buildkite/deployment.yml
```diff
@@ -15,6 +15,8 @@ steps:
 
   - label: ":github: Deploy Artifacts"
     command: ".buildkite/steps/ghartifacts.sh"
+    retry:
```
clems4ever: how many times this will be retried before eventually failing?

nightah: Default behaviour for automatic retries if no attempts are specified is an additional 2 retries for a total of 3 (1 initial failure and 2 additional attempts).

clems4ever: :+1:

PR #664, opened 2020-02-29
[MISC] Add Detailed DUO Push Logging

### LINKED ISSUE #650: Duo Push - No Logging When Duo Has no Matching User
I'm a little new to Authelia so I'm probably configuring something wrong.

I registered for new account on Duo.  The name associated with the duo account is the same name that I use to authenticate with Authelia.  I'm using however a users.yml file for authentication -- and not LDAP.

After creating Duo Account and choosing Applications->Partner Auth API, I received three credentials 
1. Integration Key
2. Secret Key
3. API Hostname

I'm using authelia:latest docker image controlled with docker-compose.
To the configuration.yml file I've added:
duo_api:
  hostname: <hostname from #3 above>
  integration_key: <Integration Key from #1 above>
  secret_key: <Secret Key from #2 above>

I'm getting this error however with the authelia portal 
Screen Shot 2020-02-26 at 12 07 05 PM

Not sure how to address problems.  I looked in nginx athelia access and error log files, but didn't see anything mentioned in the log.  Log level is info. 

Thanks for any help.

james-d-elliott:
- Added trace logging for all response data from the DUO API
- Added warning messages on auth failures
- Added debug logging when DUO auth begins

When this PR is merged it will close #650


internal/handlers/handler_sign_duo.go
```diff
@@ -21,11 +22,15 @@ func SecondFactorDuoPost(duoAPI duo.API) middlewares.RequestHandler {
 		}
 
 		userSession := ctx.GetSession()
+		log := logging.Logger()
+		remoteIP := ctx.RemoteIP().String()
+
+		log.Debugf("Starting 2FA Duo Push Auth Attempt for %s from IP %s", userSession.Username, remoteIP)
```
clems4ever: since you have an AutheliaCtx at hand you can use the dedicated request logger like here: https://github.com/authelia/authelia/blob/a268a16c6e2cd91bbd3513d2851047d3cc6b8cbc/internal/handlers/handler_firstfactor.go#L45

internal/duo/duo.go
```diff
@@ -21,6 +23,7 @@ func (d *APIImpl) Call(values url.Values) (*Response, error) {
 	if err != nil {
 		return nil, err
 	}
+	logging.Logger().Tracef("Duo Auth Response Data: %s", string(responseBytes))
```
clems4ever: Use the request logger to correlate logs with an actual HTTP request, this will be useful to trace the error in case Authelia is highly used.

clems4ever: See how to do it in my comment below.

internal/handlers/handler_sign_duo.go
james-d-elliott: Done!

internal/duo/duo.go
james-d-elliott: Think this is resolved, fixed and pushed :)

PR #685, opened 2020-03-06
[MISC] Update Golang and QEMU to v1.14.0 and v4.2.0-6 respectively

nightah:
Housekeeping updates.

Also add some fixes around Argon2id.


clems4ever:
Why are there commits about argon2 in this PR? Rebase issue?

clems4ever:
LGTM

PR #687, opened 2020-03-06
[BUGFIX] Set username retrieved from authentication backend in session.

clems4ever:
In some setups, binding is case insensitive but Authelia is case
sensitive and therefore need the actual username as stored in the
authentication backend in order for Authelia to work correctly.

### DIFF
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -9,13 +9,12 @@ breaking changes and about what you should do to overcome those changes.
 ## Breaking in v4.7.0
 
 * `logs_level` configuration key has been renamed to `log_level`.
-* `users_filter` was a pattern of research for a given user with {0} matcher replaced by
+* `users_filter` was a search pattern for a given user with the `{0}` matcher replaced with the
 actual username. In v4.7.0, `username_attribute` has been introduced. Consequently, the computed
-user filter used by the LDAP search finding the user is a conjonction of a filter based
-on the username attribute and the filter provided in `users_filter`. Concretely, `users_filter` now
-reduces the scope of users targeted by the LDAP search query. For instance if
-`username_attribute` is set to `uid` and `users_filter` is set to `(objectClass=user)` then
-the computed filter is `(&(uid=john)(objectClass=user))`.
+user filter utilised by the LDAP search query is a combination of filters based on the
+`username_attribute` and `users_filter`. `users_filter` now reduces the scope of users targeted by
+the LDAP search query. For instance if `username_attribute` is set to `uid` and `users_filter` is
+set to `(objectClass=person)` then the computed filter is `(&(uid=john)(objectClass=person))`.
 
 ## Breaking in v4.0.0
 
--- a/BREAKING.md
+++ b/BREAKING.md
@@ -8,7 +8,14 @@ breaking changes and about what you should do to overcome those changes.
 
 ## Breaking in v4.7.0
 
-`logs_level` configuration key has been renamed to `log_level`.
+* `logs_level` configuration key has been renamed to `log_level`.
+* `users_filter` was a pattern of research for a given user with {0} matcher replaced by
+actual username. In v4.7.0, `username_attribute` has been introduced. Consequently, the computed
+user filter used by the LDAP search finding the user is a conjonction of a filter based
+on the username attribute and the filter provided in `users_filter`. Concretely, `users_filter` now
+reduces the scope of users targeted by the LDAP search query. For instance if
+`username_attribute` is set to `uid` and `users_filter` is set to `(objectClass=user)` then
+the computed filter is `(&(uid=john)(objectClass=user))`.
 
 ## Breaking in v4.0.0
 
--- a/config.template.yml
+++ b/config.template.yml
@@ -74,7 +74,7 @@ authentication_backend:
     # case insensitive search queries: #561).
     # Microsoft Active Directory usually uses 'sAMAccountName'
     # OpenLDAP usually uses 'uid'
-    username_attribute: cn
+    username_attribute: uid
     # An additional dn to define the scope to all users
     additional_users_dn: ou=users
     # The users filter used to find the user DN
--- a/config.template.yml
+++ b/config.template.yml
@@ -56,7 +56,7 @@ duo_api:
 # and retrieve information such as email address and groups
 # users belong to.
 #
-# There are two supported backends: `ldap` and `file`.
+# There are two supported backends: 'ldap' and 'file'.
 authentication_backend:
   # LDAP backend configuration.
   #
@@ -82,12 +82,14 @@ authentication_backend:
     # An additional dn to define the scope to all users
     additional_users_dn: ou=users
 
-    # This attribute is optional. The user filter used in the LDAP search query
-    # is a conjonction of this filter and a filter based on the username attribute.
-    # Concretely, this filter is used to reduce the scope of users targeted by the LDAP
-    # search query.
-    # For instance, if the username attribute is set to uid, the computed filter is
+    # This attribute is optional. The user filter used in the LDAP search queries
+    # is a combination of this filter and the username attribute.
+    # This filter is used to reduce the scope of users targeted by the LDAP search query.
+    # For instance, if the username attribute is set to 'uid', the computed filter is
     # (&(uid=<username>)(&(objectCategory=person)(objectClass=user)))
+    # Recommended settings are as follows:
+    # Microsoft Active Directory '(&(objectCategory=person)(objectClass=user))'
+    # OpenLDAP '(objectClass=person)' or '(objectClass=inetOrgPerson)'
     users_filter: (&(objectCategory=person)(objectClass=user))
 
     # An additional dn to define the scope of groups
@@ -137,7 +139,7 @@ authentication_backend:
 # Access control is a list of rules defining the authorizations applied for one
 # resource to users or group of users.
 #
-# If 'access_control' is not defined, ACL rules are disabled and the `bypass`
+# If 'access_control' is not defined, ACL rules are disabled and the 'bypass'
 # rule is applied, i.e., access is allowed to anyone. Otherwise restrictions follow
 # the rules defined.
 #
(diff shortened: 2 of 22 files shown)


internal/authentication/ldap_user_provider.go
```diff
@@ -114,71 +120,65 @@ func (p *LDAPUserProvider) getUserAttribute(conn LDAPConnection, username string
 		baseDN = p.configuration.AdditionalUsersDN + "," + baseDN
 	}
 
+	attributes := []string{"dn",
+		p.configuration.MailAttribute,
+		p.configuration.UsernameAttribute}
```
james-d-elliott: I was going to suggest doing multi-attribute lookups, nice!

internal/authentication/ldap_user_provider.go
```diff
@@ -75,12 +75,12 @@ func (p *LDAPUserProvider) CheckUserPassword(username string, password string) (
 	}
 	defer adminClient.Close()
 
-	userDN, err := p.getUserDN(adminClient, username)
+	profile, err := p.getUserProfile(adminClient, username)
```
james-d-elliott: getUserProfile actually doesn't use the ldap connection param you're passing to it "adminClient". It reconnects with a new connection, I doubt this is intentional.

clems4ever: good catch!

clems4ever: Should be fixed now.

internal/authentication/ldap_user_provider.go
clems4ever: :)

config.template.yml
```diff
@@ -68,18 +68,22 @@ authentication_backend:
     skip_verify: false
     # The base dn for every entries
     base_dn: dc=example,dc=com
+    # The attribute holding the username of the user (introduced to handle
+    # case insensitive search queries: #561).
+    # Microsoft Active Directory usually uses 'sAMAccountName'
+    # OpenLDAP usually uses 'uid'
+    username_attribute: cn
```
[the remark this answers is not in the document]
clems4ever: default to `uid` instead of `cn`.

clems4ever:
> I think now we have this config value we may want to consider changing how we format the usersFilter [here](https://github.com/authelia/authelia/blob/59838e436e402df6b664b258556486459618803b/internal/authentication/ldap_user_provider.go#L117) to be something along the lines of this:
> 
> usersFilter := fmt.Sprintf("(&(%s=%s)%s)", p.configuration.UsernameAttribute, username, p.configuration.UsersFilter)
> 
> Basically my idea is to add the UsernameAttribute and the entered value into the search query. I'm not 100% sure on the way Go operates with this but the actual ldap filter should end up looking like this for AD:
> 
> Username entered: john
> UsernameAttribute: sAMAccountName
> UsersFilter: (&(objectCategory=person)(objectClass=user))
> 
> Result usersFilter: (&(sAMAccountName=john)(&(objectCategory=person)(objectClass=user)))

That sounds reasonable and we should probably do the same for groups then. It means introducing a memberAttribute or GroupAttribute and the groups filter could be

```
groupsFilter := fmt.Sprintf("(&(%s=%s)%s)", p.configuration.GroupAttribute, profile.userDN, p.configuration.GroupsFilter)
```

What do you think @james-d-elliott ?

james-d-elliott:
> 
> 
> > I think now we have this config value we may want to consider changing how we format the usersFilter [here](https://github.com/authelia/authelia/blob/59838e436e402df6b664b258556486459618803b/internal/authentication/ldap_user_provider.go#L117) to be something along the lines of this:
> > usersFilter := fmt.Sprintf("(&(%s=%s)%s)", p.configuration.UsernameAttribute, username, p.configuration.UsersFilter)
> > Basically my idea is to add the UsernameAttribute and the entered value into the search query. I'm not 100% sure on the way Go operates with this but the actual ldap filter should end up looking like this for AD:
> > Username entered: john
> > UsernameAttribute: sAMAccountName
> > UsersFilter: (&(objectCategory=person)(objectClass=user))
> > Result usersFilter: (&(sAMAccountName=john)(&(objectCategory=person)(objectClass=user)))
> 
> That sounds reasonable and we should probably do the same for groups then. It means introducing a memberAttribute or GroupAttribute and the groups filter could be
> 
> ```
> groupsFilter := fmt.Sprintf("(&(%s=%s)%s)", p.configuration.GroupAttribute, profile.userDN, p.configuration.GroupsFilter)
> ```
> 
> What do you think @james-d-elliott ?

I've given this a lot of thought and I'm 50/50 on this idea, leaning towards it's not something we should do. While I think it makes absolute sense for user filtering, the reasons for that are:
- the username attribute is required for us to do backend things
- the attribute defined must match the username we enter 
- we're already making a breaking change
- it will overall reduce configuration since we're not repeating ourselves
- it technically makes users_filter optional
- is a breaking change but has an auxiliary purpose (fixing a bug)

The reasons for changing it for groups however are as follows:
- it may be slightly easier for new users
- it will work as intended
- it technically may make groups_filter optional in some cases

The reasons against:
- it's another breaking change and another configuration variable
- this change is not necessary for an auxiliary reason like the users change

clems4ever:
> > > I think now we have this config value we may want to consider changing how we format the usersFilter [here](https://github.com/authelia/authelia/blob/59838e436e402df6b664b258556486459618803b/internal/authentication/ldap_user_provider.go#L117) to be something along the lines of this:
> > > usersFilter := fmt.Sprintf("(&(%s=%s)%s)", p.configuration.UsernameAttribute, username, p.configuration.UsersFilter)
> > > Basically my idea is to add the UsernameAttribute and the entered value into the search query. I'm not 100% sure on the way Go operates with this but the actual ldap filter should end up looking like this for AD:
> > > Username entered: john
> > > UsernameAttribute: sAMAccountName
> > > UsersFilter: (&(objectCategory=person)(objectClass=user))
> > > Result usersFilter: (&(sAMAccountName=john)(&(objectCategory=person)(objectClass=user)))
> > 
> > 
> > That sounds reasonable and we should probably do the same for groups then. It means introducing a memberAttribute or GroupAttribute and the groups filter could be
> > ```
> > groupsFilter := fmt.Sprintf("(&(%s=%s)%s)", p.configuration.GroupAttribute, profile.userDN, p.configuration.GroupsFilter)
> > ```
> > 
> > 
> > What do you think @james-d-elliott ?
> 
> I've given this a lot of thought and I'm 50/50 on this idea, leaning towards it's not something we should do. While I think it makes absolute sense for user filtering, the reasons for that are:
> 
> * the username attribute is required for us to do backend things
> * the attribute defined must match the username we enter
> * we're already making a breaking change
> * it will overall reduce configuration since we're not repeating ourselves
> * it technically makes users_filter optional
> * is a breaking change but has an auxiliary purpose (fixing a bug)
> 
> The reasons for changing it for groups however are as follows:
> 
> * it may be slightly easier for new users
> * it will work as intended
> * it technically may make groups_filter optional in some cases
> 
> The reasons against:
> 
> * it's another breaking change and another configuration variable
> * this change is not necessary for an auxiliary reason like the users change
> * is possibly less configurable in certain circumstances (not sure)

I'm fine with the reasoning so we'll keep groups_filter as it is, at least for the moment.

BREAKING.md
```diff
@@ -8,7 +8,14 @@ breaking changes and about what you should do to overcome those changes.
 
 ## Breaking in v4.7.0
 
-`logs_level` configuration key has been renamed to `log_level`.
+* `logs_level` configuration key has been renamed to `log_level`.
+* `users_filter` was a pattern of research for a given user with {0} matcher replaced by
```
nightah:
```suggestion
* `users_filter` was a search pattern for a given user with the `{0}` matcher replaced by
```

BREAKING.md
```diff
@@ -8,7 +8,14 @@ breaking changes and about what you should do to overcome those changes.
...
+actual username. In v4.7.0, `username_attribute` has been introduced. Consequently, the computed
```
nightah:
```suggestion
the actual username. In v4.7.0, `username_attribute` has been introduced. Consequently, the computed
```

BREAKING.md
```diff
@@ -8,7 +8,14 @@ breaking changes and about what you should do to overcome those changes.
...
+user filter used by the LDAP search finding the user is a conjonction of a filter based
```
nightah:
```suggestion
user filter utilised by the LDAP search is a combination of a fields based
```

BREAKING.md
```diff
@@ -8,7 +8,14 @@ breaking changes and about what you should do to overcome those changes.
...
+on the username attribute and the filter provided in `users_filter`. Concretely, `users_filter` now
```
nightah:
```suggestion
on the `username_attribute` and the filter provided in `users_filter`. Effectively, `users_filter` now
```

config.template.yml
```diff
@@ -66,32 +66,46 @@ authentication_backend:
...
     username_attribute: uid
+    
     # An additional dn to define the scope to all users
     additional_users_dn: ou=users
-    # The users filter used to find the user DN
-    # {0} is a matcher replaced by username.
-    users_filter: (uid={0})
+
+    # This attribute is optional. The user filter used in the LDAP search query
+    # is a conjonction of this filter and a filter based on the username attribute.
```
nightah:
```suggestion
    # is a combination of this filter and a filter based on the username attribute.
```

config.template.yml
```diff
@@ -66,32 +66,46 @@ authentication_backend:
...
+    # Concretely, this filter is used to reduce the scope of users targeted by the LDAP
```
nightah:
```suggestion
    # This filter is used to reduce the scope of users targeted by the LDAP
```

docs/configuration/authentication/ldap.md
```diff
@@ -35,9 +35,13 @@ authentication_backend:
         # An additional dn to define the scope to all users
         additional_users_dn: ou=users
         
-        # The users filter used to find the user DN
-        # {0} is a matcher replaced by username.
-        users_filter: (uid={0})
+        # This attribute is optional. The user filter used in the LDAP search query
+        # is a conjonction of this filter and a filter based on the username attribute.
```
nightah:
```suggestion
        # is a combination of this filter and a filter based on the `username_attribute`.
```

docs/configuration/authentication/ldap.md
```diff
@@ -35,9 +35,13 @@ authentication_backend:
...
+        # Concretely, this filter is used to reduce the scope of users targeted by the LDAP
```
nightah:
```suggestion
        # Effectively, this filter is used to reduce the scope of users targeted by the LDAP
```

docs/configuration/authentication/ldap.md
```diff
@@ -35,9 +35,13 @@ authentication_backend:
...
+        # search query.
+        # For instance, if the username attribute is set to uid, the computed filter is
```
nightah:
```suggestion
        # For instance, if the `username_attribute` is set to `uid`, the computed filter is
```

config.template.yml
```diff
@@ -66,28 +66,48 @@ authentication_backend:
...
+
+    # This attribute is optional. The user filter used in the LDAP search queries
+    # is a combination of this filter and the username attribute.
+    # This filter is used to reduce the scope of users targeted by the LDAP search query.
+    # For instance, if the username attribute is set to 'uid', the computed filter is
+    # (&(uid=<username>)(&(objectCategory=person)(objectClass=user)))
+    # Recommended settings are as follows:
+    # Microsoft Active Directory '(&(objectCategory=person)(objectClass=user))'
+    # OpenLDAP '(objectClass=person)' or '(objectClass=inetOrgPerson)'
+    users_filter: (&(objectCategory=person)(objectClass=user))
```
james-d-elliott: Might want to make this blank by default since we have the examples. Additionally the rest of the template uses a default of OpenLDAP instead of AD, but this is specific to AD.

docs/configuration/authentication/ldap.md
```diff
@@ -25,22 +25,33 @@ authentication_backend:
...
-        users_filter: (cn={0})
+        # This attribute is optional. The user filter used in the LDAP search queries
+        # is a combination of this filter and the username attribute.
+        # This filter is used to reduce the scope of users targeted by the LDAP search query.
+        # For instance, if the username attribute is set to 'uid', the computed filter is
+        # (&(uid=<username>)(&(objectCategory=person)(objectClass=user)))
+        # Recommended settings are as follows:
+        # Microsoft Active Directory '(&(objectCategory=person)(objectClass=user))'
+        # OpenLDAP '(objectClass=person)' or '(objectClass=inetOrgPerson)'
+        users_filter: (&(objectCategory=person)(objectClass=user))
```
james-d-elliott: Same as above.

config.template.yml
nightah: The only reason the example is there is so the lines above the recommendation makes sense.
Can definitely comment the filter though as a default.

PR #689, opened 2020-03-07
[FEATURE] HTTP method access control

### LINKED ISSUE #648: Failing http-post requests because of automatic preflight-requests
We are using traefik 2.x and authelia using forwardAuth in traefik, which is working fine for many internal websites now... 

But we have one website (behind traefik/authelia) which is doing a 
javascript-XMLHttpRequest() against an other http-rest-api-site 
(running behind traefik/authelia too) using:

```javascript
var url = 'https://<http-rest-api-site>/....';
var xhr = new XMLHttpRequest();
xhr.open("POST", url, true);
xhr.withCredentials = true;
xhr.setRequestHeader("Content-Type", "application/json");
...
```

This will cause a browser to send a http-options-request (so called preflight)
see https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS before the actual  http-post-request.
This http-options-request can't contain the credentials/cookies (by definition), 
so traefik/authelia will always responds with an HTTP-302 (redirect), which is not allowed as answer for a preflight so the browser will completely fail here ...

We are using other requests (get) too. These reqeusts are working fine, so I think we have
set everything related to CORS in a proper way ...

We are wondering how this problem can be solved?
- in general don't use authelia in front of the http-rest-api-site (using access-tokens instead)?
- define more complex traefik-rules to not use forwardAuth/authelia if a preflight/http-options-request is detected (how to define such a traefik-rule?)
- maybe a function in authelia to detect if an http-options-request is done?

Maybe I'm completely wrong here, but perhaps you have an idea here?

james-d-elliott:
This is a draft because it's more of a RFC rather than a ready to merge PR.

If merged this should fix #648

### DIFF
--- a/internal/authorization/authorizer_test.go
+++ b/internal/authorization/authorizer_test.go
@@ -32,13 +32,13 @@ func NewAuthorizerTester(config schema.AccessControlConfiguration) *AuthorizerTe
 	}
 }
 
-func (s *AuthorizerTester) CheckAuthorizations(t *testing.T, subject Subject, requestURI string, expectedLevel Level) {
+func (s *AuthorizerTester) CheckAuthorizations(t *testing.T, subject Subject, requestURI string, method []byte, expectedLevel Level) {
 	url, _ := url.ParseRequestURI(requestURI)
 	level := s.GetRequiredLevel(Subject{
 		Groups:   subject.Groups,
 		Username: subject.Username,
 		IP:       subject.IP,
-	}, *url)
+	}, *url, method)
 
 	assert.Equal(t, expectedLevel, level)
 }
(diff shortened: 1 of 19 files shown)


internal/authorization/method_matcher.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+var warningShown = false
+
+// isMethodMatching check whether request method is in the rules methods
+func isMethodMatching(method []byte, methods []string) (matching bool) {
+	log := logging.Logger()
+	// If no method is provided in the rule, we match any method
+	if len(methods) == 0 {
+		log.Trace("method matching skipped as no methods have been defined for the access_control rule")
```
clems4ever:
```suggestion
		log.Trace("Method matching skipped as no methods have been defined for the access_control rule")
```

internal/authorization/method_matcher.go
```diff
@@ -0,0 +1,35 @@
...
+		return true
+	}
+
+	if method == nil {
+		if !warningShown {
+			warningShown = true
+			log.Warn("rules have been configured with methods but your proxy does not seem to be" +
```
clems4ever:
```suggestion
			log.Warn("Rules have been configured with methods but your proxy does not seem to be" +
```

internal/authorization/method_matcher.go
```diff
@@ -0,0 +1,35 @@
...
+func isMethodMatching(method []byte, methods []string) (matching bool) {
+	log := logging.Logger()
+	// If no method is provided in the rule, we match any method
+	if len(methods) == 0 {
+		log.Trace("method matching skipped as no methods have been defined for the access_control rule")
+		return true
+	}
+
+	if method == nil {
+		if !warningShown {
```
clems4ever: Rather than logging a warning, you can check whether X-Forwarded-Method is required by checking if at least one rule needs it. In that case displaying an error for every request is just fine. Otherwise, no message will be displayed.

internal/authorization/method_matcher.go
```diff
@@ -0,0 +1,35 @@
...
+			warningShown = true
+			log.Warn("rules have been configured with methods but your proxy does not seem to be" +
+				"sending the X-Forwarded-Method header, these rules will effectively be disabled")
+		}
+		return false
+	}
+
+	methodStr := string(method)
+	if !utils.IsStringInSlice(methodStr, schema.HTTPRequestMethods) {
+		log.Warnf("an invalid X-Forwarded-Method header was received, its value was %s", methodStr)
```
clems4ever:
```suggestion
		log.Warnf("An invalid X-Forwarded-Method header was received, its value was %s", methodStr)
```

internal/authorization/authorizer_test.go
```diff
@@ -194,8 +194,34 @@ func (s *AuthorizerSuite) TestShouldCheckGroupMatching() {
 		}).
 		Build()
 
-	tester.CheckAuthorizations(s.T(), John, "https://protected.example.com/", Bypass)
-	tester.CheckAuthorizations(s.T(), Bob, "https://protected.example.com/", Denied)
+	tester.CheckAuthorizations(s.T(), John, "https://protected.example.com/", []byte("GET"), Bypass)
+	tester.CheckAuthorizations(s.T(), Bob, "https://protected.example.com/", []byte("GET"), Denied)
+}
```
clems4ever: Can you also add a test checking the authorization module works when no `X-Forwarded-Method` is provided, meaning CheckAuthorizations is called with `nil` instead of `[]byte("GET")`.

PR #690, opened 2020-03-07
[Buildkite] Ignore build and deploy steps for docs/* only based commits

nightah:
This PR will skip code based CI steps if only files in the `docs/` directory have been edited.
This is to ensure we do not have to wait 10-20 minutes for document changes to propagate.

The pipelines had to be converted from static `.yml` pipelines to dynamic pipelines like our other dynamic pipelines which are utilised for individual steps to be able to achieve this.
While doing so I've also taken the opportunity to convert all of the existing steps and the new dynamic pipelines to heredocs for readability.


nightah:
So for point 1 it’s true that any commit which starts with [DOCS] would skip non-docs publishing steps. But this would only really count for branch builds or PRs.

The reason being that when a PR is made, as part of the review we will be able to determine if the PR commit message needs to be changed if there’s code changes.

Point 2 isn’t correct, the docs step has not been made conditional, we perform that on every push to master and will continue to. This change only looks to make the other steps conditional.

clems4ever:
@nightah , you're right for point 2, my bad.

> The reason being that when a PR is made, as part of the review we will be able to determine if the PR commit message needs to be changed if there’s code changes.

I think it's just safer to not rely on a human changing the commit message according to content of the PR. It could be a problem for newcomers who don't know the rules.

nightah:
I don’t disagree on the fact that we shouldn’t rely on humans. I guess even without knowing the coding practices with tagging, it effectively means that PRs and such will likely still result in a build.

Realistically the people that will need to change titles are members of the org, which we already do today anyway.

So it’s probably extremely common that someone will submit changes and they will result in a build. When there’s a PR, based on our review it’ll be up to the team to vet that last element and if it’s just a docs change, rename accordingly.

If you’re still not convinced then I’m happy to try look at alternatives, I just don’t see a native way to determine what’s changed and building that into our conditionals. It might need to be a script that will then determine next steps.

nightah:
Just for reference our CI tooling also had to be updated to utilise gnu `sed` as the busybox variant included with alpine did not have all the options to be able to parse our diff inspection command.

PR #701, opened 2020-03-10
[MISC] Adjust TLS log messages

nightah:


internal/server/server.go
```diff
@@ -110,7 +110,7 @@ func StartServer(configuration schema.Configuration, providers middlewares.Provi
 			configuration.TLSCert, configuration.TLSKey,
 			middlewares.LogRequestMiddleware(router.Handler)))
 	} else {
-		logging.Logger().Infof("Authelia is listening on %s and do not use TLS", addrPattern)
+		logging.Logger().Infof("Authelia is listening on %s and does not use TLS", addrPattern)
```
james-d-elliott: I saw this too, was thinking the more correct wording would be:

"Authelia is listening for TLS connections on port %s"
"Authelia is listening for non-TLS connections on port %s"

nightah: Great suggestion, agreed!

PR #804, opened 2020-03-31
[Buildkite] Adjust CI bypasses

nightah:
This change will allow us to bypass CI/CD steps for changes that do not relate to code, so far this includes the following:
- `BREAKING.md`
- `CONTRIBUTING.md`
- `README.md`
- `docs/*`

Any additional files or directories can be added to the existing regex to expand in future.

The only consideration that needs to be made here is if we merge this change, we will need to move our release management trigger out of the `README.md`. Along with this change it would no longer trigger a proper deployment.

@clems4ever, @james-d-elliott we need to think about what the best candidate is outside of the aforementioned files to manage our releases.


james-d-elliott:
As far as release meta data I'll have to think on it.

PR #806, opened 2020-03-31
[FEATURE] Autofocus on authentication and OTP pages

### LINKED ISSUE #511: Set cursor focus to input fields
When the login page first loads, you can't start typing your username until you click with your mouse on the username field.
Simalarly, on the TOTP input page, you have to click with your mouse on the first digit box before you can start typing.
This slows down the workflow of logging in, it would be much better to have the focus set automatically so users can just start typing, then tab to the password field, then enter to login.

nightah:
This change sets the input focus on the first factor authentication and OTP pages.

The behaviour for the first factor authentication page has also been amended slightly, if an incorrect username or password is provided the password field will be cleared and set as the focus.

One thing to note is that the OTP page does not focus on any re-rendering and this is because the component doesn't handle focusing. This means that the OTP input only is auto-focused when you first visit it, if you enter an incorrect OTP there will be no focus.

Ideally we should be looking for a different library or writing a component for this ourselves in future.

Closes #511.


web/src/views/LoginPortal/FirstFactor/FirstFactorForm.tsx
```diff
@@ -28,6 +28,12 @@ export default function (props: Props) {
     const [password, setPassword] = useState("");
     const [passwordError, setPasswordError] = useState(false);
     const { createErrorNotification } = useNotifications();
+    const usernameRef = useRef() as MutableRefObject<HTMLInputElement>;
+    const passwordRef = useRef() as MutableRefObject<HTMLInputElement>;
+    useEffect(() => {
+        const timeout = setTimeout(() => usernameRef.current.focus(), 10);
```
clems4ever: As discussed I will have a look at how we can better fix it. In the meantime we use that PR to merge in the change of error message and the auto focus of the OTP field. We'll do the username field in another PR.

PR #811, opened 2020-04-01
[MISC] Fix dev workflow for Standalone suite

nightah:
Since the TLS change was implemented, the Standalone suite attempts to reach `authelia-frontend:3001`. This only works for the dist/CI based workflow as nginx is run in place of node and the respective ports are exposed.
This change will attempt `authelia-frontend:3001` initially for the Standalone suite but given the server is unreachable in the dev workflow, it will mark `authelia-frontend:3001` as unavailable for 180s and route all requests to `authelia-frontend:3000` instead.


clems4ever:
As I said, I'm doing the PR to use TLS in all cases so that we can remove the duplicated configs and simplify the whole thing. We won't need this change anymore.

clems4ever:
It's seems more complex than what I thought to use TLS in the backend when using HAProxy so I'll merge in your PR in the meantime @nightah .

nightah:
No problems. I don’t think swapping everything to TLS is super high priority anyway. This change will work fine for both instances.

PR #814, opened 2020-04-02
[FEATURE] Include darwin based binaries for OSX

nightah:
This PR provides binary artifacts for Authelia which can be run directly from OSX.


cmd/authelia-scripts/cmd_docker.go
```diff
@@ -14,7 +14,7 @@ import (
 
 var arch string
 
-var supportedArch = []string{"amd64", "arm32v7", "arm64v8"}
+var supportedArch = []string{"amd64", "arm32v7", "arm64v8", "darwin"}
```
clems4ever, line 17:
```suggestion
var supportedArch = []string{"amd64", "arm32v7", "arm64v8", "darwin-amd64"}
```

.buildkite/steps/buildimages.sh
```diff
@@ -1,18 +1,39 @@
...
+      - "authelia-${BUILD_OS}-${BUILD_ARCH}.tar.gz.sha256"
     env:
       ARCH: "${BUILD_ARCH}"
-    key: "build-docker-${BUILD_ARCH}"
+      OS: "${BUILD_OS}"
+    key: "build-docker-${BUILD_OS}-${BUILD_ARCH}"
 EOF
+    done
+  else
+    for BUILD_ARCH in amd64; do
```
clems4ever, line 23: I'd rather use a hash table with the OS as the key and the list of architectures as values in order to avoid duplicating the code. It's also more generic since we won't have to change the code if we add new architectures for Darwin or add a new OS.

.buildkite/hooks/pre-artifact
```diff
@@ -8,11 +8,17 @@ if [[ $BUILDKITE_LABEL =~ ":docker: Build Image" ]]; then
...
-  sha256sum authelia-linux-"${ARCH}".tar.gz > authelia-linux-"${ARCH}".tar.gz.sha256
+  if [[ $OS == "linux" ]]; then
+    tar -czf authelia-"${OS}"-"${ARCH}".tar.gz authelia-"${OS}"-"${ARCH}" authelia.service config.template.yml public_html
+  else
+    tar -czf authelia-"${OS}"-"${ARCH}".tar.gz authelia-"${OS}"-"${ARCH}" config.template.yml public_html
+  fi
+  sha256sum authelia-"${OS}"-"${ARCH}".tar.gz > authelia-"${OS}"-"${ARCH}".tar.gz.sha256
   # Saving image for push to docker hub
-  docker save ${DOCKER_IMAGE} | zstdmt -T0 -12 > authelia-image-"${ARCH}".tar.zst
+  if [[ $BUILDKITE_LABEL != ":docker: Build Image [darwin]" ]]; then
```
clems4ever, line 21: Please add a comment to explain why we don't save darwin images in cache. It's clear to me but might not be for others.

cmd/authelia-scripts/cmd_docker.go
nightah: This would mean the Dockerfile also needs to be renamed. I went with only Darwin because I don’t believe Macs come in any other Architecture so we will only ever build amd64.

clems4ever: I actually don't know whether they support other architectures. No big deal if you want to keep it like this here. I just proposed to be explicit.

PR #829, opened 2020-04-06
[FEATURE] Session IP Validation

### LINKED ISSUE #595: Optional Cookie Validity Tied to Client IP (Theft Prevention)
I believe it may be a useful feature to consider. My thoughts are that having a configuration option (default disabled) to enable cookies to automatically be invalidated if the IP's don't match would be useful. We would need to record the IP of the user when they first authenticate against the cookie, and then when cookies are validated we'd have to additionally check the IP is the same.

#166 already covers cookie theft, this would just be an additional method someone setting up Authelia could use. The reason it'd be off by default is there is significant inconvenience with using this method, due to dynamic IP's etc.

james-d-elliott:
Hopefully I picked all the correct files, I accidentally did a gofmt -s -w ./

Closes #595

### DIFF
--- a/docs/security/measures.md
+++ b/docs/security/measures.md
@@ -117,9 +117,9 @@ what is 2 hours, and the soft limit or the time a user can be inactive for is 10
 
 You can also decrease the security by setting disable_ip_check to `true`. The IP check ensures
 that when a users remote IP address changes that their session is completely destroyed and they 
-must reauthenticate. This can help prevent stolen devices with saved credentials compremising
+must re-authenticate. This can help prevent stolen devices with saved credentials compromising
 security, as such it's recommended you do not alter this option. 
- 
+
 ### More protections measures with Nginx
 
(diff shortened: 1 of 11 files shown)


docs/security/measures.md
```diff
@@ -115,6 +115,11 @@ to 0 to disable this feature, and set an expiration of something like 2 hours an
 of 10 minutes. This means the hard limit or the time the session will be destroyed no matter
 what is 2 hours, and the soft limit or the time a user can be inactive for is 10 minutes. 
 
+You can also decrease the security by setting disable_ip_check to `true`. The IP check ensures
+that when a users remote IP address changes that their session is completely destroyed and they 
+must reauthenticate. This can help prevent stolen devices with saved credentials compremising
```
nightah, line 120:
```suggestion
must re-authenticate. This can help prevent stolen devices with saved credentials compromising
```

PR #830, opened 2020-04-06
[FEATURE] Customizable Email Subject

### LINKED ISSUE #578: Adjust email subject line
Hello,

Noticed that the SUBJECT in the emails that get sent when user makes an adjustment to his account are in the following syntax:

`[Authelia] Reset/Register bla bla`

Would it be possible/reasonable to adjust the string *Authelia* to the `totp: issuer` string in the config or maybe some other `mail : issuer` ?

Thanks!

james-d-elliott:
Closes #578


docs/configuration/notifier/smtp.md
```diff
@@ -12,7 +12,16 @@ nav_order: 2
 It can be configured as described below.
 
 ```yaml
+# Configuration of the notification system.
+#
+# Notifications are sent to users when they require a password reset, a u2f
+# registration or a TOTP registration.
+# Use only an available configuration: filesystem, gmail
```
nightah, line 19:
```suggestion
# Use only an available configuration: filesystem, smtp
```

Issue #841, opened 2020-04-09
BasicAuth restricted to Proxy-Authentication header


omar-nahhas:
There are some cases when the same service is accessed by humans and machines (like Prometheus). Currently, authelia allows the use of basic auth, but it is restricted to basic auth passed through the header [Proxy-Authorization](https://github.com/authelia/authelia/blob/master/internal/handlers/handler_verify.go#L232). 

The problem is that load load balancers consider that header as a [hop by hop header](https://www.freesoft.org/CIE/RFC/2068/143.htm) and is [not passed to the authentication service ](https://github.com/containous/traefik/blob/0b039499c65cf79cd72bd15ade33bbaee2e71d1a/middlewares/auth/forward.go#L56) when set by the originator of the request. 

In my opinion, authelia should allow basic auth using the Authorization header. I know this could expose authelia to a brute-force attack but reverse proxies like Treafic can be [configured](https://github.com/containous/traefik/blob/d7ed42d614fe5ba70b44465f17501bfe112e953f/docs/configuration/backends/kubernetes.md#L355) to strip that header off before passing it to the auth service.

clems4ever:
Hello, @omar-nahhas using Proxy-Authorization was used because I considered Authelia like an authentication proxy meaning two authentications could be encapsulated but beware that Authelia is not a proxy, so it does not forward the header to your backends, the LB does.

Regarding the link that you sent, Proxy-Authorization is not in the list, Proxy-Authenticate is.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authenticate
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization

I'm pretty certain there is a way to configure your loadbalancer to forward/transform the right header to the right backend. I'm pretty sure you can already handle your use case without even using your PR.

However we might still consider your PR for ease of configuration on the LB side if my mates are ok with that. But beware that your problem is likely a configuration problem, not a missing feature in Authelia.

james-d-elliott:
https://tools.ietf.org/html/rfc7235#section-4.4

https://github.com/vulcand/oxy/blob/master/forward/headers.go#L4-L38

Top link is  RFC for the header, second is the implementation of what is considered a Hop header by Traefik. Theoretically all removed from upstream requests. We could test that with whoami.

nightah:
[...]

PR #843, opened 2020-04-09
Fix basic authentication using "Authorization" header

omar-nahhas:
There are some cases when the same service is accessed by humans and machines (like Prometheus). Currently, authelia allows the use of basic auth, but it is restricted to basic auth passed through the header Proxy-Authorization.

The problem is that load load balancers consider that header as a hop by hop header and is not passed to the authentication service when set by the originator of the request.

In my opinion, authelia should allow basic auth using the Authorization header. I know this could expose authelia to a brute-force attack but reverse proxies like Treafic can be configured to strip that header off before passing it to the auth service if basic auth is not expected.

Fixes issue [841](https://github.com/authelia/authelia/issues/841)

### DIFF
--- a/docs/features/single-factor.md
+++ b/docs/features/single-factor.md
@@ -20,10 +20,18 @@ resources matching `app1.example.com/admin` with two factors.
 To know more about the configuration of the feature, please visit the
 documentation about the [configuration](../configuration/access-control.md).
 
+Authelia also supports machine to machine authentication. The authentication
+backed stay the same as UI driven authentication. But since there is not 
+user to input the credentials, those credentials will try to be extracted 
+from either the `Proxy-Authorization` header or the `Authorization` header.
 
 ## Proxy-Authorization header
 
-Authelia reads credentials from the header `Proxy-Authorization` instead of
-the usual `Authorization` header. This is because in some circumstances both Authelia
-and the application could require authentication in order to provide specific
-authorizations at the level of the application.
+By default, Authelia reads credentials from the header `Proxy-Authorization` 
+instead of the usual `Authorization` header. This header is considered a 
+[Hop header](https://tools.ietf.org/html/rfc7235#section-4.4) and many proxies 
+will remove the header before sending the authentication request to Authelia. 
+
+## Authorization header
+
+If the Proxy-Authorization header is not found, but the Authorization header is present, Authelia will try to use that header to extract username and password for authentication. This could cause issues when the `Authentication` header is not directed to Authelia. Still, to the destination application, in that case, the proxy in front of Authelia should be instructed to explicitly remove the headers before sending an authentication request to Authelia. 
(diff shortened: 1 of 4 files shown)


clems4ever:
Hello @omar-nahhas , thank you for that contribution.

omar-nahhas:
Thank you for your feedback @clems4ever and @nightah, our use case is simple. We have a set of Prometheus instances behind a set of traefic proxies ( we use Authelia as an authentication server, not as an authentication proxy). We want to protect the Prometheus instances with Authelia for user access, but at the same time, we need to allow Grafana instances (and other federated Prometheus instances) to scrape the protected Prometheus endpoints.

Currently, there is no apparent way to make this happen, as if we set the Proxy-Authentication header on the calling services, it is removed by Traffic with no way to route the header to the request sent to Authelia for authentication. If we set the Authentication proxy, it is ignored by Authelia.

Regarding this change, the more I think about this change, the more I feel like it is not good enough. Although, the use case is solid and I believe many people will benefit from it. I think more work needs to be done to mark this work as production-ready.  These are the concerns I have:

1 - This change will a breaking change for users who have basic auth on their applications on top of Authelia. Although having two authentication mechanisms is not the best user experience, I anticipate that this use-case may be relatively common. In order to not break those use cases, this change should be driven by configuration instead of convention, or at least being able to overridden by configuration. 

2 - I believe that the max retries lockout is not applied when using basic auth, if that is the case this could open the doors for DoS/brute force attacks.

3 - For API calls, having to contact the LDAP server on every call includes a delay that in many cases would not be acceptable. Although In my use-case is fine as we call the endpoint every 30 seconds, maybe some sort of caching could help in this case, maybe using the Redis service already present, I would cache usernames, hashes of the passwords and group belonging data for a configurable ammount of time.

What are your thoughts? Is this something the project would like to see implemented?

nightah:
> 3 - For API calls, having to contact the LDAP server on every call includes a delay that in many cases would not be acceptable. Although In my use-case is fine as we call the endpoint every 30 seconds, maybe some sort of caching could help in this case, maybe using the Redis service already present, I would cache usernames, hashes of the passwords and group belonging data for a configurable ammount of time.
> 
> What are your thoughts? Is this something the project would like to see implemented?

Contributions are definitely most welcome, I would suggest that you create an issue around a proposed feature to discuss or head over to [Matrix](https://riot.im/app/#/room/#authelia:matrix.org) the team is active as it's our primary chat platform.

Regarding the caching it's something we have definitely discussed at a high level in the past, it would also be useful/necessary so that we could determine groups a user belongs to without having to invalidate a session.

Today if you have ACLs based on LDAP groups and a group is added to a user in the LDAP, the session isn't made aware of that change. Having a cache for LDAP users/groups would make implementation of a couple other features and quality of life changes significantly easier.

PR #851, opened 2020-04-10
[BUGFIX] Fix indenting on lite compose bundle

xenuser:
docker-compose up -d no longer complains about the public service because of wrong indentation


nightah:
Thanks @xenuser, seems I didn't realise I messed up the indenting in two places :)!

PR #865, opened 2020-04-13
[BUGFIX] Allow username_attribute == mail_attribute

Sohalt:
I don't have a separate `username_attribute` and instead identify all users by their email address (I know this is not ideal, but that's just how it is). This would result in an empty name, which lead to authentication failing ("An anonymous user cannot be authenticated.")
This commit allows using the same attribute for `username_attribute` and `mail_attribute`.


nightah:
> I don't have a separate `username_attribute` and instead identify all users by their email address (I know this is not ideal, but that's just how it is).

It's not that this isn't just ideal, if you don't conform to a standard/RFC this could lead to issues down the track with other features. Is something specific stopping you from being able to do so?

james-d-elliott:
Actually, that's a good question. If it's just you want users to be able to login via email, we can solve that easily as long as they also have a username.

Sohalt:
> It's not that this isn't just ideal, if you don't conform to a standard/RFC this could lead to issues down the track with other features.

I think my setup is still standard compliant. I'm using `inetOrgPerson`, which only requires `cn`, `objectClass`, and `cn` (https://tools.ietf.org/html/rfc2798#section-3)
All users have a `cn` and a `sn`. Neither of  them has to be unique however, so I can't use them for authentication.

Sohalt:
I guess it would be nice if authentication by email was supported separately (with appropriate labels in the UI etc)

We would have to be careful however with the `Remote-User` header, because the `username_attribute` would have to be unique for all users (and in my scenario the only guaranteed unique attribute is `mail`)

nightah:
> I guess it would be nice if authentication by email was supported separately (with appropriate labels in the UI etc)

This is an easy fix, we could check the `users_filter` to see if it includes `{mail_attribute}={input}` and if so allow the frontend to display `Username/Email` I actually think we mocked this when we initially enabled logging in with email, it wasn't conditional though.

If your username attribute isn't unique across users there will be clashes and issues for storage of security keys, totp keys etc.

nightah:
Could you not use unique placeholders for each of your users?
Even if there is no intent to allow them to login with the username, you could use numbered placeholders for the `uid` field to avoid any clashes and only allow them to login with their email via the `users_filter`.

Sohalt:
> Could you not use unique placeholders for each of your users?

Yes, I could do that, although I don't really see, how that would be an improvement over just using the email as the `username_attribute`. Let's say I set `uid := mail`. Then I have a unique "placeholder" for each user, which just happens to be the email, so I end up in the same situation as just using `username_attribute = mail` (with the exception that I need to do extra work to make sure uid is populated for every user).

Sohalt:
> we're discussing the possibility of adding another field in the configuration which would allow administrators to specify LDAP attributes as the display name.

If it is desired to show the user's name in the UI, e.g. to greet them (not really necessary in my opinion), that would probably be the cleanest option. `interOrgPerson` for example defines the `displayName` attribute exactly for this purpose (https://tools.ietf.org/html/rfc2798#section-2.3)

Sohalt:
> This is an easy fix, we could check the users_filter to see if it includes `{mail_attribute}={input}` and if so allow the frontend to display `Username/Email`

I like that solution.

nightah:
Even RFC2798 intends for the `uid` attribute to carry the username, if you were to try and use this in an actual environment it would not work with how you have it configured.

I don't particularly want to support every LDAP schema that exists in the wild, we won't be able to determine all instances where the schema could have implications to our existing code or code for future releases.

Also the thought here is that we would be adapting Authelia just to conform to your specific use case, which isn't something we would want to do without proper consideration.

This is why I would prefer for you to adapt your schema to the standard.
Having said that, @clems4ever @james-d-elliott this is probably worth your thoughts/input too.

Sohalt:
> Even RFC2798 intends for the uid attribute to carry the username, if you were to try and use this in an actual environment it would not work with how you have it configured.

I am using this in an "actual environment" (with several services and ~30 users), which is why I want it to be supported in the first place. If I were not already using this, I would have no problem adapting my schema.

> Also the thought here is that we would be adapting Authelia just to conform to your specific use case, which isn't something we would want to do without proper consideration.

I very much get that, but if you look at the patch, it's a 3 line change, which keeps all existing configurations working exactly the same and removes an (imho) arbitrary restriction, which actually came as a big surprise to me and took me quite some time to debug.

If you want to keep the current behavior, fine (although I really can't see why you would), but then at least check for the case where `username_attribute == email_attribute` and make that a configuration error.
With the current situation it just silently sets the username to the empty string, which breaks authentication.

nightah:
> I am using this in an "actual environment" (with several services and ~30 users), which is why I want it to be supported in the first place. If I were not already using this, I would have no problem adapting my schema.

Adapting your schema should have no implication to your existing setup, it's should just be time and effort, unless there's something I'm missing?
While you may not be strictly conforming to RFC2798, it also doesn't mean you are not conforming to it.

I wouldn't take what I've said as we will not accept this PR or your use case, however, we need to think and consider the implications before we accept it.

There are real-life examples where this might cause issues, but might not for your specific instance because of how you run your environment.
For example:
* users may have multiple email addresses (which should be considered correct?)
* what happens when/if an email changes

Sohalt:
> • users may have multiple email addresses (which should be considered correct?)
• what happens when/if an email changes

This is only an issue when you configure `username_attribute == email_attribute`. If you do that, you basically say: there can only be one email and it cannot change, because it identifies the user. This is the current situation with my schema and it's why I said it's not ideal. But it's valid and it works and it should not be of concern to authelia.

clems4ever:
I don't see why using the same attribute for mail and username would not be valid as long as the username attribute still uniquely identifies the user. One first protection for uniqueness is that the attribute must be a single value and not a list: https://github.com/authelia/authelia/blob/730e88df9dd7cf1dcdc280c0c21f512c87377893/internal/authentication/ldap_user_provider.go#L166.
However we cannot completely guarantee uniqueness technically anyway because even if the RFC states that we should use uid as unique identifier, one could just use the same uid for two entries uniquely identified by cn instead (it's valid for OpenLDAP at least since cn was the previous unique identifier used in suites). So for me, it's the responsibility of the admin to guarantee that uniqueness.
However, the admin does need to be aware of what he is doing and we for sure won't bend Authelia to adapt to weird non-RFC compliant cases in the future, that I fully agree with you @nightah . So if I were to do it, I would probably duplicate the email in uid and take that as the unique key to conform to the RFC.

We should probably be really clear about what `uniquely identifies` means in the doc and the config template, i.e., single value attribute and value cannot be changed later because it would break the relationship between the user and his/her devices and many other things stored in DB. We could also emphasize the advice of using the proper attribute. Moreover, it must be really clear from the doc that the design choice of using email as the primary key for users also means no multi-email setup ever with the current design because otherwise the attribute would not be uniquely identifying users anymore (multi value field not allowed).

I don't see any other expected impact since any attribute could be used for username_attribute as long as it uniquely identifies the user (which should be guaranteed by the admin). We could technically do a pre-flight check against the LDAP to help guarantee the uniqueness of entries at startup though but that's all we can do to help...

> This is an easy fix, we could check the users_filter to see if it includes {mail_attribute}={input} and if so allow the frontend to display Username/Email

That solution seems acceptable to me.

Sohalt:
> So for me, it's the responsibility of the admin to guarantee that uniqueness.

Precisely

> We should probably be really clear about what uniquely identifies means in the doc and the config template, i.e., single value attribute and value cannot be changed later because it would break the relationship between the user and his/her devices and many other things stored in DB. We could also emphasize the advice of using the proper attribute. Moreover, it must be really clear from the doc that the design choice of using email as the primary key for users also means no multi-email setup ever with the current design because otherwise the attribute would not be uniquely identifying users anymore (multi value field not allowed).

Sounds very good to me. I know, that with my schema I cannot ever implement a multi-email setup and the email must not change. That's acceptable in my specific scenario however.

clems4ever:
Also, I have not mentioned the display name in my last comment but I'm for introducing `display_name_attribute` to the ldap config in a later PR because it makes sense to have it displayed in the portal.

Issue #858, opened 2020-04-11
[Guide] How to get Jira to auto-login with authelia SSO


dzervas:
I'm writing this for anyone that wants to auto-login to Jira with authelia :)
My setup is traefik with docker provider and jira & authelia as containers
I'm using a paid (but cheap) Jira plugin, [EasySSO](https://marketplace.atlassian.com/apps/1212581/easy-sso-jira-kerberos-ntlm-saml?hosting=server&tab=overview) (I'm not affiliated in any way, just searched "http header" in marketplace)

You need to do the following:
  - Add `traefik.http.middlewares.authelia.forwardauth.authResponseHeaders=Remote-User, Remote-Groups` in the labels of authelia
  - Add `traefik.http.routers.jira.middlewares=authelia@docker` in the labels of Jira
  - Install EasySSO in Jira
  - Go to EasySSO preferences and add the "Remote-User" header under HTTP and tick the "Username" checkbox.
  - Profit!

dzervas:
Closing as it's not a question

james-d-elliott:
We can probably add this to our community docs. :)

dzervas:
Do you take MRs? 😊
Better there than my blog!

james-d-elliott:
We do. Here's the docs site: https://docs.authelia.com/community/
Here's the source: https://github.com/authelia/authelia/tree/master/docs/community

If you know markdown you can easily contribute to our docs.

PR #868, opened 2020-04-14
Add jira auto-login with http headers documentation

dzervas:
As discussed in #858 !


james-d-elliott:
@dzervas if you make the file like so it will work:

docs/community/using-remote-user-header-for-sso-with-jira.md:
```yaml
---
layout: default
title: Using Remote-User header for SSO with Jira
parent: Community
nav_order: 2
---

# Using Remote-User header for SSO with Jira

You can make Jira auto-login to the user that is currently logged in to authelia.
I say "auto-login" as I couldn't find any plugin to actually be authentication
provider through HTTP headers only - LDAP though seems to have support.

So this guide is targeted to authelia users that don't use any other authentication
backend.

I'm using traefik with docker as an example, but any proxy that can forward
authelia `Remote-User` header is fine.

First of all, users should exist on both authelia and Jira AND have the same
username for this to work. Also you will have to [pay for a plugin](https://marketplace.atlassian.com/apps/1212581/easy-sso-jira-kerberos-ntlm-saml?hosting=server&tab=overview).

After both steps are done:
  - Add `traefik.http.middlewares.authelia.forwardauth.authResponseHeaders=Remote-User` in the labels of authelia
  - Add `traefik.http.routers.jira.middlewares=authelia@docker` in the labels of Jira (to actually enable authelia for the jira instance)
  - Install EasySSO in Jira
  - Go to EasySSO preferences and add the "Remote-User" header under HTTP and tick the "Username" checkbox.
  - Save

## Other Systems

While this guide is tailored for Jira, you can use a similar method with many other services like Jenkins and Grafana.
```

dzervas:
The link to this page is generated automatically on `index.md`, right? I don't need to add it manually

james-d-elliott:
Correct. It's automatic on the docs website.

PR #869, opened 2020-04-15
[FEATURE] Support multiple domains and multiple subjects in ACLs

dustins:
I made an attempt at adding support for multiple Subjects and Domains in a single rule. All the tests seem to pass, and I added a few more to test the functionality I added.


nightah:
Thanks for your contribution @dustins, could you please update the reference documentation and the configuration template for the associated changes over at:
- https://github.com/authelia/authelia/blob/master/docs/configuration/access-control.md
- https://github.com/authelia/authelia/blob/master/config.template.yml

config.template.yml
```diff
@@ -205,7 +205,8 @@ access_control:
       # Network based rule, if not provided any network matches.
       networks:
         - 192.168.1.0/24
-    - domain: secure.example.com
+
+    - domain: ["secure.example.com", "private.example.com"]
```
nightah, line 209:
```suggestion
    - domain:
      - secure.example.com
      - private.example.com
```

config.template.yml
```diff
@@ -215,8 +216,9 @@ access_control:
     - domain: "mx2.mail.example.com"
       subject: "group:admins"
       policy: deny
+
     - domain: "*.example.com"
-      subject: "group:admins"
+      subject: ["group:admins", "group:moderators"]
```
nightah, line 221:
```suggestion
      subject:
        - "group:admins"
        - "group:moderators"
```

docs/configuration/access-control.md
```diff
@@ -104,7 +101,7 @@ access_control:
       networks:
       - 192.168.1.0/24
 
-    - domain: secure.example.com
+    - domain: ["secure.example.com", "private.example.com"]
```
nightah, line 104:
```suggestion
    - domain:
      - secure.example.com
      - private.example.com
```

docs/configuration/access-control.md
```diff
@@ -115,7 +112,7 @@ access_control:
       policy: deny
 
     - domain: "*.example.com"
-      subject: "group:admins"
+      subject: ["group:admins", "group:moderators"]
```
nightah, line 115:
```suggestion
      subject:
        - "group:admins"
        - "group:moderators"
```

internal/configuration/schema/access_control.go
```diff
@@ -8,9 +8,9 @@ import (
 
 // ACLRule represent one ACL rule
 type ACLRule struct {
-	Domain    string   `mapstructure:"domain"`
+	Domains   []string `mapstructure:"domain,weak"`
```
clems4ever, line 11: Please add a comment to justify why we use `weak` here. From my understanding it's for being backward-compatible, isn't it?

internal/configuration/schema/access_control.go
```diff
@@ -33,16 +33,19 @@ func IsNetworkValid(network string) bool {
 
 // Validate validate an ACL Rule
 func (r *ACLRule) Validate(validator *StructValidator) {
-	if r.Domain == "" {
+
+	if len(r.Domains) == 0 {
 		validator.Push(fmt.Errorf("Domain must be provided"))
```
clems4ever, line 38:
```suggestion
validator.Push(fmt.Errorf("At least one domain must be provided"))
```

internal/authorization/authorizer.go
```diff
@@ -47,8 +47,16 @@ func selectMatchingSubjectRules(rules []schema.ACLRule, subject Subject) []schem
...
-		if isSubjectMatching(subject, rule.Subject) && isIPMatching(subject.IP, rule.Networks) {
-			selectedRules = append(selectedRules, rule)
+		if len(rule.Subjects) > 0 {
+			for _, subjectRule := range rule.Subjects {
+				if isSubjectMatching(subject, subjectRule) && isIPMatching(subject.IP, rule.Networks) {
+					selectedRules = append(selectedRules, rule)
+				}
+			}
+		} else {
+			if isSubjectMatching(subject, "") && isIPMatching(subject.IP, rule.Networks) {
```
clems4ever, line 57: Why using `isSubjectMatching(subject, "") && ` in the condition? From what I can see, this will always return true according to the implementation: https://github.com/authelia/authelia/blob/4d1c2f38bef9708463b3f7ede4f12043de8df939/internal/authorization/subject_matcher.go#L11.
So IMO, we can just strip it.

clems4ever:
Thank you @dustins for this contribution. I also added a few comments.

internal/configuration/schema/access_control.go
dustins: You're correct, having `weak` here allows someone to still define a single value and coerces it into a string slice. I thought this would be best to make the change as minimally invasive to what the docs currently show and how people have been using Authelia to date.

From what I can tell from [mapstructure's documentation](https://godoc.org/github.com/mitchellh/mapstructure#DecoderConfig.WeaklyTypedInput) decoding should otherwise function similarly since the destination type is a string.

internal/configuration/schema/access_control.go
dustins: I had thought about making this change with my original PR, although it felt weird to change this warning to talking about multiple when most users will likely have just one and when the docs primarily show a single value here. I feel like there is a slightly lowered chance of confusing someone by just telling someone the domain needed to be defined instead of how many, but I get it is probably 6 eggs, one half dozen type of thing.

But as the primary dev, just reply you'd like it this way and I'll make the change =)

internal/authorization/authorizer.go
dustins: You're right. I originally had left this because I wanted to preserve how it was being processed before without looking at the implementation details. But it always returns true in this scenario so I just stripped it out.

internal/configuration/schema/access_control.go
clems4ever: Excellent, so just add a comment so that people will keep that in sight while reading the code.

internal/configuration/schema/access_control.go
clems4ever: That makes sense. No big deal. I'm fine with the current version.

PR #901, opened 2020-04-22
[FEATURE] Config Validation

### LINKED ISSUE #776: configuration.yml validation
**User-Story**
As an admin of an authelia installation using CI/CD mechanics, I would like to validate configuration files to be valid before they are deployed.

**Problem**
In v4.7.0 `logs_level` got renamed to `log_level` - yamllint for example does not know about config key names, so this change would have slipped through the safety net when dependabot or renovate the docker image version in a docker-compose.yml.

**Solution**
During the CI steps of linting various file types, I would like to add another step to verify the config file to be valid on a more thorough level, to avoid breaking changes to the config file slipping through by accident.

A subcommand / tool that knows about available configuration parameters can be used to validate if the config is still valid with the new version.

**Examples**
* prometheus promtool: https://github.com/prometheus/prometheus/tree/master/cmd/promtool
* CloudFormation Linter: https://github.com/aws-cloudformation/cfn-python-lint

james-d-elliott:
Didn't spend much time on this, and we didn't really agree on a design in advance so see what you think.

If a key doesn't exist in the list in const.go it will fail unless it's in the map. The map is used to configure custom errors in the situations where we expect people may get it wrong and to give specific advice.

* check configuration for invalid keys on startup
* allow users to manually trigger all configuration validation on a file using a cmd
* setup all defaults in config template and run tests against it to prevent accidents
* use tests to check bad configuration values are caught
* use tests to check old configuration values are caught

Closes #776

### DIFF
--- a/internal/commands/validate.go
+++ b/internal/commands/validate.go
@@ -1,0 +1,41 @@
+package commands
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/authelia/authelia/internal/configuration"
+)
+
+// ValidateConfigCmd uses the internal configuration reader to validate the configuration.
+var ValidateConfigCmd = &cobra.Command{
+	Use:   "validate-config [yaml]",
+	Short: "Check a configuration against the internal configuration validation mechanisms.",
+	Run: func(cobraCmd *cobra.Command, args []string) {
+		configPath := args[0]
+		if _, err := os.Stat(configPath); err != nil {
+			fmt.Printf("Error Loading Configuration: %s\n", err)
+			os.Exit(2)
+		}
+
+		// TODO: Actually use the configuration to validate some providers like Notifier
+		_, errs := configuration.Read(configPath)
+		if len(errs) != 0 {
+			str := "Errors"
+			if len(errs) == 1 {
+				str = "Error"
+			}
+			fmt.Printf("%s occurred parsing configuration:\n", str)
+			for _, err := range errs {
+				fmt.Printf("\t%s\n", err)
+			}
+			os.Exit(1)
+		} else {
+			fmt.Printf("Configuration parsed successfully without errors.\n")
+			os.Exit(0)
+		}
+	},
+	Args: cobra.MinimumNArgs(1),
+}
--- a/internal/commands/validate.go
+++ b/internal/commands/validate.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"log"
 	"os"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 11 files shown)


internal/commands/validate.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// ValidateConfigCmd uses the internal configuration reader to validate the configuration.
+var ValidateConfigCmd = &cobra.Command{
+	Use:   "validate-config [yaml]",
+	Short: "Check a configuration against the internal configuration validation mechanisms.",
+	Run: func(cobraCmd *cobra.Command, args []string) {
+		configPath := args[0]
+		if _, err := os.Stat(configPath); err != nil {
+			fmt.Printf("Error Loading Configuration: %s\n", err)
```
clems4ever, line 19: Use the logger module instead.

internal/commands/validate.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		// TODO: Actually use the configuration to validate some providers like Notifier
+		_, errs := configuration.Read(configPath)
+		if len(errs) != 0 {
+			str := "Errors"
+			if len(errs) == 1 {
+				str = "Error"
+			}
+			fmt.Printf("%s occurred parsing configuration:\n", str)
```
clems4ever, line 30: Use the logger module instead.

internal/commands/validate.go
```diff
@@ -0,0 +1,41 @@
...
+			for _, err := range errs {
+				fmt.Printf("\t%s\n", err)
```
clems4ever, line 32: Use the logger module instead.

internal/commands/validate.go
```diff
@@ -0,0 +1,41 @@
...
+			}
+			os.Exit(1)
+		} else {
+			fmt.Printf("Configuration parsed successfully without errors.\n")
```
clems4ever, line 36: Use the logger module instead.

internal/configuration/validator/keys_test.go
```diff
@@ -0,0 +1,32 @@
+package validator
```
clems4ever, line 1: Add a test covering specificErrorKeys.

james-d-elliott: Done! And it highlighted a problem. :)

internal/configuration/validator/keys_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	errs := val.Errors()
+	require.Len(t, errs, 2)
+
+	assert.EqualError(t, errs[0], "config key not expected: bad_key")
+	assert.EqualError(t, errs[1], "config key not expected: totp.skewy")
+}
+
+func TestAllSpecificErrorKeys(t *testing.T) {
+	var configKeys []string
```
autheliabot, line 36: **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
Consider preallocating `configKeys` (prealloc)

PR #902, opened 2020-04-22
[FIX] File Notifier Default Permissions

james-d-elliott:
* set to 0600 for security

As discussed with Clement.

### DIFF
--- a/internal/notification/const.go
+++ b/internal/notification/const.go
@@ -1,3 +1,3 @@
 package notification
 
-const fileNotifierMode = 0755
+const fileNotifierMode = 0600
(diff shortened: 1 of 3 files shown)


nightah:
That will be an issue if anybody mounts their configuration in a container as `:ro`. I think some of our docs or examples have that too.

Might be worthwhile double checking just to make sure that’s not the case.

james-d-elliott:
> checking

It only applies to the file notifier, so the emails saved as text. If a user mounts this as RO it will fail. My startup checks will fail it for example, or when a user triggers an email sending action.

nightah:
Ah I completely misread this, thought we were talking about the configuration itself.

All good :).

PR #906, opened 2020-04-23
[DOCS] Add SECURITY.md and update README.md.

clems4ever:


SECURITY.md
```diff
@@ -0,0 +1,10 @@
+Security is taken very seriously here, therefore we follow the rule of responsible
+disclosure and we encourage you to do so.
```
james-d-elliott, line 2:
```suggestion
Authelia takes security very seriously. We follow the rule of [responsible disclosure](https://en.wikipedia.org/wiki/Responsible_disclosure), and we encourage the community to as well.
```

Thoughts?

clems4ever: yeah, I did not properly rebase I guess.

clems4ever: I aligned all documents in the next commit. Thanks.

PR #916, opened 2020-04-25
[FEATURE] Embed static assets in Go binary

### LINKED ISSUE #494: Embed web application in Authelia binary
This could be done according to: https://www.thepolyglotdeveloper.com/2017/03/bundle-html-css-javascript-served-golang-application/

nightah:
This change will allow us to ship Authelia as a standalone binary for deployment without an associated `public_html` directory accompanying it.

~Authelia can still be utilised with local assets on the filesystem by setting the configuration option `asset`. With this set Authelia will, attempt to serve assets from the location specified in the option, if unset or empty it will default to serving via the embedded virtual filesystem.~

There's still a couple more housekeeping changes that need to be completed to be able to merge this, including:

- [X] Adjust AUR packages to remove `public_html`.
- [X] Include the static assets on Buildkite and GitHub artifacts

*Note: This is fine to go to master, I need to be present when tagged as a release though because there's some other changes that need to happen immediately after for the AUR packages*

Closes #494.


internal/server/index.go
```diff
@@ -14,9 +15,40 @@ import (
...
 
-// ServeIndex serve the index.html file with nonce generated for supporting
-// restrictive CSP while using material-ui.
-func ServeIndex(publicDir string) fasthttp.RequestHandler {
+// ServeEmbeddedIndex serve the index.html file with nonce generated for supporting
+// restrictive CSP while using material-ui from the embedded virtual filesystem.
+func ServeEmbeddedIndex(publicDir string) fasthttp.RequestHandler {
+	f, err := br.Open(publicDir + "/index.html")
+	if err != nil {
+		logging.Logger().Fatalf("Unable to open index.html: %v", err)
```
clems4ever, line 23: we should write an error response and return from the function here. Can you also fix it in the other handler (I forgot to do it when I introduced it).

internal/server/index.go
```diff
@@ -14,9 +15,40 @@ import (
...
+	}
+
+	b, err := ioutil.ReadAll(f)
+	if err != nil {
+		logging.Logger().Fatalf("Unable to read index.html: %v", err)
```
clems4ever, line 28: Write an error response and return as suggested above. Same thing in the other handler.

internal/server/index.go
```diff
@@ -14,9 +15,40 @@ import (
...
+	}
+
+	tmpl, err := template.New("index").Parse(string(b))
+	if err != nil {
+		logging.Logger().Fatalf("Unable to parse index.html template: %v", err)
```
clems4ever, line 33: Write an error response and return as suggested above. Same thing in the other handler.

internal/server/index.go
```diff
@@ -14,9 +15,40 @@ import (
...
+	}
+
+	return func(ctx *fasthttp.RequestCtx) {
+		nonce := utils.RandomString(32, alphaNumericRunes)
```
clems4ever, line 37: This piece of code is duplicated, can you factorize it?

internal/server/server.go
```diff
@@ -21,16 +22,29 @@ import (
 // StartServer start Authelia server with the given configuration and providers.
 func StartServer(configuration schema.Configuration, providers middlewares.Providers) {
 	autheliaMiddleware := middlewares.AutheliaMiddleware(configuration, providers)
-	publicDir := os.Getenv("PUBLIC_DIR")
-	if publicDir == "" {
-		publicDir = "./public_html"
+	publicDir := "/public_html"
+
+	if os.Getenv("LOCAL_ASSETS") == "true" {
```
clems4ever, line 27: What do you think about removing PUBLIC_DIR and only keeping LOCAL_ASSETS env var containing the path? If it's not provided or empty we take the embedded files. It's just to avoid stacking env vars.
For old setups having the PUBLIC_DIR env var after the bump, we can easily and safely display a warning saying that setting this env var is not doing anything anymore.

We can go even further by moving this config from env var into the config file. It's my fault if env vars were introduced at the beginning and there was no good reason to do that, my bad.

ishkanan: Agree with this. Perhaps name it to "ASSETS_DIR" and just document the behaviour in the README.

Also not sure if everyone agrees with this, but I generally only use environment variables for sensitive info (passwords, etc.) because they usually don't show in process explorers. If everything else can go in a config file/CLI/some other way, I try to do it.

Sorry to have a complete stranger butt in - /2c

nightah: > For old setups having the PUBLIC_DIR env var after the bump, we can easily and safely display a warning saying that setting this env var is not doing anything anymore.
> 
> We can go even further by moving this config from env var into the config file. It's my fault if env vars were introduced at the beginning and there was no good reason to do that, my bad.

I've taken the approach of adding an `Assets` key in the root configuration alongside `Host` and `Port`.

If not provided/empty it will utilise the embedded assets, otherwise it will utilise the path provided to attempt to load assets.

internal/server/index.go
```diff
@@ -15,21 +17,39 @@ import (
...
+
+	if publicDir == "/public_html" {
+		f, err = br.Open(publicDir + "/index.html")
+	} else {
+		f, err = os.Open(publicDir + "/index.html")
+	}
+
 	if err != nil {
 		logging.Logger().Fatalf("Unable to open index.html: %v", err)
+		return func(ctx *fasthttp.RequestCtx) {
```
clems4ever, line 34: Sorry, now that I see it written it looks weird to me. I think it would be better to make the `ServeIndex` function return (fasthttp.RequestHandler, error) and panic in the caller in case of error. Indeed, Authelia should not even start in those cases since this code is only executed at startup.

internal/server/server.go
```diff
@@ -21,16 +22,24 @@ import (
 // StartServer start Authelia server with the given configuration and providers.
 func StartServer(configuration schema.Configuration, providers middlewares.Providers) {
 	autheliaMiddleware := middlewares.AutheliaMiddleware(configuration, providers)
-	publicDir := os.Getenv("PUBLIC_DIR")
-	if publicDir == "" {
-		publicDir = "./public_html"
+
+	// TODO: Remove in v4.18.0
+	if os.Getenv("PUBLIC_DIR") != "" {
+		logging.Logger().Warn("PUBLIC_DIR environment variable has been deprecated, assets are now embedded," +
```
clems4ever, line 28: Nice!

internal/server/index.go
nightah: I've just realised neither of these should be required, `logging.Logger().Fatalf` will obviously exit gracefully once the error has been logged.

Any issues with the embedded filesystem are handled and you handle executing the template by returning a 503, perhaps instead we actually change this to a 500?

clems4ever: Ah yes you're right, Fatalf does the job. For 500 vs 503, just align with the other handlers.

internal/server/index.go
```diff
@@ -1,10 +1,10 @@
+//go:generate broccoli -src ../../public_html -o public_html
```
autheliabot, line 1: **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
package comment should be of the form "Package server ..." (golint)

Issue #949, opened 2020-05-01
User enumeration of file-based backend based on timing attack


TheHllm:
It is possible to tell from the outside if a username is valid, by simply performing a login attempt.
If the server responds quickly to the request the username is invalid, if however the server dose not respond quickly the password is at fault.

To fix this issue the server should take a fixed minimum amount of time to respond to login requests.
This would have to be implemented in a way, such that the server only takes (significantly longer) for a valid login.

In pseudo-code:
```
int fixedTime = 10; // should be configurable so that this time is grater than the time it takes to hash a password
 int login(u,p){
 if(username_valid(u)){
  string hash = getHash(u);
  if(checkHash(hash, pw)){
   //do some other login stuff that may take longer that a wait would have.....
   return 200;
  }else{
   waitForRemaingTime(Date.now, time);
   return 403;
  }
 }else{
   waitForRemaingTime(Date.now, time);
   return 403;
 }
}
```

james-d-elliott:
Hey @TheHllm thanks for reporting this. Is my understanding correct that you're using the file based authentication backend?

TheHllm:
@james-d-elliott  Yes

james-d-elliott:
Alright. So the reason this is happening most likely, if the user doesn't exist we don't has the input password to compare it. I will discuss it with the other maintainers and see what they think. While it's good for performance it slightly reduces security.

To put your mind at ease, [regulation](https://docs.authelia.com/configuration/regulation.html) can be used to block brute force attacks.

TheHllm:
The relevant code is (as far as i can tell) in [handler_firstfactor.go](https://github.com/authelia/authelia/blob/master/internal/handlers/handler_firstfactor.go) but i have never done any go....
It should only impact performance for invalid logins.
And i agree with you that brute force is unrealisic if you use regulation

james-d-elliott:
Related due to inability to solve this: #478

clems4ever:
> Alright. So the reason this is happening most likely, if the user doesn't exist we don't has the input password to compare it. I will discuss it with the other maintainers and see what they think. While it's good for performance it slightly reduces security.
> 
> While [regulation](https://docs.authelia.com/configuration/regulation.html) can be used to block brute force attacks, if they're looking for usernames.. this would get them.

[...]

PR #959, opened 2020-05-02
[SECURITY] Fix Authentication HTTP Status Codes

### LINKED ISSUE #478: Make first factor failures non-success status codes for easy log filtering
Hey @clems4ever,
I have noticed another inconvenience. :sweat_smile: 

Failures in authelia's first factor authentication currently return a 200 status code.
Which is imo not good practice, as the login call was really not 'successful', and very inconvenient for logging authentication failures.

I have traefik configured to log client errors (status codes 400-499), which works nicely for Basic Auth, but fails to detect any failures using authelia's web interface.

I am feeding my logs into fail2ban. I know, that authelia's has its own banning feature, but I am using fail2ban for multiple services on my system (including other traefik services) and would like to reuse it for authelia as well.

From what I have seen browsing the source code, you are using axios on the frontend side, which [should still work fine with 400 codes](https://github.com/clems4ever/authelia/blob/master/web/src/services/Client.ts#L7) and take the same branch.

On the backend side, it seems that [Error](https://github.com/clems4ever/authelia/blob/master/internal/middlewares/authelia_context.go#L49) and [ReplyError](https://github.com/clems4ever/authelia/blob/master/internal/middlewares/authelia_context.go#L62) would need to be adjusted to return different status codes.

Do you think this approach would work, or might that break other parts of authelia? I am not to familiar with go, but I guess I could make a quick PR, if that is all this takes? (And you are agreeing with this request of course.)

(Note: The same might apply to the second factor, but I have not tested that *yet*)

james-d-elliott:
Will close #478 when merged. Adds additional mitigation options against #949

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401


internal/handlers/handler_sign_duo.go
```diff
@@ -60,15 +60,16 @@ func SecondFactorDuoPost(duoAPI duo.API) middlewares.RequestHandler {
 		err = ctx.Providers.SessionProvider.RegenerateSession(ctx.RequestCtx)
 
 		if err != nil {
-			ctx.Error(fmt.Errorf("Unable to regenerate session for user %s: %s", userSession.Username, err), authenticationFailedMessage)
+			// TODO: Shouldn't this be mfaValidationFailedMessage like the rest?
```
clems4ever, line 63: yes you're right.

internal/handlers/handler_sign_totp.go
```diff
@@ -14,40 +14,41 @@ func SecondFactorTOTPPost(totpVerifier TOTPVerifier) middlewares.RequestHandler
...
-			ctx.Error(fmt.Errorf("Wrong passcode during TOTP validation for user %s", userSession.Username), mfaValidationFailedMessage)
+			handleErrorResponse(ctx, fmt.Errorf("Wrong passcode during TOTP validation for user %s", userSession.Username), mfaValidationFailedMessage)
 			return
 		}
 
 		err = ctx.Providers.SessionProvider.RegenerateSession(ctx.RequestCtx)
 
 		if err != nil {
-			ctx.Error(fmt.Errorf("Unable to regenerate session for user %s: %s", userSession.Username, err), authenticationFailedMessage)
+			// TODO: Shouldn't this be mfaValidationFailedMessage like the rest?
```
clems4ever, line 42: yes, you're right.

internal/handlers/handler_sign_u2f_step1.go
```diff
@@ -13,12 +13,14 @@ import (
 
 // SecondFactorU2FSignGet handler for initiating a signing request.
 func SecondFactorU2FSignGet(ctx *middlewares.AutheliaCtx) {
+	// TODO: Should this be operationFailedMessage?
```
clems4ever, line 16: I would keep it because we don't want the public to know whether X-Forward-... is set or not in the backend...

clems4ever: I mean changing it would allow information disclosure for an attacker.

internal/handlers/handler_sign_u2f_step1.go
```diff
@@ -13,12 +13,14 @@ import (
...
 	if ctx.XForwardedProto() == nil {
 		ctx.Error(errMissingXForwardedProto, operationFailedMessage)
 		return
 	}
 
 	if ctx.XForwardedHost() == nil {
+		// TODO: Should this be operationFailedMessage?
```
clems4ever, line 23: Same response as above.

internal/handlers/handler_sign_u2f_step2.go
```diff
@@ -43,15 +43,16 @@ func SecondFactorU2FSignPost(u2fVerifier U2FVerifier) middlewares.RequestHandler
 		err = ctx.Providers.SessionProvider.RegenerateSession(ctx.RequestCtx)
 
 		if err != nil {
-			ctx.Error(fmt.Errorf("Unable to regenerate session for user %s: %s", userSession.Username, err), authenticationFailedMessage)
+			// TODO: Shouldn't this be mfaValidationFailedMessage like the rest?
```
clems4ever, line 46: you're right.

internal/handlers/handler_firstfactor.go
```diff
@@ -36,15 +38,14 @@ func FirstFactorPost(ctx *middlewares.AutheliaCtx) {
...
-		ctx.Error(fmt.Errorf("Error while checking password for user %s: %s", bodyJSON.Username, err.Error()), authenticationFailedMessage)
+		handleErrorResponse(ctx, fmt.Errorf("Error while checking password for user %s: %s", bodyJSON.Username, err.Error()), authenticationFailedMessage)
 		return
 	}
 
 	if !userPasswordOk {
 		ctx.Logger.Debugf("Mark authentication attempt made by user %s", bodyJSON.Username)
 		ctx.Providers.Regulator.Mark(bodyJSON.Username, false) //nolint:errcheck // TODO: Legacy code, consider refactoring time permitting.
-
+		ctx.SetStatusCode(fasthttp.StatusUnauthorized)
```
clems4ever, line 48: you should also use `handleErrorResponse` here otherwise there is an easy way to know that branch has been reached by a hacker. I did it wrong at the beginning, we don't want to differentiate the message.

internal/handlers/handler_sign_u2f_step1.go
james-d-elliott: @clems4ever I mean, in master we have it set to operationFailedMessage instead of mfaValidationFailedMessage.

PR #961, opened 2020-05-02
[CI] Add goconst linter

nightah:


internal/configuration/schema/access_control.go
```diff
@@ -17,7 +17,7 @@ type ACLRule struct {
 
 // IsPolicyValid check if policy is valid.
 func IsPolicyValid(policy string) bool {
-	return policy == "deny" || policy == "one_factor" || policy == "two_factor" || policy == "bypass"
+	return policy == defaultPolicy || policy == "one_factor" || policy == "two_factor" || policy == "bypass"
```
clems4ever, line 20: would not denyPolicy be more accurate?

clems4ever: Don't you also cover the other parts of the condition?

nightah: Per our discussion, agreed on `denyPolicy` and goconst only looks for assignments that are made for the same thing more than once, so the other conditions aren't necessary yet.

PR #971, opened 2020-05-04
[SECURITY] Enable Delay for First Factor Post

james-d-elliott:
Will close #949 if merged.

1. On startup:
  1. If file provider: hash a randomly generated password of 20 characters with the hash password settings, we time this, plus add a random interval of time between 150-200ms. If this delay isn't above 450ms we randomly assign it a value from the next step. This is handled in FirstFactorPost() which now returns a func with the values calculated prior to that.
  2. Otherwise: set the startup delay to 450-500ms.
2. When the post comes in we record the time we received it. After the authentication passes or fails, we check the authentication attempt will not be returned early and if it would be, we sleep until it should be returned.

TODO:
- [ ] figure out why the tests failed when not disabling the delay and fix it, it's likely this is causing other issues
- [ ] design some tests for this
- [ ] documentation

PR #977, opened 2020-05-05
[CI] Add gocritic linter

nightah:
The outstanding recommendations are due to be addressed in #959 and #971 respectively.
They should be merged first.

### DIFF
--- a/cmd/authelia/main.go
+++ b/cmd/authelia/main.go
@@ -73,6 +73,7 @@ func startServer() {
 	}
 
 	var storageProvider storage.Provider
+
 	switch {
 	case config.Storage.PostgreSQL != nil:
 		storageProvider = storage.NewPostgreSQLProvider(*config.Storage.PostgreSQL)
@@ -85,6 +86,7 @@ func startServer() {
 	}
 
 	var notifier notification.Notifier
+
 	switch {
 	case config.Notifier.SMTP != nil:
 		notifier = notification.NewSMTPNotifier(*config.Notifier.SMTP)
--- a/cmd/authelia/main.go
+++ b/cmd/authelia/main.go
@@ -37,6 +37,7 @@ func startServer() {
 		for _, err := range errs {
 			logging.Logger().Error(err)
 		}
+
 		panic(errors.New("Some errors have been reported"))
 	}
 
--- a/cmd/authelia/main.go
+++ b/cmd/authelia/main.go
@@ -62,33 +62,37 @@ func startServer() {
 
 	var userProvider authentication.UserProvider
 
-	if config.AuthenticationBackend.File != nil {
+	switch {
+	case config.AuthenticationBackend.File != nil:
 		userProvider = authentication.NewFileUserProvider(config.AuthenticationBackend.File)
-	} else if config.AuthenticationBackend.Ldap != nil {
+	case config.AuthenticationBackend.Ldap != nil:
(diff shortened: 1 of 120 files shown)


internal/notification/smtp_notifier.go
```diff
@@ -98,17 +98,18 @@ func (n *SMTPNotifier) startTLS() error {
 		return nil
 	}
 
-	ok, _ := n.client.Extension("STARTTLS")
-	if ok {
-		log.Debugf("Notifier SMTP server supports STARTTLS (disableVerifyCert: %t, ServerName: %s), attempting", n.tlsConfig.InsecureSkipVerify, n.tlsConfig.ServerName)
+	if n.disableRequireTLS {
+		log.Warn("Notifier SMTP server does not support STARTTLS and SMTP configuration is set to disable the TLS requirement (only useful for unauthenticated emails over plain text)")
+	}
```
james-d-elliott, line 103: **DO NOT APPLY THIS SUGGESTION.** This should be under the case false if we want to preserve current functionality I think, additionally this will throw errors when we shouldn't we need another case. Something like this:

```suggestion
        case false:
        	switch n.disableRequireTLS {
		case true:
			log.Warn("Notifier SMTP server does not support STARTTLS and SMTP configuration is set to disable the TLS requirement (only useful for unauthenticated emails over plain text)")
		case false:
			return errors.New("Notifier SMTP server does not support TLS and it is required by default (see documentation if you want to disable this highly recommended requirement)")
		}
	}
```

internal/utils/time.go
```diff
@@ -14,7 +14,8 @@ func ParseDurationString(input string) (time.Duration, error) {
 	var duration time.Duration
 
 	matches := parseDurationRegexp.FindStringSubmatch(input)
-	if len(matches) == 3 && matches[2] != "" {
+	switch {
```
autheliabot, line 17: **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
anonymous switch statements should never be cuddled (wsl)

cmd/authelia/main.go
```diff
@@ -63,31 +63,34 @@ func startServer() {
...
+	case config.AuthenticationBackend.Ldap != nil:
 		userProvider = authentication.NewLDAPUserProvider(*config.AuthenticationBackend.Ldap)
-	} else {
+	default:
 		log.Fatalf("Unrecognized authentication backend")
 	}
 
 	var storageProvider storage.Provider
-	if config.Storage.PostgreSQL != nil {
+	switch {
```
autheliabot, line 76: **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
anonymous switch statements should never be cuddled (wsl)

cmd/authelia/main.go
```diff
@@ -63,31 +63,34 @@ func startServer() {
...
+	case config.Storage.Local != nil:
 		storageProvider = storage.NewSQLiteProvider(config.Storage.Local.Path)
-	} else {
+	default:
 		log.Fatalf("Unrecognized storage backend")
 	}
 
 	var notifier notification.Notifier
-	if config.Notifier.SMTP != nil {
+	switch {
```
autheliabot, line 88: **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
anonymous switch statements should never be cuddled (wsl)

PR #993, opened 2020-05-08
[FEATURE] Delay 1FA Authentication

james-d-elliott:
Left to do:

- [x] docs
- [x] tests if I or anyone can figure out what/how to test
- [x] logging cleanup


internal/handlers/handler_firstfactor.go
```diff
@@ -10,119 +12,168 @@ import (
 	"github.com/authelia/authelia/internal/session"
 )
 
+func movingAverageIteration(value time.Duration, successful *bool, movingAverageCursor *int, execDurationMovingAverage *[]time.Duration) float64 {
+	if *successful {
```
clems4ever, line 16: We need to protect this method with a mutex.

internal/handlers/handler_firstfactor.go
```diff
@@ -10,119 +12,168 @@ import (
...
-	err := ctx.ParseBody(&bodyJSON)
+func FirstFactorPost(msInitialDelay time.Duration, delayEnabled bool) middlewares.RequestHandler {
+	var execDurationMovingAverage = make([]time.Duration, movingAverageWindow)
+
+	var movingAverageCursor = 0
 
-	if err != nil {
-		handleAuthenticationUnauthorized(ctx, err, authenticationFailedMessage)
-		return
+	for i := range execDurationMovingAverage {
```
clems4ever, line 49: Can be done only if delayEnabled is true.

PR #999, opened 2020-05-09
[FIX] Hash Password Cmd Not Encoding Provided Salt

### LINKED ISSUE #996: File provider argon2id salt formatted incorrectly
When using a file provider for authentication the salt is not encoded in base64. This is an issue with the authelia code base - the crypt library expects either a base64 encoded salt as input, or an empty salt which it then generates. 

This problem seems to manifest itself in two ways - when generating a password with the built in utility you can see clearly that the salt is not base64 encoded.

If a salt longer than 16 characters is input, the input is simply truncated.

```
docker run -it --rm authelia/authelia:latest authelia hash-password -s thisisatestsalt mypassword
Password hash: $argon2id$v=19$m=1048576,p=8$thisisatestsals$LUhYIsK1LRzeMMJismC3gDSSKkagFCWRjI2OSx0nULU
```

When attempting to decode a valid string generated via: https://argon2.online/ the validity check of the salt is performed without any base64 operations. This results in a valid argon2id hash being discarded due to length comparison issues (the base64 encoded string is longer than the expected 16 character salt), and the following error is displayed in the log (the below string is using a randomised salt):

`msg="Error while checking password for user test: Salt input of UmczWWdNUXZIcklWRTduSQ is invalid (22 characters), it must be 16 or fewer characters"`

This means the only utility capable of generating valid hash strings is the built-in tool, which deviates from the expected standard.

james-d-elliott:
Fixes #996


internal/authentication/password_hash.go
```diff
@@ -123,7 +123,11 @@ func HashPassword(password, salt string, algorithm CryptAlgo, iterations, memory
 	}
 
 	if salt == "" {
-		salt = utils.RandomString(saltLength, HashingPossibleSaltCharacters)
+		if algorithm == HashingAlgorithmArgon2id {
+			salt = crypt.Base64Encoding.EncodeToString([]byte(utils.RandomString(saltLength, HashingPossibleSaltCharacters)))
+		} else {
+			salt = utils.RandomString(saltLength, HashingPossibleSaltCharacters)
```
clems4ever, line 129: Since we've found that sha512 must be b64 valid we can probably simplify this code with 
```
if salt == "" {
      		salt = crypt.Base64Encoding.EncodeToString([]byte(utils.RandomString(saltLength, HashingPossibleSaltCharacters)))
	}
```

PR #1021, opened 2020-05-14
[DEPRECATE] Remove Google Analytics

james-d-elliott:
* it doesn't work with our current CSP
* it's probably not used by anyone
* it isn't in harmony with our security purposes
* literally removes all use of it


internal/configuration/validator/const.go
```diff
@@ -150,6 +149,7 @@ var specificErrorKeys = map[string]string{
 	"authentication_backend.file.hashing.salt_length":          "config key incorrect: authentication_backend.file.hashing should be authentication_backend.file.password",
 	"authentication_backend.file.hashing.memory":               "config key incorrect: authentication_backend.file.hashing should be authentication_backend.file.password",
 	"authentication_backend.file.hashing.parallelism":          "config key incorrect: authentication_backend.file.hashing should be authentication_backend.file.password",
+	"google_analytics": "config key incorrect: google_analytics - this functionality has been removed",
```
nightah, line 152: Indenting should be adjusted here.

docs/configuration/google-analytics.md
```diff
@@ -1,15 +0,0 @@
----
-layout: default
-title: Google Analytics
-parent: Configuration
-nav_order: 4
```
nightah, line 5: With this now being removed we should probably adjust the nav_order for the pages in the Configuration section.

james-d-elliott: It actually just changes precedence I think. It was already missing #1 I think.

internal/handlers/handler_configuration_test.go
```diff
@@ -25,44 +25,35 @@ func (s *ConfigurationSuite) TearDownTest() {
 }
 
 func (s *ConfigurationSuite) TestShouldReturnConfiguredGATrackingID() {
```
clems4ever, line 27: I think this test should be removed completely.

internal/configuration/validator/const.go
```diff
@@ -130,8 +129,9 @@ var validKeys = []string{
 }
 
 var specificErrorKeys = map[string]string{
-	"logs_file_path": "config key replaced: logs_file is now log_file",
-	"logs_level":     "config key replaced: logs_level is now log_level",
+	"logs_file_path":   "config key replaced: logs_file is now log_file",
+	"logs_level":       "config key replaced: logs_level is now log_level",
+	"google_analytics": "config key incorrect: google_analytics - this functionality has been removed",
```
nightah, line 134:
```suggestion
	"google_analytics": "config key removed: google_analytics - this functionality has been deprecated",
```

I feel like suggesting the key is “incorrect” isn’t actually appropriate language.

PR #1027, opened 2020-05-18
[FEATURE] Allow Authelia to listen on a specified path

james-d-elliott:


nightah:
**Note:** This will need some changes to the AUR packages too, so should be merged when the changes are ready to be deployed.

internal/configuration/validator/server_test.go
```diff
@@ -33,3 +33,24 @@ func TestShouldRaiseOnNegativeValues(t *testing.T) {
...
+	require.Len(t, validator.Errors(), 1)
+	assert.Error(t, validator.Errors()[0], "server path must only be alpha numeric characters")
+}
+
+func TestShouldRaiseOnForwardSlashInPath(t *testing.T) {
+	validator := schema.NewStructValidator()
+	config := schema.ServerConfiguration{
+		Path: "app/le",
+	}
+
```
nightah, line 52:
```suggestion
```

internal/configuration/validator/server.go
```diff
@@ -2,15 +2,28 @@ package validator
...
 )
 
 var defaultReadBufferSize = 4096
 var defaultWriteBufferSize = 4096
 
 // ValidateServer checks a server configuration is correct.
 func ValidateServer(configuration *schema.ServerConfiguration, validator *schema.StructValidator) {
+	switch {
+	case strings.Contains(configuration.Path, "/"):
+		validator.Push(fmt.Errorf("server path must not contain any forward slashes"))
```
clems4ever, line 19: why not support this?

internal/server/index.go
```diff
@@ -36,9 +36,9 @@ func ServeIndex(publicDir string) fasthttp.RequestHandler {
 		nonce := utils.RandomString(32, alphaNumericRunes)
 
 		ctx.SetContentType("text/html; charset=utf-8")
-		ctx.Response.Header.Add("Content-Security-Policy", fmt.Sprintf("default-src 'self'; style-src 'self' 'nonce-%s'", nonce))
+		ctx.Response.Header.Add("Content-Security-Policy", fmt.Sprintf("base-uri 'none'; default-src 'self'; object-src 'none'; require-trusted-types-for 'script'; script-src 'self' 'unsafe-inline' 'nonce-%s'; style-src 'self' 'nonce-%s'", nonce, nonce))
```
clems4ever, line 39: `'unsafe-inline'` and `require-trusted-types-for 'script'` are regression to the policy. I suggest to either use the nonce in the script tag or set the base value in an html tag to avoid modifying the CSP.

internal/utils/strings.go
```diff
@@ -3,8 +3,24 @@ package utils
 import (
 	"math/rand"
 	"time"
+	"unicode"
 )
 
+// IsStringAlphaNumeric returns false if any rune in the string is not alpha-numeric.
+func IsStringAlphaNumeric(input string) bool {
+	if len(input) == 0 {
```
clems4ever, line 11: I think you can safely remove that if block.

internal/middlewares/strip_path.go
```diff
@@ -0,0 +1,22 @@
+package middlewares
+
+import (
+	"bytes"
+
+	"github.com/valyala/fasthttp"
+)
+
+// StripPathMiddleware strips the first level of a path.
+func StripPathMiddleware(next fasthttp.RequestHandler) fasthttp.RequestHandler {
```
clems4ever, line 10: Why not let the reverse proxy do this with a path rewrite configuration? That would also avoid sed-ding the `href=` tags in the Dockerfiles, authelia-scripts and in server.go. What do you think?

internal/server/index.go
clems4ever: Apparently the nonce is provided to the script tag so I guess those 2 attributes are not needed anymore.

clems4ever:
you should check branch `feature-support-subpath2` to see how to use the tag attribute method and consequently avoid the unwanted CSP attributes and the type casting. There was also an issue with the dev workflow that I fixed in my branch by setting REACT_APP_BASEPATH env vars in dev and production config files.

internal/configuration/validator/server.go
nightah: I believe it had to do with how the path is cleaned, though James can probably clarify.

It's also to avoid any confusion for users attempting to publish multiple level routes, ie `/auth/authelia`. We only want to support base_url that's a single level deep.

internal/server/index.go
nightah: >Specifying nonce makes a modern browser ignore 'unsafe-inline' which could still be set for older browsers without nonce support.

`'unsafe-inline'` is [ignored if a nonce is provided in CSP2+ supported browser](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src) the only reason it was added was for browser backwards compatibility.

Perhaps my understanding of `require-trusted-types-for 'script'` was incorrect, [here is](https://web.dev/trusted-types/) what I was basing that change off of. I don't seem to understand how that's a regression in the policy.

internal/utils/strings.go
nightah: Adjusted.

internal/middlewares/strip_path.go
nightah: I think supporting base_url should means that there's minimal work required in the reverse proxy.

I cannot think of an application I utilise that supports a configurable `base_url` that also requires a rewrite, I would like to avoid this primarily from a user experience perespective but secondaryily from a support perspective.

We cannot avoid templating the `href=` tags because those can only be specified during the frontend build, I believe templating is the only way to assign this dynamically though happy to look at another implementation if you can think of one.

If we go down the rewrite path with the reverse proxy it means that users will actually need to intercept asset requests from `/static` as opposed to `/<path>/static` and that could conflict with other routes they may have setup in their reverse proxy, which really is messy because they want to share the domain in question hence the base_url modification.

internal/configuration/validator/server.go
clems4ever: Ok, good enough for now and until someone ask for more (it will likely happen in my opinion).

internal/server/index.go
clems4ever: My understanding was not good, I tried to read the w3c documentation and I missed some of the essence. Your document explains it better, thanks for sharing. So you're right, it's not a regression, it's the contrary, we lock down some functions. Do you think it's safe? Won't we break some users?

internal/middlewares/strip_path.go
clems4ever: fair enough, it was just an open question.

config.template.yml
```diff
@@ -16,6 +16,8 @@ server:
   read_buffer_size: 4096
   # Write buffer size configures the http server's maximum outgoing response size in bytes.
   write_buffer_size: 4096
+  # Set the path Authelia listens on, must be alphanumeric chars.
```
clems4ever, line 19: To me it's not clear that the path must be a single level path. Let's improve the comment
```suggestion
  # Set the single level path Authelia listens on, must be alphanumeric chars and slash is forbidden.

File: docs/configuration/server.md [hunk 7]
```diff
@@ -20,10 +20,28 @@ server:
   read_buffer_size: 4096
   # Write buffer size configures the http server's maximum outgoing response size in bytes.
   write_buffer_size: 4096
+  # Set the path Authelia listens on, must be alphanumeric chars.
```
clems4ever, line 23:
```suggestion
  # Set the single level path Authelia listens on, must be alphanumeric chars and slash is forbidden.

File: docs/configuration/server.md [hunk 8]
```diff
@@ -20,10 +20,28 @@ server:
...
+### Path
+
+Authelia by default is served from the root `/` location, either via its own domain or subdomain.
+Example: https://auth.example.com, https://example.com
+```yaml
+server:
+  path: ""
+```
+
+Modifying this setting will allow you to serve Authelia out from a specified base path.
```
clems4ever, line 42:
```suggestion
Modifying this setting will allow you to serve Authelia out from a specified base path. Please note only single level paths are currently supported so the character slash is forbidden.

internal/configuration/validator/server.go
james-d-elliott: I also wanted to keep the input as controlled as possible so we know what we're working with. Some things expect a leading slash, some do not.

PR #1040, opened 2020-05-21
[MISC] Remove executable permission of nginx backend files.

clems4ever:


clems4ever:
Here I rollback the last change which made all html files executable. The initial issue was on my side because of a restrictive umask messing up my with file permissions and consequently impacting the nginx container. I fixed it by changing my umask but we still need to rollback html file to make them non-executable.

nightah:
Yep there needs to be a slight adjutment to the tarred files too for the Kubernetes suite, I'll fix that.

PR #1052, opened 2020-05-26
[CI] Add PathPrefix integration test suite

nightah:


internal/suites/const.go
```diff
@@ -1,10 +1,25 @@
 package suites
 
-import "fmt"
+import (
+	"fmt"
+	"os"
+)
+
+// GetLoginBaseURL returns the URL of the login portal and the path prefix if specified.
+func GetLoginBaseURL() string {
```
nightah, line 9: @clems4ever is this okay to reside here or would you rather I move it to something like `hooks.go`?

internal/suites/example/compose/traefik2/docker-compose.pathprefix.yml
```diff
@@ -0,0 +1,28 @@
+version: '3'
```
clems4ever, line 1: Do you really need to create a new file? Cannot we simply template the original file with ${PathPrefix}. In the default case the env var would be empty.

cmd/authelia-scripts/cmd_suites.go
```diff
@@ -143,6 +143,10 @@ func runSuiteSetupTeardown(command string, suite string) error {
 	cmd.Stderr = os.Stderr
 	cmd.Env = os.Environ()
 
+	if selectedSuite == suitePathPrefix {
```
clems4ever, line 146: OPT: I think this could be improved by generalizing the concept of loading env var in any suite. It could be something like reading env variables from a file stored in the suite resources for instance.

internal/suites/example/compose/traefik2/docker-compose.pathprefix.yml
nightah: Yep you should be right here, good catch!

cmd/authelia-scripts/cmd_suites.go
nightah: Addressed in https://github.com/authelia/authelia/pull/1052/commits/84eb500267d29895b0a6b0daaff3af36f31d22c1.

cmd/authelia-scripts/cmd_suites.go
```diff
@@ -143,6 +144,22 @@ func runSuiteSetupTeardown(command string, suite string) error {
 	cmd.Stderr = os.Stderr
 	cmd.Env = os.Environ()
 
+	suiteEnv := "internal/suites/" + selectedSuite + "/.env"
```
clems4ever, line 147: Great! authelia-suites binary already handles the suite resources, I think it would be better to move it there so that everything related to resources is colocated. It will avoid the duplication of code too. What do you think?

PR #1057, opened 2020-05-27
[MISC] Storage Schema Versioning Model

### LINKED ISSUE #913: Design: Storage Schema Versioning (db schema)
This is to discuss how we version the schema for upgrade convenience. I have a basic design in mind and plan to try to outline it here as a RFC.

1. We add a table called config.
    * Has two or more columns (could have multiple key columns): key varchar(20), value varchar/text
    * Could have an additional version int column (as well as informational ones like created at), which would get incremented on each update to config, ~would have to investigate if sqlite3 can use `LIMIT` in `SELECT` but think it can~ and we'd just have to do a `SELECT value FROM config WHERE key = 'schema.version' ORDER BY version DESC LIMIT 1`. This would allow us to record configuration states for future use. This could theoretically come at a later schema version which I think I prefer, especially since it has very limited use for this context except recording when a schema was upgraded and what version it was upgraded to.
2. We create a const representing the required/current schema version for each storage provider, usually these would be the same but there is a chance we will have to patch specific DB's. 
3. ~We store schema version upgrades in a `map[int]map[int]string` for each storage provider, where the first int is schema version, second int is order, and the string is the statement itself.~ We store schema version upgrades in their own function, and have a main upgrade function with a switch fallthrough flow.
4. At startup we check config table exists
    1. Detect the schema version:
        * If it doesn't exist, schema version is 0
        * If it does we do `SELECT value FROM config WHERE key = 'schema.version'`, converting the value to int 
    2. If the detected schema version is less than the current version , execute the upgrade command which will:
        1. Run `BEGIN TRANSACTION`
        2. Run individual schema migration statements.
        3. `UPDATE config SET value = %d WHERE key = 'schema.version'`
        4. If any statement fails will `ROLLBACK` and fail startup (we don't want to partially upgrade)
        5. `COMMIT` the changes since they should be successful at this point.

james-d-elliott:
When merged, this will close #913


internal/storage/sql_provider.go
```diff
@@ -36,50 +43,98 @@ type SQLProvider struct {
...
 		if err != nil {
-			return fmt.Errorf("Unable to create table %s: %v", authenticationLogsTableName, err)
+			return err
+		}
+
+		switch version {
+		case 0:
+			err := p.upgradeSchemaVersionTo001(tx, tables)
+			if err != nil {
+				_ = tx.Rollback()
```
clems4ever, line 109: I think we can log the error to be able to know what happens if the statement fails.

internal/storage/upgrades.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = tx.Exec(p.sqlConfigSetValue, "schema", "version", "1")
+	if err != nil {
+		return err
+	}
+
+	if !utils.IsStringInSlice(preferencesTableName, tables) {
```
clems4ever, line 21: what do you think about creating a map with the table name as the key and the sql statement as value and iterate over it to avoid the boilerplate?

internal/storage/sql_provider.go
```diff
@@ -36,50 +43,98 @@ type SQLProvider struct {
 
 	sqlInsertAuthenticationLog     string
 	sqlGetLatestAuthenticationLogs string
+
+	sqlGetExistingTables string
+	sqlCheckTableExists  string
```
clems4ever, line 48: Maybe I missed something but I don't see where this is actually used. Can't we just simply remove it?

internal/storage/upgrades.go
james-d-elliott: That should be easy to do, and a good pattern going forward.

internal/storage/mysql_provider.go
```diff
@@ -17,6 +16,38 @@ type MySQLProvider struct {
...
+			sqlGetLatestAuthenticationLogs: fmt.Sprintf("SELECT successful, time FROM %s WHERE time>? AND username=? ORDER BY time DESC", authenticationLogsTableName),
+
+			sqlGetExistingTables: "SELECT table_name FROM information_schema.tables WHERE table_type='BASE TABLE' AND table_schema=database()",
+
+			sqlConfigSetValue: fmt.Sprintf("REPLACE INTO %s (category, key_name, value) VALUES(?, ?, ?)", configTableName),
+			sqlConfigGetValue: fmt.Sprintf("SELECT value FROM %s WHERE category=? AND key_name=?", configTableName),
+		},
+	}
+
+	provider.sqlUpgradesCreateTableStatements[1][authenticationLogsTableName] = "CREATE TABLE %s (username VARCHAR(100), successful BOOL, time INTEGER, INDEX usr_time_idx (username, time))"
```
clems4ever, line 49: Why is this table treated differently than other? I don't get it.

internal/storage/sql_provider.go
```diff
@@ -6,19 +6,31 @@ import (
...
+	sqlUpgradesCreateTableStatements        map[int]map[string]string
+	sqlUpgradesCreateTableIndexesStatements map[int][]string
 
-	sqlCreateUserPreferencesTable            string
-	sqlCreateIdentityVerificationTokensTable string
-	sqlCreateTOTPSecretsTable                string
-	sqlCreateU2FDeviceHandlesTable           string
-	sqlCreateAuthenticationLogsTable         string
-	sqlCreateAuthenticationLogsUserTimeIndex string
+	/*
```
clems4ever, line 25: can you remove this comment block?

internal/storage/const.go
```diff
@@ -1,44 +1,37 @@
...
-// SQLCreateU2FDeviceHandlesTable common SQL query to create u2f_device_handles table.
-var SQLCreateU2FDeviceHandlesTable = fmt.Sprintf(`
-CREATE TABLE IF NOT EXISTS %s (
-	username VARCHAR(100) PRIMARY KEY,
-	keyHandle TEXT,
-	publicKey TEXT
-)`, u2fDeviceHandlesTableName)
+// sqlUpgradeCreateTableStatements is a map of the schema version number, plus a map of the table name and the statement used to create it.
+// The statement is fmt.Sprintf'd with the table name as the first argument.
+var sqlUpgradeCreateTableStatements = map[int]map[string]string{
```
clems4ever, line 21: Can we create a dedicated type for version?  It is basically an int but at the call site it would be clearer. I was unsure when I saw `sqlUpgradeCreateTableStatements[1]` until I re-read this type definition. With my suggestion that statement would become `sqlUpgradeCreateTableStatements[Version(1)]`

internal/storage/mysql_provider.go
james-d-elliott: Because `CREATE INDEX IF NOT EXIST` statement does not work for MySQL. It does for MariaDB.

internal/storage/const.go
james-d-elliott: Done locally. Also added a ToString() method for it.

internal/storage/sql_provider.go
james-d-elliott: Yeah was just missed in my cleanup due to only committing part of my changes (individual lines). Fixed locally.

internal/storage/mysql_provider.go
clems4ever: Makes sense. Thanks.

internal/storage/sqlmock_provider.go
```diff
@@ -0,0 +1,60 @@
+package storage
```
clems4ever, line 1: I don't see where this mock is used. It seems to be dead code. Please either drop it or add a test using it.

PR #1060, opened 2020-05-29
[DEV] Debug authelia when running a suite.

clems4ever:
This runs dlv debugger within authelia backend container so that an IDE can remote debug the application using port 2345.


internal/suites/example/compose/authelia/docker-compose.backend.dev.yml
```diff
@@ -7,8 +7,13 @@ services:
       args:
         USER_ID: ${USER_ID}
         GROUP_ID: ${GROUP_ID}
+    security_opt:
+      - seccomp:unconfined
```
nightah, line 11:
```suggestion
      - seccomp=unconfined
```

nightah:
Looks like the my nit is the wrong syntax, which means the other line needs to change.

internal/suites/example/compose/authelia/docker-compose.backend.dev.yml
nightah: This is correct syntax.

internal/suites/example/compose/authelia/docker-compose.backend.dev.yml
```diff
@@ -7,8 +7,13 @@ services:
...
+      - apparmor=unconfined
```
nightah, line 12:
```suggestion
      - apparmor:unconfined
```

PR #1065, opened 2020-05-31
[CI] Add Codecov support

### LINKED ISSUE #1061: Add code coverage to Authelia
We don't really know what is properly covered and what is not currently

Codecov might do the job for the time being: https://codecov.io/

nightah:
Closes #1061.


nightah:
So up for discussion @clems4ever @james-d-elliott:

1. Do we limit coverage for the frontend to just `src/components/*` and `/src/App.tsx`?

You can see more details around the items [here](https://buildkite.com/authelia/authelia/builds/2162#63ca1056-ee97-4b18-a793-69a841bff426).

clems4ever:
> So up for discussion @clems4ever @james-d-elliott:
> 
> 1. Do we limit coverage for the frontend to just `src/components/*` and `/src/App.tsx`?
> 
> You can see more details around the items [here](https://buildkite.com/authelia/authelia/builds/2162#63ca1056-ee97-4b18-a793-69a841bff426).

In my opinion, everything should be covered either with unit tests or integration tests so I'd rather not limit the coverage. I think it is low because integration tests are not yet included in the report, right?

nightah:
So including the suites test outputs only gave us a +2.17% increase.

It looks like the reason our coverage is so low is because a lot of the frontend code doesn't have explicit unit tests, even though we may be thoroughly testing it via our integration tests.

You can see a better breakdown here: https://codecov.io/gh/authelia/authelia/tree/170f16aaf36f9bc96416e04758c54a434a2e1ead.

nightah:
Alright this should now be ready to merge.
I've adjusted the tests into three distinct groups, each with their own commit status for coverage too:
- backend
- frontend
- integration

Tests are also only uploaded to codecov assuming that the test run passed successfully, this will ensure that we don't have additional coverage (by virtue of failure paths) being executed in tests.

Though this does beg the question on coverage in general and what/how we should plan to tackle and address this properly. Let's discuss on Matrix so we can plan accordingly.

clems4ever:
> So including the suites test outputs only gave us a +2.17% increase.
> 
> It looks like the reason our coverage is so low is because a lot of the frontend code doesn't have explicit unit tests outside of the components, even though we may be thoroughly testing it via our integration tests.
> 
> You can see a better breakdown here: https://codecov.io/gh/authelia/authelia/tree/170f16aaf36f9bc96416e04758c54a434a2e1ead.

I don't know what brings the 2.17% but it's probably something else than authelia tested in the suites because looking at the report, none of the notifiers are covered as you can see here: https://codecov.io/gh/authelia/authelia/tree/170f16aaf36f9bc96416e04758c54a434a2e1ead/internal/notification

clems4ever:
> Alright this should now be ready to merge.
> I've adjusted the tests into three distinct groups, each with their own commit status for coverage too:
> 
> * backend
> * frontend
> * integration
> 
> Tests are also only uploaded to codecov assuming that the test run passed successfully, this will ensure that we don't have additional coverage (by virtue of failure paths) being executed in tests.
> 
> Though this does beg the question on coverage in general and what/how we should plan to tackle and address this properly. Let's discuss on Matrix so we can plan accordingly.

Great, let's merge it and discuss the rest on Matrix as suggested.

PR #1092, opened 2020-06-07
[MISC] Validate domain set in session is not a wildcard domain.

clems4ever:
Some users spend several hours figuring out the domain must be plain root domain instead of a wildcard. Here we try to let them know as fast as possible with a configuration check.


CHBMB:
@nightah Alright, there was no need to name and shame me!  :grinning: 

Appreciated the help guys, I really did!

nightah:
Just naming no shaming! I’ve probably wasted more hours than I care to share over the journey myself too, so it’s all about making it easier for everybody.

PR #1112, opened 2020-06-12
Example of authelia lite on docker swarm

xplreitr:
Just an example docker-compose file for the deployment of Authelia lite on docker swarm. Its just a modified version of the regular docker-compose file and should be compatible with the configuration and users file. Primary changes were removing depreciated options (name, exposed ports, and restart), moving the labels under deploy, enabling swarm mode on Traefik and added a loadbalencer for each available service.

The only difference in deployment would be `docker stack deploy $STACK_NAME -c docker-compose-swarm.yml` instead of `docker-compose up -d`


clems4ever:
Hello @xplreitr , thank you for this contribution.

From what I see by comparing this new file with the original docker-compose file there is not much difference between them. I think it would be better to keep only one to avoid the duplication leading to potential drifts in the future. Why not just document the tweaks making the original docker-compose file work with Swarm? For instance you could document how to run Authelia on top of Swarm in this section: https://www.authelia.com/docs/community/.

@nightah , do you have an idea why the coverage is impacted by this change?

nightah:
Looks like the coverage issue was due to a node that was just stuck trying to download alpine packages, restarted the job and everything was fine.

@james-d-elliott also made an adjustment to that specific node so hopefully should prevent it blocking like that in future.

clems4ever:
Any news regarding my comment @xplreitr? Does my comment make sense?

clems4ever:
I'm closing this PR. Please reopen it when you have time @xplreitr .

PR #1113, opened 2020-06-14
[FEATURE] Docker simplification and configuration generation

### LINKED ISSUE #1097: Discussion:  Reconfigure docker deployment to make simpler.
Further to a comment I made in the Matrix room.

Both @aptalca and I wondered if there's a couple of ways docker deployment might be made a little easier. 

1. Get the user to map only `/etc/authelia` to a volume or folder on the host. 
2. If `configuration.yml` doesn't exist in `/etc/authelia` copy the sample one to the directory so it's available on the host for the user to start configuring.

If you then extrapolate this further and also by default configure any other files the user might be using, such as `db.sqlite3` and `users_database.yml` are also by default configured to be in `/etc/authelia` it vastly reduces the complexity of mapping these volumes/files.

`users_database.yml.sample` could also be copied across from the image to `/etc/authelia` if `users_database.yml` doesn't exist.

It shouldn't affect existing users who already have configured their `configuration.yml` and file paths and we think it might avoid a lot of the simpler issues arising with a docker deployment by giving authelia a single location on the host for user configurable files & persistent data.

The relevant parts of `configuration.yml` that would be affected are:

```
# log_file_path: /var/log/authelia
```
```
  # file:
  #   path: ./users_database.yml
  #   password:
  #     algorithm: argon2id
  #     iterations: 1
  #     key_length: 32
  #     salt_length: 16
  #     memory: 1024
  #     parallelism: 8
```
```
storage:
  # The directory where the DB files will be saved
  ## local:
  ##   path: /var/lib/authelia/db.sqlite3
```
```
notifier:
 # You can disable the notifier startup check by setting this to true.
 disable_startup_check: false

 # For testing purpose, notifications can be sent in a file
 ## filesystem:
 ##   filename: /tmp/authelia/notification.txt
```

The resulting docker compose would be
```
version: "2.1"
services:
  authelia:
    image: authelia/authelia
    container_name: authelia
    volumes:
      - <path to config folder on host>:/etc/authelia/
    ports:
      - 9091:9091
    restart: unless-stopped
```
And for docker itself
```
docker run -d \
--name=authelia \
--net=docker \
-p 9091:9091/tcp \
-v <path to config folder on host>:/etc/authelia/ \
authelia/authelia
```

As I've said, I'm already up and running (thanks to the help earlier :wink: ) but this may prove useful for new users in terms of making things easier with less potential for errors.

Posted here at the request of @clems4ever for discussion amongst you all.

Thanks again for this wonderful piece of work, I love it!

nightah:
The Authelia binary now will attempt to generate configuration based on the latest template assuming that the config location specified on startup does not exist. If a file based backend is selected and the backend cannot be found similarly it will generate a `users_database.yml` based on the following template:

```yaml
users:
  authelia:
    password: "$argon2id$v=19$m=1048576,t=1,p=8$WGo2dW1IYzBYZi9INHFrZg$oy4aymXxjNYWXwmFAiHJrCkxT75cCUTrkr3nnL5xXl4" # Password is 'authelia'
    email: authelia@authelia.com
    groups:
      - admins
      - dev
```

This will allow more seamless bootstrapping of an environment no matter the deployment method.

We have also squashed the Docker volume requirement down to just `/config` thus removing the requirement for `/var/lib/authelia` this is primarily in attempts to simplify the Docker deployment. 

Users with the old volume mappings have two options:
1. Change their mappings to conform to `/config`
2. Change the container entrypoint from `authelia --config /config/configuration.yml` to their old mapping

**Tasks:**
- [x] Add configuration generation
- [x] Add user database generation for file based backend
- [x] Refactor Docker volumes and configuration to `/config/` path
- [x] Update `BREAKING.md` for existing setups

Closes #1097.


config.template.yml
```diff
@@ -22,7 +22,7 @@ server:
 # Level of verbosity for logs: info, debug, trace
 log_level: debug
 ## File path where the logs will be written. If not set logs are written to stdout.
-# log_file_path: /var/log/authelia
+# log_file_path: /config.log
```
clems4ever, line 25:
```suggestion
# log_file_path: /config/authelia.log
```

docs/configuration/miscellaneous.md
```diff
@@ -55,7 +55,7 @@ Logs can be stored in a file when file path is provided. Otherwise logs
 are written to standard output.
 
 ```yaml
-log_file_path: /var/log/authelia.log
+log_file_path: /config.log
```
clems4ever, line 58:
```suggestion
log_file_path: /config/authelia.log
```

internal/authentication/file_user_provider.go
```diff
@@ -86,6 +97,46 @@ func checkPasswordHashes(database *DatabaseModel) error {
...
+	if err != nil {
+		return fmt.Errorf("Unable to open users_database.template.yml: %v", err)
+	}
+
+	b, err := ioutil.ReadAll(f)
+	if err != nil {
+		return fmt.Errorf("Unable to read users_database.template.yml: %v", err)
+	}
+
+	err = ioutil.WriteFile(path, b, 0644) //nolint:gosec // User database does not need to be 0600.
```
clems4ever, line 132: I think it's better to use 0600 mode as the DB contains hashed password. We don't want any unrelated user on the host or in the container to read it.

internal/configuration/reader.go
```diff
@@ -67,3 +80,22 @@ func Read(configPath string) (*schema.Configuration, []error) {
...
+	if err != nil {
+		return fmt.Errorf("Unable to open config.template.yml: %v", err)
+	}
+
+	b, err := ioutil.ReadAll(f)
+	if err != nil {
+		return fmt.Errorf("Unable to read config.template.yml: %v", err)
+	}
+
+	err = ioutil.WriteFile(configPath, b, 0644) //nolint:gosec // Configuration file does not need to be 0600.
```
clems4ever, line 95: What would it cost to be more restrictive by default using 0600? I mean in a good setup admins should use secrets stored in files as suggested in the doc but what if the admin still provide passwords in the config? They will be available to all users on the host. I think it's safer to be on the restrictive side.
Plus, I don't foresee any direct side effect due to using 0600 instead of 0644. What do you think?

clems4ever:
I also tested the dev workflow on my side, it seems to work properly.

CHBMB:
I've tested it from a users perspective.  `configuration.yml` generated in `/config` on first boot.  Unhashed file based config and hey presto, `users_database.yml` appeared.

I agree with @clems4ever that might as well put the logs in there as well by default if file based logging is configured.  

As far as the sqlite database perms go, fwiw I guess most restrictive is the sensible approach for a security based project.

I mean, I don't think anyone will moan about "overly restrictive" perms, but I can envisage a situation where someone raises concerns, whether those concerns are valid or not is probably not important, but it's not a good issue to have to debate from a PR point of view.

nightah:
Happy to adjust them down to 600 my initial thought was that if we did that the user would need to sudo to modify the files due to the fact that they would be 600 as root, but I guess that isn’t actually a bad thing.

The config not being in the directory was purely an oversight and typo, good catch @clems4ever.

I’ll adjust those elements shortly.

internal/authentication/file_user_provider.go
nightah: Agreed, adjusted.

internal/configuration/reader.go
nightah: There's no particularly downside to doing this so I've adjusted accordingly.

dchidell:
It's worth noting these changes broke my existing docker setup, I am using the lite version with SQLite.

For anyone that it may benefit (like me who updates and only reads the changelog if something breaks) I fixed it by doing the following:

1) Stop the existing container. Mine was restarting continuously given the lack of a MySQL server and using the lite version is not a default config, so it wasn't picking up those differences with the changes to the file paths.

2) Update docker-compose.yml accordingly (I have a custom file for my own file paths).

3) Move the sqlite DB and configuration files into the right place on the parent filesystem

4) Remove the docker volume that was created by the existing container (preventing a mapping being used). This can be found by inspecting the existing stopped authelia container and finding the ID of the volume used for /config, and deleting it with 'docker volume rm <id>' (you will need to delete the authelia container first).

5) Bring up the container as usual with 'docker-compose up -d'

For reference, the error I was getting in the log was as follows (might help someone googling, even though the error doesn't describe the problem well):

```
 level=fatal msg="Unable to initialize SQL database: Unable to create table user_preferences: dial tcp 127.0.0.1:3306: connect: connection refused"
```

nightah:
@dchidell you would have seen that specific error because Authelia would have been using the default configuration which was generated in place of your missing config.

We cover issues that would be breaking specifically in the [`BREAKING.md`](https://github.com/authelia/authelia/blob/master/BREAKING.md).
This is also directly propagated into our releases, so in future check the [respective release](https://github.com/authelia/authelia/releases/tag/v4.20.0) if there's an issue.

dchidell:
Thanks, I ddin't dig deep enough to find the BREAKING.md file, however perhaps I should have put a little more emphasis into the issue that was most perplexing.

When the container was updated (by image only) a docker volume for /config was created. This meant each time a new container was created it defaulted to the docker volume, instead of the volume host mapping in the docker-compose.yml file. The docker volume must be deleted first.

PR #1122, opened 2020-06-17
[MISC] Use /usr/bin/env bash for scripts

vdot0x23:
Hardcoding /bin/bash prevents some users (e.g. NixOS users) from running the scripts.

I got Standalone running and have not done further testing.

### DIFF
--- a/.buildkite/deployment.sh
+++ b/.buildkite/deployment.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -u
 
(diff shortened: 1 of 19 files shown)


nightah:
Thanks for this @vdot0x23 the only two scripts that you would actually be running as an end user would be the `bootstrap.sh` and `cmd/authelia-scripts/authelia-scripts` so the change definitely does make sense in those files.

Just for reference, the others aren't really relevant because they're either run by our CI or within a suite (read: container) but it's not really a breaking change so happy to merge it in.

PR #1123, opened 2020-06-17
[DEV] Set chromedriver path with env variable

vdot0x23:
Hardcoding prevents some users (e.g. NixOS users) from running the scripts.


nightah:
@vdot0x23 the browser path can be set with envionment variable `BROWSER_PATH`.

The only element that's hard-coded is the chromedriver path and you need to ensure that it is in `/usr/bin/chromedriver` path.

You could also initialise this with an environment variable like the `BROWSER_PATH`, but you cannot to set the path with `/usr/bin/env` like it's set in all the shell scripts.

james-d-elliott:
For reference I symlink the location to the chromedriver on distros I have issues on.

clems4ever:
I'd be ok with a CHROMEDRIVER_PATH env var to override the path as we did with the browser path but @vdot0x23 you could also do as @james-d-elliott suggested.

vdot0x23:
@nightah, sorry I didn't realize WIP PRs would be looked at (I used `sed` blindly before actually reading the source).

@clems4ever good suggestion, I did this :)

nightah:
@vdot0x23 no need to apologise, I typically will provide feedback whenever I can get to it because sometimes I may not be able to with large gaps.

Just wanted to steer you in the right direction, no hard feelings either way :).

PR #1124, opened 2020-06-18
[FEATURE] Add configurable display name to frontend

nightah:
This feature allows users with a LDAP backend to specify an attribute (default is "displayname") to retrieve a users name for the portal greeting.
Similarly for the file based backend a new required key "displayname" has been introduced.

This can also be used down the line with OIDC as a separate scope.

Per the `BREAKING.md` if utilising a file based user database:
* Administrators will need to update users and include the `displayname` key.
* Users with long-lived sessions will need to recreate the session (logout and login) to propagate the changes.


config.template.yml
```diff
@@ -151,6 +151,9 @@ authentication_backend:
     # one returned by the LDAP server is used.
     mail_attribute: mail
 
+    # The attribute holding the display name of the user. This will be used to greet an authenticated user.
+    name_attribute: displayname
```
clems4ever, line 155: What do you think about using `display_name_attribute` as the config key to make it more obvious that it's not a first name nor a last name but the actual name that will be displayed. Personally, I find it more explicit and if we add other name attributes in the future (first, last for whatever reason), the distinction will already be made without breaking change.

internal/authentication/file_user_provider.go
```diff
@@ -29,6 +29,7 @@ type FileUserProvider struct {
 // UserDetailsModel is the model of user details in the file database.
 type UserDetailsModel struct {
 	HashedPassword string   `yaml:"password" valid:"required"`
+	Name           string   `yaml:"name" valid:"required"`
```
clems4ever, line 32: I would use `DisplayName` if we decide to agree on my previous comment.

internal/authentication/ldap_user_provider.go
```diff
@@ -126,6 +127,7 @@ func (p *LDAPUserProvider) resolveUsersFilter(userFilter string, inputUsername s
 	// in configuration.
 	userFilter = strings.ReplaceAll(userFilter, "{username_attribute}", p.configuration.UsernameAttribute)
 	userFilter = strings.ReplaceAll(userFilter, "{mail_attribute}", p.configuration.MailAttribute)
+	userFilter = strings.ReplaceAll(userFilter, "{name_attribute}", p.configuration.NameAttribute)
```
clems4ever, line 130: Don't forget to replace the placeholder if we go for display_name_attribute.

internal/handlers/handler_extended_configuration.go
```diff
@@ -8,6 +8,7 @@ import (
 // ExtendedConfigurationBody the content returned by extended configuration endpoint.
 type ExtendedConfigurationBody struct {
 	AvailableMethods    MethodList `json:"available_methods"`
+	DisplayName         string     `json:"display_name"`
```
clems4ever, line 11: ah here you use it :).

config.template.yml
nightah: I'm not particularly fussed either way if we want it to be more explicit happy to rename it to `display_name_attribute` though I think that's a little long and wordy, but I can definitely see your thought process here.

If we decide to introduce split attributes first and/or last names I feel like they would be better suited to be prefixed.

internal/authentication/file_user_provider.go
nightah: Do you envisage that with the file based backend we would introduce keys like `firstname` `lastname` and displayname`?

Again am happy to adjust here if we think it's necessary, though in the case of the file based backend I think if we decided to introduce `firstname` and `lastname` we could still have `name` alongside it and not introduce a breaking change.

config.template.yml
clems4ever: I think the fact that it's wordy is not a big deal as the admin will only read it once in a while, it's not like it's modified so often. But the advantage of it being explicit is that it should be clear in the admin's mind what this attribute is used for by simply reading it. It's also clearer to devs what this variable is supposed to be used for in the code.

internal/authentication/file_user_provider.go
clems4ever: I don't think firstname and lastname will ever be introduced, but who knows, but at least we won't have a conflict if that happens. Anyway, I think the argument about readability (both for admins and devs) is the most important here.

PR #1135, opened 2020-06-20
[MISC] Template global config and refactor some /api endpoints

nightah:
* /api/configuration has been removed in favour of templating said global config
* /api/configuration/extended has been renamed to /api/configuration and display_name has been removed
* /api/user/info has been modified to include display_name


nightah:
Good addition, LGTM!

PR #1148, opened 2020-06-22
[MISC] Detect issue due to breaking change in v4.20.

clems4ever:


nightah:
I not sure that this really fixes the problem in the grand scheme of things. This change would mean that non-Docker deployments; ie: the AUR and possibly even FreeBSD Ports would get a warning if they're conforming to the traditional deployment method/locations.

All reported cases of users that have said their instance was broken was because they run watchtower or an equivalent service and it's blindly upgrading the version of the Authelia, this really goes directly against our recommendation in https://github.com/authelia/authelia/blob/master/BREAKING.md#breaking-changes.
I'm more inclined to just point users that have that issue to the release page for `v4.20.0` or the appropriate section in the `BREAKING.md`.

There might be a little bit of short-term pain and repetativeness but this should be overcome soon.

PR #1157, opened 2020-06-25
[New Feature] Addition of theming base, dark theme and custom theme, customisable with two color codes

BankaiNoJutsu:
I create this PR, for everyone to review the changes/ give feedback.

The base color of the default "light" theme is another blue than original.

 I am unsure on how to do this, but it makes sense to keep the index.tsx as clean as possible. Having a dedicated component for theming/styling is preferred, per Clément's comment. https://github.com/authelia/authelia/pull/1157#r40089773

I would also really like to find a way to improve the color definition, in a cleaner way, in these files:
web/src/layouts/LoginLayout.tsx
web/src/views/LoadingPage/LoadingPage.tsx

I still prefered this method than duplicating the svg files on a per theme basis.

Suggestions welcome!


internal/configuration/schema/theme.go
```diff
@@ -0,0 +1,6 @@
+package schema
+
+// Theme represents the configuration related to Duo API.
+type ThemeConfiguration struct {
+        Theme      string `mapstructure:"theme"`
```
autheliabot, line 5: **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
File is not `gofmt`-ed with `-s` (gofmt)

internal/configuration/schema/theme.go
```diff
@@ -0,0 +1,6 @@
+package schema
+
+// Theme represents the configuration related to Duo API.
```
autheliabot, line 3: **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
comment on exported type ThemeConfiguration should be of the form "ThemeConfiguration ..." (with optional leading article) (golint)

internal/configuration/validator/configuration.go
```diff
@@ -46,6 +47,18 @@
...
 
+	if configuration.Notifier == nil {
+		validator.Push(fmt.Errorf("A notifier configuration must be provided"))
+	} else {
+		ValidateNotifier(configuration.Notifier, validator)
+	}
+
+	if configuration.Theme == "" {
+		configuration.Theme = defaultTheme
+  } else if configuration.Theme != "dark" && configuration.Theme != "light" {
```
autheliabot, line 58: **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
File is not `gofmt`-ed with `-s` (gofmt)

james-d-elliott:
Just thinking about future proofing. Is there any situation where we'd want to add additional configuration variables to themes? I personally think so. If we agree it may make more sense to do this:

```yaml
theme:
  name: dark
```
Instead of:

```yaml
theme: dark
```

internal/configuration/validator/configuration.go
```diff
@@ -9,6 +9,7 @@ import (
 
 var defaultPort = 8080
 var defaultLogLevel = "info"
+var defaultTheme = "default"
```
nightah, line 12:
```suggestion
var defaultTheme = "light"
```

internal/configuration/validator/configuration.go
```diff
@@ -46,6 +47,18 @@ func ValidateConfiguration(configuration *schema.Configuration, validator *schem
...
 
+	if configuration.Notifier == nil {
+		validator.Push(fmt.Errorf("A notifier configuration must be provided"))
+	} else {
+		ValidateNotifier(configuration.Notifier, validator)
+	}
+
+	if configuration.Theme == "" {
+		configuration.Theme = defaultTheme
+  } else if configuration.Theme != "dark" && configuration.Theme != "light" {
```
nightah, line 58: Per reviewdog the indentation here is a little off

internal/configuration/validator/configuration.go
```diff
@@ -46,6 +47,18 @@ func ValidateConfiguration(configuration *schema.Configuration, validator *schem
 		configuration.TOTP = &schema.DefaultTOTPConfiguration
 	}
 
+	if configuration.Notifier == nil {
```
nightah, line 50: I'm curious why this got moved from the bottom of the validation to near the top?

internal/server/index.go
```diff
@@ -16,7 +16,7 @@ var alphaNumericRunes = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUV
 // ServeIndex serve the index.html file with nonce generated for supporting
 // restrictive CSP while using material-ui from the embedded virtual filesystem.
 //go:generate broccoli -src ../../public_html -o public_html
-func ServeIndex(publicDir, base, rememberMe, resetPassword string) fasthttp.RequestHandler {
+func ServeIndex(publicDir, base, theme, rememberMe, resetPassword string) fasthttp.RequestHandler {
```
nightah, line 19:
```suggestion
func ServeIndex(publicDir, base, rememberMe, resetPassword, theme string) fasthttp.RequestHandler {
```

This is a nit more than anything but can we alphabetise these?

internal/server/index.go
```diff
@@ -38,7 +38,7 @@ func ServeIndex(publicDir, base, rememberMe, resetPassword string) fasthttp.Requ
 		ctx.SetContentType("text/html; charset=utf-8")
 		ctx.Response.Header.Add("Content-Security-Policy", fmt.Sprintf("default-src 'self'; object-src 'none'; style-src 'self' 'nonce-%s'", nonce))
 
-		err := tmpl.Execute(ctx.Response.BodyWriter(), struct{ Base, CSPNonce, RememberMe, ResetPassword string }{Base: base, CSPNonce: nonce, RememberMe: rememberMe, ResetPassword: resetPassword})
+		err := tmpl.Execute(ctx.Response.BodyWriter(), struct{ Base, CSPNonce, ThemeName, RememberMe, ResetPassword string }{Base: base, CSPNonce: nonce, ThemeName: theme, RememberMe: rememberMe, ResetPassword: resetPassword})
```
nightah, line 41:
```suggestion
		err := tmpl.Execute(ctx.Response.BodyWriter(), struct{ Base, CSPNonce, RememberMe, ResetPassword, ThemeName string }{Base: base, CSPNonce: nonce, RememberMe: rememberMe, ResetPassword: resetPassword, ThemeName: theme})
```
Here too please :).

web/src/layouts/LoginLayout.tsx
```diff
@@ -3,6 +3,15 @@ import { Grid, makeStyles, Container, Typography, Link } from "@material-ui/core
 import { ReactComponent as UserSvg } from "../assets/images/user.svg";
 import { grey } from "@material-ui/core/colors";
 
+import { useTheme } from '../hooks/Theme';
+
+var color = "#000";
```
nightah, line 8: I think this can be simplified to:

```ts
var color = "#000";
if (useTheme() === "dark") {
  color = "#929aa5"
}

File: web/src/views/LoadingPage/LoadingPage.tsx [hunk 10]
```diff
@@ -1,14 +1,23 @@
 import React from "react";
 import ReactLoading from "react-loading";
 import { Typography, Grid } from "@material-ui/core";
+import { useTheme } from '../../hooks/Theme';
+
+var color = "#000";
```
nightah, line 6: And here too:
```ts
var color = "#000"; 
if (useTheme() === "dark") {
  color = "#929aa5"
}

**Comment by nightah** [comment]
> Just thinking about future proofing. Is there any situation where we'd want to add additional configuration variables to themes? I personally think so. If we agree it may make more sense to do this:
> 
> ```yaml
> theme:
>   name: dark
> ```
> 
> Instead of:
> 
> ```yaml
> theme: dark
> ```

I don't disagree with, thoughis there an example you can think of where we would potentially do this?

**Comment by james-d-elliott** [comment]
I was mainly thinking from the perspective of specific configurable elements. Like maybe you want dark but you want font color to be #34eb3a.

**Comment by BankaiNoJutsu** [comment]
I implemented a custom theme, with this in mind. Currently one main and one secondary color.

I need to update the docs, but so far it is like these keys:

theme:
  maincolor: "#290056"
  secondarycolor: "#fff"

File: internal/configuration/schema/theme.go [hunk 11]
```diff
@@ -0,0 +1,15 @@
...
+        Name                 string `mapstructure:"name"`
+        MainColor            string `mapstructure:"maincolor"`
+        SecondaryColor       string `mapstructure:"secondarycolor"`
+}
+
+// DefaultThemeConfiguration represents default configuration parameters for Theming/Styling.
+var DefaultThemeConfiguration = ThemeConfiguration{
+	Name: "light",
+	MainColor: "#1976d2", //Blue
+  SecondaryColor: "#fff",
```
nightah, line 14: Identation here is off.

internal/configuration/validator/configuration.go
```diff
@@ -42,6 +43,24 @@ func ValidateConfiguration(configuration *schema.Configuration, validator *schem
 		}
 	}
 
+	if configuration.Theme == nil {
+		configuration.Theme = &schema.DefaultThemeConfiguration
+	} else if configuration.Theme.Name != "dark" && configuration.Theme.Name != "light" && configuration.Theme.Name != "custom" {
+		validator.Push(fmt.Errorf("Theme value: %s is not valid, valid themes are: \"light\", \"dark\" or \"custom\"", configuration.Theme.Name))
+	} else if configuration.Theme.Name == "custom" {
+		if configuration.Theme.MainColor == "" {
+			validator.Push(fmt.Errorf("MainColor value: %s is not valid, valid color values are hex from: \"#000000\" to \"#FFFFFF\"", configuration.Theme.Name))
```
nightah, line 52:
```suggestion
			validator.Push(fmt.Errorf("MainColor value: %s is not valid, valid color values are hex from: \"#000000\" to \"#FFFFFF\"", configuration.Theme.MainColor))
```

internal/configuration/validator/configuration.go
```diff
@@ -42,6 +43,24 @@ func ValidateConfiguration(configuration *schema.Configuration, validator *schem
...
+		} else if configuration.Theme.SecondaryColor == "" {
+			validator.Push(fmt.Errorf("SecondaryColor value: %s is not valid, valid color values are hex from: \"#000000\" to \"#FFFFFF\"", configuration.Theme.Name))
```
nightah, line 54:
```suggestion
			validator.Push(fmt.Errorf("SecondaryColor value: %s is not valid, valid color values are hex from: \"#000000\" to \"#FFFFFF\"", configuration.Theme.SecondaryColor))
```

internal/server/index.go
```diff
@@ -16,7 +16,7 @@ var alphaNumericRunes = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUV
 // ServeIndex serve the index.html file with nonce generated for supporting
 // restrictive CSP while using material-ui from the embedded virtual filesystem.
 //go:generate broccoli -src ../../public_html -o public_html
-func ServeIndex(publicDir, base, rememberMe, resetPassword string) fasthttp.RequestHandler {
+func ServeIndex(publicDir, base, rememberMe, resetPassword, thememaincolor, themename, themesecondarycolor string) fasthttp.RequestHandler {
```
nightah, line 19:
```suggestion
func ServeIndex(publicDir, base, rememberMe, resetPassword, themeMainColor, themeName, themeSecondaryColor string) fasthttp.RequestHandler {
```

internal/server/index.go
```diff
@@ -38,7 +38,7 @@ func ServeIndex(publicDir, base, rememberMe, resetPassword string) fasthttp.Requ
 		ctx.SetContentType("text/html; charset=utf-8")
 		ctx.Response.Header.Add("Content-Security-Policy", fmt.Sprintf("default-src 'self'; object-src 'none'; style-src 'self' 'nonce-%s'", nonce))
 
-		err := tmpl.Execute(ctx.Response.BodyWriter(), struct{ Base, CSPNonce, RememberMe, ResetPassword string }{Base: base, CSPNonce: nonce, RememberMe: rememberMe, ResetPassword: resetPassword})
+		err := tmpl.Execute(ctx.Response.BodyWriter(), struct{ Base, CSPNonce, RememberMe, ResetPassword, ThemeMainColor, ThemeName, ThemeSecondaryColor string }{Base: base, CSPNonce: nonce, RememberMe: rememberMe, ResetPassword: resetPassword, ThemeMainColor: thememaincolor, ThemeName: themename, ThemeSecondaryColor: themesecondarycolor})
```
nightah, line 41:
```suggestion
		err := tmpl.Execute(ctx.Response.BodyWriter(), struct{ Base, CSPNonce, RememberMe, ResetPassword, ThemeMainColor, ThemeName, ThemeSecondaryColor string }{Base: base, CSPNonce: nonce, RememberMe: rememberMe, ResetPassword: resetPassword, ThemeMainColor: themeMainColor, ThemeName: themeName, ThemeSecondaryColor: themeSecondaryColor})
```

internal/configuration/schema/theme.go
```diff
@@ -0,0 +1,15 @@
...
+type ThemeConfiguration struct {
+        Name                 string `mapstructure:"name"`
+        MainColor            string `mapstructure:"maincolor"`
+        SecondaryColor       string `mapstructure:"secondarycolor"`
+}
+
+// DefaultThemeConfiguration represents default configuration parameters for Theming/Styling.
+var DefaultThemeConfiguration = ThemeConfiguration{
+	Name: "light",
+	MainColor: "#1976d2", //Blue
```
clems4ever, line 13: So right now anyone would be able to dynamically customize the primary and secondary colors and it's great but I'm wondering (open question) whether we'd really like to have every part of the theme dynamically customizable in the future (font? margins? paddings? borders? etc...). Where do we put the cursor?

I think stopping there is good as it will reduce the scope of what is dynamically customizable to the bare minimum to avoid introducing more code in the backend to support themes (KISS). This could obviously be challenged later but I'd rather keep the code minimal.

However, a user willing to introduce a customization to any part of it should be able do easily do it by creating a new theme easily and rebuild Authelia with it. It's a bit less user friendly but doable and it's also why the theme name brings value otherwise we could just skip the theme name and customize all variables assuming they would be dynamically customizable.

Let me know what you think.

web/src/views/LoadingPage/LoadingPage.tsx
```diff
@@ -1,14 +1,22 @@
 import React from "react";
 import ReactLoading from "react-loading";
 import { Typography, Grid } from "@material-ui/core";
+import { useTheme, useMainColor } from '../../hooks/Theme';
+
+var color = "#000";
+if (useTheme() === "dark") {
```
clems4ever, line 7: A better way to retrieve the color from the theme is to use `useTheme` from `"@material-ui/core"` within the component. With that variable you can easily access the color that has been set in the material theme instead of retrieving the raw data. Thus, you would not this snippet in every part of the app requiring the color.

web/src/layouts/LoginLayout.tsx
```diff
@@ -2,7 +2,14 @@ import React, { ReactNode } from "react";
 import { Grid, makeStyles, Container, Typography, Link } from "@material-ui/core";
 import { ReactComponent as UserSvg } from "../assets/images/user.svg";
 import { grey } from "@material-ui/core/colors";
+import { useTheme, useMainColor } from '../hooks/Theme';
 
+var color = "#000";
+if (useTheme() === "dark") {
```
clems4ever, line 8: A better way to retrieve the color from the theme is to use `useTheme` from `"@material-ui/core"` within the component. With that variable you can easily access the color that has been set in the material theme instead of retrieving the raw data. Thus, you would not need this snippet in every part of the app requiring the color.

internal/configuration/schema/theme.go
clems4ever: Also if the theme name and the colors are not mutually exclusive in the configuration, I'm not sure to understand the need for a `custom` theme. Can you clarify? I mean someone could take any theme as the base and customize the colors, what does `custom` brings here?
If `custom` is just another base theme, why not give it a proper self-explaining name like we did with `light` and `dark`.

web/src/index.tsx
```diff
@@ -4,8 +4,30 @@ import ReactDOM from 'react-dom';
...
+      return themes.light;
+    case 'custom':
+      return themes.custom;
+    default:
+      return themes.light;
+  }
+}
+
+ReactDOM.render(
+  //<ThemeProvider theme={useTheme() === "custom" ? themes.custom : themes.light}>
```
clems4ever, line 25: Please remove dead code.

clems4ever:
Also, there is no test covering the configuration of the theme. Let's keep a good level of coverage.

internal/configuration/schema/theme.go
BankaiNoJutsu: I think it is more maintainable to keep it simple. I would suggest to limit it to colors being fully customizable. As we could go further, but it seems to me that this, as you say, would only make the codebase more complicated than necessary But who knows, maybe there will be a need/envy to extend the theming option in the future with background svg's or else.

I liked the idea of having for any contributor, the ability to create a theme, do a pull request to implement his theme with a defined name, like "light" or "dark", to in the end have an easy way for new users, to change their layouts/colors with predefined/tested themes.

As you say, the other approach would be to have the default theme as of now, and just make both, primary and secondary color variables available for customisation. Without even providing any defined themes.

My idea with the custom theme was that users could try easily/quickly color combinations with just setting two variables in the conf, and that's it. It's really more oriented towards new users which would have not much idea on how to create a new theme, while at the same time having also pre-created themes at their disposal. (for inspiration for example).

I see in fact a change needed, in terms of, if custom is NOT selected then primary_color and secondary_color keys should not be allowed? As they would be ignored anyways.

Not sure you understand my logic with this. If something is still unclear, let me know and i'll try to be more specific.

Maybe you have some other approach/idea ?
Especially if we already consider how we would handle this once there is an admin/user space available for defining theming/unbanning etc.

web/src/layouts/LoginLayout.tsx
BankaiNoJutsu: I am sorry, but i do not fully understand what you mean by this. You seem to see a better approach, which i'm glad. May i ask you to elaborate?

clems4ever: There is hook called `useTheme` in the `@material-ui/core` library. Please use it instead of yours: https://material-ui.com/styles/api/#usetheme-theme

internal/configuration/schema/theme.go
clems4ever: Ok, it's good we are aligned on the customization.

Now regarding the custom theme, I think I got your logic. I'll try to deploy the custom theme to see what's the difference with light basically. Because only reading the code makes me think there might not be much difference between them but maybe I'm wrong and if it's the case we should probably consider changing `custom` into something better describing the look and feel. I'm thinking about something like `raw`, `minimal` or equivalent. Do you see what I mean?

nightah:
Please re-submit this PR from a branch other than `master`.

PR #1332, opened 2020-09-20
Docs Improvement of Traefik2 Example - Add X-ORIGINAL-URL Example

benyanke:
Title says it all! I was able to get this working for my own use, and thought this might be a good thing to have in the traefik example yaml, as I assume most users would like to be redirected back to their app after logging in.

This does have the traefik imposed limitation that it can't detect the original access URL and store that, but at the very least, it seems better to get back to a well known static route in the app instead of going to the authelia "authenticated!" page.

### DIFF
--- a/docs/deployment/supported-proxies/traefik2.x.md
+++ b/docs/deployment/supported-proxies/traefik2.x.md
@@ -105,7 +105,6 @@ services:
       - 'traefik.http.routers.nextcloud.middlewares=nextcloud-redirect@docker,authelia@docker'
       - 'traefik.http.middlewares.nextcloud-redirect.headers.customrequestheaders.X-ORIGINAL-URL=https://nextcloud.example.com"
 
-
     expose:
       - 443
     restart: unless-stopped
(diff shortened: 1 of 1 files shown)


nightah:
@benyanke this should not be necessary with a Traefik based setup. I'm not sure where the following comment comes from:
> This does have the traefik imposed limitation that it can't detect the original access URL and store that

The redirect should occur without specifying `X-Original-URL` as Traefik utilises the other `X-Forwarded-*` headers to construct this dynamically. The other downside of using `X-Original-URL` with Traefik is that you will have a hardcoded redirect no matter what URI you visit on that specific site/host.

I believe that there is either a configuration error in your setup or a potential bug.

Can I suggest that you lodge an issue with the following information to understand the your setup a little better please:
- Authelia Configuration
- Traefik Configuration for:
  - Authelia portal including middlewares
  - Protected endpoint

benyanke:
I can do that! It might take a few days to find time, but I'll get you a bare minimal replication of the issue and put the docker compose file. It should be simple, the backend I'm using is just a static file serving upstream Apache container.

nightah:
Thanks that'd be great.
I've just gone back and double checked manually that redirects are working successfully with Traefik2.

I'll await your issue and in the interim close off this PR.

benyanke:
Followup in #1334, confirmed it's an issue even in a bare-minimum proof of concept.

PR #1333, opened 2020-09-22
[BUGFIX] Fix HAProxy redirects

### LINKED ISSUE #1331: Always being redirected to default_redirection_url
Hi,

I have the exact same problem that #1291 , with very similar haproxy configuration. The user has closed his issue, don't know why, but after several hours I don't know what is going on.

I have made other tests : without the `param default_redirection_url` set in config file, I have a red message "There was an issue fetching the current user state", due to a 404 error when the JS query `Original-URL/api/state`, which obviously don't have such path.

I have ckecked headers, I have the Refer header is there, with a good value.

Finally, if I refresh the page with a good old F5, I access my protected final backend without any problem, because I have the right Authelia cookie afterall.

The only problematic part is the redirect part.

Any idea?

nightah:
This also includes updates to docs examples.

Fixes #1331.

### DIFF
--- a/docs/deployment/supported-proxies/haproxy.md
+++ b/docs/deployment/supported-proxies/haproxy.md
@@ -94,7 +94,7 @@ frontend fe_http
     http-request set-var(req.scheme) str(http) if !{ ssl_fc }
     http-request set-var(req.questionmark) str(?) if { query -m found }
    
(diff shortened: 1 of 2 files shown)


james-d-elliott (approved):
LGTM

PR #1370, opened 2020-10-08
[FEATURE] Container privilege de-escalation

akusei:
This will allow authelia to run as a non-root user. It uses su-exec to step-down privileges from root to the user and group provided by PUID and PGID environment variables. There's also a /log volume in case you'd like to direct your logs to a different location on the host; you'd have to change your configuration.yml to point logs to /logs/authelia.log.

Default user and group are still root and the entrypoint supports running root commands, including /bin/sh through docker run. I wasn't able to test this on all architectures so some additional testing on this would be appreciated.


nightah:
@akusei I'm curious why someone would use this as opposed to just specifying the `user` option to either the `docker run` command or within the `docker-compose.yml`?

**docker run syntax**
```
   -u, --user=""
      Sets the username or UID used and optionally the groupname or GID for the specified command.

   The followings examples are all valid:
      --user [user | user:group | uid | uid:gid | user:gid | uid:group ]
```

**docker compose syntax**
```
user: 1000:1000
```

This doesn't require an addition of an entry script, or adding su-exec into the container.
Also adding the logging volume isn't necessary for a user to redirect the logs themselves, as you have mentioned just changing the location in the `configuration.yml` achieves the same outcome.

akusei:
using the `user` option works fine if you don't have any volumes you'd like to have mounted with non-root user privs. For example, given this docker-compose:

```
version: '3.8'
services:
  authelia:
    image: authelia/authelia:4.22
    container_name: authelia
    restart: unless-stopped
    user: 1000:1000
    expose:
      - 9091
    volumes:
      - ${CONFIG_PATH}:/config
      - ${LOG_PATH}:/logs
    labels:
      traefik.enable: true
      traefik.http.routers.authelia.service: authelia
      traefik.http.services.authelia.loadbalancer.server.port: 9091
      traefik.http.routers.authelia.rule: Host(`${DOMAIN}`)
      traefik.http.routers.authelia.entrypoints: https
      traefik.http.routers.authelia.tls: true
      traefik.http.middlewares.authelia.forwardauth.address: http://authelia:9091/api/verify?rd=https://${DOMAIN}
      traefik.http.middlewares.authelia.forwardauth.trustForwardHeader: true
      traefik.http.middlewares.authelia.forwardauth.authResponseHeaders: Remote-User, Remote-Groups'
```

the above will result in docker creating the missing paths (if you don't pre-create them) as root, when authelia runs with `user: 1000:1000` it won't have access to the mounted volumes because root actually created them. With this method, root is initially running for the entrypoint script which allows you to properly set the mounted volume permissions for the user you're running as then the script steps-down to a non-root user who will then have access to the previously owned by root directories. Does that make sense or am I totally missing the mark on how to do this properly?

akusei:
As for the log volume, you're right, it's not necessary but it's always better to be explicit than not. Like with ENV vars, you don't need to define them in the Dockerfile but it's a good place to do so in case someone is reading your Dockerfile for insights on how to use it. It's not necessary to define PUID or PGID either, but like I said, better to be explicit. Adding the log volume can be removed from this PR but I'd suggest against it, logging is a feature of authelia so why not give the user a nicely defined mount point for overriding the default behavior? They would need to mount a volume if they wanted to change the log config in configuration.yml anyway, so why not.

nightah:
Ah yes of course you're right regarding the first point and using the `user` option.

Personally I would probably be more in favour of adding a section to the docs to step down privileges from root with the `user` option, specifying that the user will also need to create the volume location they plan to bind from the host, with the appropriate permissions. Though I'd also like to hear @james-d-elliott and @clems4ever's thoughts.

Regarding the log volume, we had a lot of trouble in the past when we had multiple bind volumes with new users, I'd prefer to keep the construct as a single volume that can be bind mounted. Having said that, I wouldn't be averse to changing the default logging location from `/config/authelia.log` to `/config/logs/authelia.log` that way so it sits within a dedicated logging directory inside `/config`.

Let's see what the rest of the team think and then we can decide how to move forward.

akusei:
another reason for doing it this way is if your application needs access to files within the container that aren't meant to be mounted externally, if you use the `user` option it will most likely cause the application to fail when it tries to access files within the container where that user doesn't exist. Authelia doesn't have this problem at the moment but it could in the future as it's not an uncommon issue to have with docker which is why you see a lot of other docker images using this same kind of thing

starting as root and stepping down to a non-root user allows for any kind of permission corrections that should be done. I wish this was something that simply using `user` would solve but I honestly don't think there's a solution for that unless your container is completely self contained and has no persistent data.

So for example, if the application requires read access on a file in /etc and you use the `user` option, if the file in /etc isn't world readable the application won't be able to read it. Your only option at that point is to mount a volume to that file in /etc and set the owner on it to what you used in `user`. I've just found that unless you're using a completely ephemeral container with no mounted volumes, the `user` option is generally a pain to work with

akusei:
Any input on this? Using the docker user option isn't valid for authelia in its current state. If you set **user** to a non-root user id and group id with a volume mount to **/config**, authelia will not start. Below is the output of running authelia with no extra configuration from a fresh start using the docker **user** option set to a non-root user:

```
Attaching to authelia
authelia    | time="2020-10-15T05:30:22Z" level=error msg="Unable to find config file: /config/configuration.yml"
authelia    | time="2020-10-15T05:30:22Z" level=error msg="Generating config file: /config/configuration.yml"
authelia    | time="2020-10-15T05:30:22Z" level=error msg="Unable to generate /config/configuration.yml: open /config/configuration.yml: permission denied"
authelia exited with code 1
```

So I really don't see why this is such a sticking point, it won't work with user unless that user is root or the directory structure was created before hand and the appropriate permissions set before hand. If you'd rather force user's to create the directory and set permissions before hand, that's fine, I'll just continue to use my "patch" and you can close this PR.

james-d-elliott:
How will this affect someone who's already set the user using docker (that has the fs setup correctly)? I have a feeling it will fail on an unprivileged user. Otherwise it looks good, maybe a comment about the first conditional.

akusei:
@james-d-elliott the default is 0:0 if you don't specify the PUID and PGID env variables. This shouldn't affect anyone who has already setup the directory structure the way they want it, as long as they use the same user id and group id of the user who owns the directory. If they use a different user id and group id than what they setup for the directory before hand, it will automatically chown to the provided PUID and PGID. I can add another env variable to disable the chown functionality if you'd like.

akusei:
Something like this in the entrypoint should take care of your concerns:

```
currentUser=$(whoami 2>&1)
if [[ ${currentUser} == "root" ]] && [[ ${CHANGE_CONFIG_OWNERSHIP} == "true" ]]; then
  chown -R ${PUID:-0}:${PGID:-0} /logs /config
fi
```

nightah:
The concern here more specifically is if someone has a working setup with the docker user, if they don't add the PUID and PGID environment variables the entry script will run as their defined user and won't be able to privilege escalate to the default (root) hence breaking their setup.

This can be handled in the entry point script a little better for that scenario.

I'm not in front of a pc right now but I'll make some suggestions when I am.

I also still want to drop the separate log volume, so I'll address that in the suggestions too.

akusei:
@nightah The concern is valid and the entrypoint script does need to change to account for that. I'll make some changes and do some tests. Once I get a fully working solution that accounts for your concerns I'll push it up. I'll also remove the logs volume.

james-d-elliott:
Yeah so my theory is the best of both worlds would be by default make it so that the current functionality is 100% preserved. So if both PUID and PGID is 0, just run Authelia without changing user. So if docker is running the container as UID 1000, then as long as the files are accessible by that UID then there will be no issue.

However if either of those are not 0, the new functionality takes over. Does that make sense? This way people also have to be deliberate about this change; which is the only technical way it has an advantage. This may already be a reality, I've not specifically verified what happens if a container is running as UID 1000 and PUID = 0.

akusei:
@james-d-elliott that makes sense and I agree. I almost have a working solution for that but it might take me a bit to nail down and make sure it won't affect existing functionality

akusei:
This should work as you describe, I did testing but it's always good to have others look at it as well. The behavior is as follows:

* if the docker user is set to anything other than root it will use that as the user and will not change dir permissions and will not attempt to step down
* if the user is root and PUID/PGID is defined it will use those as the step down user and change permissions on the directory to match
* no user or PUID/PGID defined will result in authelia running as root
* if PUID or PGID are undefined they will default to 0:0

I did some tests and everything seems to work as expected. The only thing that does not work is running authelia with docker user set to anything other than root with no volumes defined but this is existing behavior. Authelia running as non-root without a volume attempts to write to the internal /config directory which is owned by root.

nightah (changes requested):
So a couple suggestions for changes here.

Can you please also move `entrypoint.sh` to sit in the root project instead of `scripts/`.
The chmod is not necessary as long as you ensure that the executable bit exists on the file when you commit this to the repo, you can ensure this exists when you move the file to the root by running `git update-index --chmod=+x entrypoint.sh`.

I also think `entrypoint.sh` can be simplified significantly, you're defining defaults for the PUID and PGID variables even though they've already been set in the respective Dockerfiles, something like this should do:

```bash
#!/bin/sh

if [[ $(id -u) != 0 ]] && [[ $(id -g) != 0]]; then
  exec authelia "$@"
else
  chown -R ${PUID}:${PGID} /config
  exec "su-exec ${PUID}:${PGID} authelia" "$@"
fi
```

This also should cater for the scenario of a user that has modified the container to run as a different user directly with Docker.

Also lastly could you please rebase from `master`.

akusei:
@nightah Thanks for the suggestions, I'm not the best with bash and forgot about `id` which makes things much simpler. I was initially attempting to account for to case that a user supplied 0 for the user id and a different group id than root. If that's not a concern for you, this will work just fine but there's a couple things I'd ask you to reconsider:

Taking out `if [[ ! -z ${1:-} ]] && [[ "${1:0:1}" != "-" ]]; then exec "$@"` will prevent `docker run authelia/authelia:latest /bin/sh` from working properly as it will always run authelia which will result in the following full command being run `authelia /bin/bash`. The user could override --entrypoint, so it's really not that big of a deal but most people are used to adding a command to the end of the CLI as opposed to overriding --entrypoint. This relates to CLI only, not docker-compose so again, not that big of a deal but a nice thing to have

The chmod change is a little confusing to me, it's not adding a new layer and wouldn't increase the image size much at all. I think it's better to be explicit in the Dockerfile than to rely on git commands to enforce permissions on files inside the image. Again though, this is your project and if that's how you want it, I'll comply but I don't necessarily agree.

The defaulted PUID and PGID in the entrypoint script won't hurt anything if it's kept in, it's explicit and it accounts for the user defining PUID and PGID as empty. In your case, if they specifically define those as empty, the actual value in PUID and PGID in the container will be undefined and will not be set to 0 which causes the entrypoint to fail.

Thoughts?

nightah:
> I was initially attempting to account for to case that a user supplied 0 for the user id and a different group id than root. If that's not a concern for you

When I initially posted the conditional was `&&` I did however change it to `||` so that should cover either scenario.
Basically if the user has specified either a different user or group from the default (root) then we should be executing the entrypoint as it is today.

> Taking out `if [[ ! -z ${1:-} ]] && [[ "${1:0:1}" != "-" ]]; then exec "$@"` will prevent `docker run authelia/authelia:latest /bin/sh` from working properly as it will always run authelia which will result in the following full command being run `authelia /bin/bash`.

Yep on second thought it also means a number of commands that we recommend in other parts of the docs won't work if we don't include said conditional, maybe something like this:

```bash
#!/bin/sh

if [[ ! -z ${1} ]] && [[ ${1} != "--config" ]]; then
  exec "$@"
elif [[ $(id -u) != 0 ]] || [[ $(id -g) != 0 ]]; then
  exec authelia "$@"
else
  chown -R ${PUID}:${PGID} /config
  exec "su-exec ${PUID}:${PGID} authelia" "$@"
fi
```
Changing the second part of the conditional to look for `--config` like our default entrypoint instead of just `-` will mean it's less likely to catch false negatives when running other commands.

I'm not sure I understand what you mean by relying on git commands to enforce permissions, you're setting the executable bit on the file and enforcing the permissions before you commit the file. Once committed whoever clones the repo will pull the file with the executable bit, as intended and it's distributed as such as part of the repo. Other examples include our `bootstrap.sh` and a couple others, really it's just a once off thing and it's leveraged by our CI/CD pipeline. That approach renders the chmod's as unnecessary.

You're absolutely right that not having a default value for PUID and PGID could cause issues if a person decided to deliberately set PUID and PGID to a blank value though. I don't really see why anybody would want to do that, though I have definitely seen more odd things occur. The real point is that if we describe how the variables are intended to be used in the documentation there should be no ambiguity and there's no reason for someone to set either of those values to something blank.
Either way it's not a hill I'm going to die on, you can go ahead and include them if you prefer.

akusei:
Going to be running some tests with the changed entrypoint, will push when done. The defaulted PUID and PGID in the Dockerfile is fine with me, if someone wants to go against documentation and configure things incorrectly, they shouldn't expect it to work so I'll remove the defaults in the entrypoint.

The point I was making with the permissions being managed by git is just that some people use linux, some windows and different file systems have different ways of managing ACLs and permissions so if someone wanted to build from source they may have some issues but if you're confident that won't be a problem, I have no argument against it other than this. So I'll check it in with the execute bit

akusei:
@nightah I think I captured all your suggested changes, I also rebased from master (at least I think I did, I'm not great with git, so if I didn't let me know)

entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+if [[ ! -z ${1} ]] && [[ ${1} != "--config" ]]; then
+  exec "$@"
+elif [[ $(id -u) != 0 ]] && [[ $(id -g) != 0 ]]; then
```
nightah, line 5:
```suggestion
elif [[ $(id -u) != 0 ]] || [[ $(id -g) != 0 ]]; then
```

nightah (changes requested):
Looks mostly good to me just a small nit in the entrypoint script, thanks for adjusting per the feedback.

Are you comfortable adding to the docs for the new environment variable options? If so [`docs/security/measures.md` ](https://www.authelia.com/docs/security/measures.html) is probably a good location, otherwise let me know.

akusei:
I'm not sure why this is failing the some tests. It would make sense if it failed all tests in a given suite, especially if the issue is from the new entrypoint. The tests don't define "user" that I'm aware of and they definitely don't define PUID/PGID so the entrypoint will run as normal as root. I'm guessing I probably rebased from master incorrectly? Any idea why these would be failing @nightah ?

entrypoint.sh
akusei: Is this what would cause the tests to fail?

akusei:
> Looks mostly good to me just a small nit in the entrypoint script, thanks for adjusting per the feedback.
> 
> Are you comfortable adding to the docs for the new environment variable options? If so [`docs/security/measures.md` ](https://www.authelia.com/docs/security/measures.html) is probably a good location, otherwise let me know.

I'll get some documentation in there but it'll most definitely need review as I'm not great at documentation at all

entrypoint.sh
akusei: hmmm, tests are passing now so I'm not sure if was anything wrong with the code. So I'll commit this

nightah:
> I'm not sure why this is failing some tests. It would make sense if it failed all tests in a given suite, especially if the issue is from the new entrypoint. The tests don't define "user" that I'm aware of and they definitely don't define PUID/PGID so the entrypoint will run as normal as root. I'm guessing I probably rebased from master incorrectly? Any idea why these would be failing @nightah ?

I wouldn't worry about the failed tests, I could have re-run them but since you were going to commit more changes they will re-run automatically.

docs/security/measures.md
```diff
@@ -192,18 +190,77 @@ ignoreregex = ^.*level=debug.*
...
+Authelia will run as root by default, there are two options to run as a non-root user. The first option is to use the
+docker `--user` option on the command line or in docker-compose. The second option is to use the `PUID` and `PGID`
+environment variables. An added benefit of using the environment variables is the mounted volumes owner will automatically
+be changed for you.
+
+### Docker user option
+
+With the docker `--user` option, docker will ensure Authelia is running as the user id and group id you specify.
+In order to use this option, you will need to mount the `/config` volume to a directory on the host and set
+the owner and group of that directory to the same user you supplied to docker. NOTE: running Authelia with `--user`
```
nightah, line 221:
```suggestion
the owner and group of that directory to the same user you supplied to docker. Running Authelia with `--user`
```

docs/security/measures.md
```diff
@@ -192,18 +190,77 @@ ignoreregex = ^.*level=debug.*
...
 chain = DOCKER-USER
 ```
 If you are not using Docker remove the the line "chain = DOCKER-USER"
+
+Finally, restart the fail2ban service.
+
+## Container privilege de-escalation
+
+Authelia will run as root by default, there are two options to run as a non-root user. The first option is to use the
+docker `--user` option on the command line or in docker-compose. The second option is to use the `PUID` and `PGID`
```
nightah, line 213:
```suggestion
Docker `--user` option on the command line or in docker-compose. The second option is to use the `PUID` and `PGID`
```

docs/security/measures.md
```diff
@@ -192,18 +190,77 @@ ignoreregex = ^.*level=debug.*
...
+environment variables. An added benefit of using the environment variables is the mounted volumes owner will automatically
+be changed for you.
+
+### Docker user option
+
+With the docker `--user` option, docker will ensure Authelia is running as the user id and group id you specify.
```
nightah, line 219:
```suggestion
With the Docker `--user` option, Docker will ensure Authelia is running as the user id and group id you specify.
```

docs/security/measures.md
```diff
@@ -192,18 +190,77 @@ ignoreregex = ^.*level=debug.*
...
+In order to use this option, you will need to mount the `/config` volume to a directory on the host and set
+the owner and group of that directory to the same user you supplied to docker. NOTE: running Authelia with `--user`
+without mounting a volume to `/config` or incorrectly setting the host system's directory owner will cause Authelia
```
nightah, line 222:
```suggestion
without mounting a volume to `/config` or incorrectly setting the host systems directory owner will cause Authelia
```

docs/security/measures.md
```diff
@@ -192,18 +190,77 @@ ignoreregex = ^.*level=debug.*
...
 ```
 If you are not using Docker remove the the line "chain = DOCKER-USER"
+
+Finally, restart the fail2ban service.
+
+## Container privilege de-escalation
+
+Authelia will run as root by default, there are two options to run as a non-root user. The first option is to use the
+docker `--user` option on the command line or in docker-compose. The second option is to use the `PUID` and `PGID`
+environment variables. An added benefit of using the environment variables is the mounted volumes owner will automatically
```
nightah, line 214:
```suggestion
environment variables. An added benefit of using the environment variables is the mounted volumes ownership will automatically
```

docs/security/measures.md
```diff
@@ -192,18 +190,77 @@ ignoreregex = ^.*level=debug.*
...
+
+### Docker user option
+
+With the docker `--user` option, docker will ensure Authelia is running as the user id and group id you specify.
+In order to use this option, you will need to mount the `/config` volume to a directory on the host and set
+the owner and group of that directory to the same user you supplied to docker. NOTE: running Authelia with `--user`
+without mounting a volume to `/config` or incorrectly setting the host system's directory owner will cause Authelia
+to exit immediately. The docker `--user` option will take precedence over the environment variables.
+
+On the command line, you would create your Authelia data directory, change the owner to your non-root user
```
nightah, line 225:
```suggestion
On the command line, you would create your Authelia data directory, change ownership to your non-root user
```

docs/security/measures.md
```diff
@@ -192,18 +190,77 @@ ignoreregex = ^.*level=debug.*
...
+    image: authelia/authelia
+    container_name: authelia
+    user: 1000:1000
+    volumes:
+      - ./authelia:/config
+```
+
+### PUID/PGID environment variables
+
+If you choose to use the environment variables, the correct owner will be applied automatically on startup of
```
nightah, line 247:
```suggestion
If you choose to use the environment variables, the correct ownership will be applied automatically on startup of
```

nightah (changes requested):
Nice work with the docs, I think once this is set we're good to go.

Also if you want to add the commits directly from the GitHub WebUI you can click on the "Files changed" tab and batch accept the recommendations instead of a singular commit for each suggestion.

akusei:
> Nice work with the docs, I think once this is set we're good to go.
> 
> Also if you want to add the commits directly from the GitHub WebUI you can click on the "Files changed" tab and batch accept the recommendations instead of a singular commit for each suggestion.

I was wondering about that, I knew there had to be a way to batch commit instead of doing those individually, thanks for tip! I'll get those committed

nightah (approved):

nightah:
I just now noticed that you moved around the WORKDIR definition and that's broken where the files are copied.

I'll fix it up in another PR.

PR #1425, opened 2020-11-03
[FEATURE] Add health checks to containers

### LINKED ISSUE #1423: Add docker healthcheck
It would be nice having a docker healthcheck integrated into the dockercompose file, something really clean and simple like polling a special status page would do or an api handle would do.

Currently I use the following healthcheck which seems to work fine:
`wget -q -O - http://localhost:9091/api/state/ | grep '"status":"OK"'`

nightah:
One thing to consider would be if the user is running Authelia with TLS enabled, perhaps an option would be to set a runtime variable that specifies TLS/Non-TLS and that can be utilised to determine the switch for http/https in the health check?

EDIT: Because we can't set the shell's environment variable inside Go's context I think we just stick to adding the healthcheck for the default deployment option (non-TLS) and perhaps add some docs regarding adjusting the healthcheck if the user deploys with TLS, thoughts?

Closes #1423.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,9 +62,11 @@ EXPOSE 9091
 
 VOLUME /config
 
-ENV PATH="/app:${PATH}"
-ENV PUID=0
-ENV PGID=0
+# Set environment variables
+ENV PATH="/app:${PATH}" \
+PUID=0 \
+PGID=0
 
 ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
 CMD ["--config", "/config/configuration.yml"]
+HEALTHCHECK --interval=30s --timeout=3s CMD wget --quiet --tries=1 --spider http://localhost:9091/api/state || exit 1
--- a/Dockerfile.arm32v7
+++ b/Dockerfile.arm32v7
@@ -68,9 +68,11 @@ EXPOSE 9091
 
 VOLUME /config
 
-ENV PATH="/app:${PATH}"
-ENV PUID=0
-ENV PGID=0
+# Set environment variables
+ENV PATH="/app:${PATH}" \
+PUID=0 \
+PGID=0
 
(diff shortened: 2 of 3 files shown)


internal/server/server.go
```diff
@@ -23,6 +23,7 @@
 func StartServer(configuration schema.Configuration, providers middlewares.Providers) {
 	autheliaMiddleware := middlewares.AutheliaMiddleware(configuration, providers)
 	embeddedAssets := "/public_html"
+	healthcheckURL := "://localhost:" + strconv.Itoa(configuration.Port) + "/api/state"
```
autheliabot, line 26: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>SA4006: this value of `healthcheckURL` is never used (staticcheck)

james-d-elliott (approved):
LGTM

Ornias1993:
> Because we can't set the shell's environment variable inside Go's context I think we just stick to adding the healthcheck for the default deployment option (non-TLS) and perhaps add some docs regarding adjusting the healthcheck if the user deploys with TLS, thoughts?

Thats fine, there is a certain movement within the docker community that says one shouldn't, but imho it's so easy to change a healthcheck that we shouldn't worry about it too much :)

Ornias1993:
Note:
You might want to setup the following too:

`--start-period=1m`

That would prevent reboot loops at startup

AndreiArdelean1:
The Authelia service is stuck in a reboot loop because the health check is done for port 9091 while configuration.yml specifies a different port.

nightah:
> Note:
> You might want to setup the following too:
> 
> `--start-period=1m`
> 
> That would prevent reboot loops at startup

This is also addressed in #1444.

PR #1426, opened 2020-11-04
[MISC] Warn for poorly tuned argon2id deployments

### LINKED ISSUE #1391: Authelia is failing silently with insufficient resources
After a few hours of debugging I've managed to discover that Authelia is failing silently when it's assigned insufficient resources in Docker Swarm.
The story is that I had assigned 1 CPU core and 1G memory to it, and in the consequence all login requests were timing out with no logs outputted, not even with the debug logging level. 
This seems like a bug to me

nightah:
The warnings are currently limited to Linux based deployments, however this covers both container and host (static binary) based deployments.
We could potentially look to expand this to FreeBSD in future too.

Closes #1391.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -1,9 +1,12 @@
 package server
 
 import (
+	"io/ioutil"
 	"net"
 	"os"
+	"runtime"
 	"strconv"
+	"strings"
 
 	duoapi "github.com/duosecurity/duo_api_golang"
 	"github.com/fasthttp/router"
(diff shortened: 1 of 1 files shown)


james-d-elliott:
For ref purposes, Anisible's methods for checking related things: https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/facts/virtual/linux.py

james-d-elliott (approved):
LGTM. Regarding the config flag, I thought on it last night. What if we just do a simple ENV var?

nightah:
Per our discussion because we're not force exiting and only warning, there will be no plans to include a config flag or env var.

The following is an example of the expected output:
```
time="2020-11-04T16:07:55+10:00" level=warning msg="Authelia's password hashing memory parameter is set to: 1GB this is 200% of the available memory: 0.5GB"
time="2020-11-04T16:07:55+10:00" level=warning msg="Please read https://www.authelia.com/docs/configuration/authentication/file.html#memory and tune your deployment"
```

If the user is running Linux kernel < 2.6.24 it will result in the following error:
> Error reading hosts memory limit: error here

As of the time of this post the following are the current Linux LTS releases, we don't plan on supporting anything less than the current LTS:

**Longterm release kernels**
| Version | Maintainer | Released | Projected EOL |
|:--------: |:------------:|:----------:|:----------------:|
| 5.4 | Greg Kroah-Hartman & Sasha Levin | 2019-11-24 | Dec, 2025 |
| 4.19 | Greg Kroah-Hartman & Sasha Levin | 2018-10-22 | Dec, 2024 |
| 4.14 | Greg Kroah-Hartman & Sasha Levin |	2017-11-12 | Jan, 2024 |
| 4.9 | Greg Kroah-Hartman & Sasha Levin | 2016-12-11 | Jan, 2023 |
| 4.4 | Greg Kroah-Hartman & Sasha Levin | 2016-01-10 | Feb, 2022 |

PR #1427, opened 2020-11-05
[FEATURE] Add Optional Check for Session Username on VerifyGet

### LINKED ISSUE #1322: Enhance api/verify to include user validation
Currently api/verify validates only on:
authelia_session cookie
X-Forwarded-Host header
X-Forwarded-Proto

I've a use case where I'd like to use pam_exec to validate against api/verify but currently this would allow anyone with a valid authelia_session to login as any user.
e.g.
curl -b "authelia_session=<session_id>" -H "X-Forwarded-Host: <host>" -H "X-Forwarded-Proto: https" -s -o /dev/null -I -w "%{http_code}" -L http://<internal-authelia-service>:80/api/verify

If api/verify was extended to validate again a user value then this could work.

For example if api/verify validated on Remote-User header. This validation would only check authelia_session matches the correct user.

james-d-elliott:
* Adding the Session-Username header to the /api/verify endpoint when using cookie auth will check the value stored in the session store for the username and the header value are the same.

If merged, should close #1322


go.mod
```diff
@@ -32,5 +32,6 @@ require (
 	github.com/tebeka/selenium v0.9.9
 	github.com/tstranex/u2f v1.0.0
 	github.com/valyala/fasthttp v1.15.1
+	golang.org/x/text v0.3.3
```
nightah, line 35: Not sure this is required to be committed back, can you turn a quick `go mod tidy` to confirm?

nightah (approved):
Only real nit is on the `go.mod`.

The warning displayed on potential session hijacking seems a bit wordy but to be honest I can't think of a way to make it any better, LGTM!

PR #1435, opened 2020-11-09
[MISC] Disable auto capitalization of username field on mobile

### LINKED ISSUE #1430: Disable auto capitalize in username field on mobile devices
At the moment when you type in your username on the Authelia page it auto capitalises the first letter.

Seeing that usernames in Authelia are case sensitive (from my own testing) it would be handy to have auto capitalisation turned off as most people will have a username that doesn't contain upper case letters.

https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/Attributes.html#//apple_ref/doc/uid/TP40008058-autocapitalize

https://developers.google.com/web/updates/2015/04/autocapitalize

nightah:
Closes #1430.

### DIFF
--- a/web/src/components/FixedTextField.tsx
+++ b/web/src/components/FixedTextField.tsx
@@ -16,7 +16,8 @@ const FixedTextField = function (props: TextFieldProps) {
                 classes: {
                     root: style.label
                 }
-            }}>
+            }}
+            inputProps={{autoCapitalize: props.autoCapitalize}}>
             {props.children}
         </TextField>
     );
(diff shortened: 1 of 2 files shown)


james-d-elliott (approved):
LGTM

nightah:
FYI this has been tested on iOS and Android and works fine, seems that if you're using a custom keyboard on mobile YMMV with results.

Ultimately this comes down to the mobile keyboard API implementation and we can't fix a poor implementation.

PR #1445, opened 2020-11-11
[BUGFIX] Fix re-rendering callbacks

nightah:
https://github.com/authelia/authelia/pull/1403/commits/b34b10322bee7193b9ff6ba05001bcba7fc0745e introduced a regression where including deps in the associated useCallback functions would cause React to re-render components.
This resulted in unexpected symptoms like multiple Duo push requests, even if a successful or errored request had already been received.

Empty deps/no re-rendering for the respective callbacks is an expected result therefore we can safely ignore these issues the linter is suggesting needs to be fixed.


james-d-elliott (approved):
LGTM

PR #1448, opened 2020-11-11
[DOCS] Update session docs to clarify encryption

### LINKED ISSUE #1411: Misleading docs related to session cookie?
On the [session docs](https://www.authelia.com/docs/configuration/session.html) it says that `session.secret` is used to "encrypt the session cookie". Which to me, means the cookie sent to the user's browser is encrypted. However, looking through the source code I never see `SessionConfiguration.Secret` used outside of Redis code blocks. [This](https://github.com/authelia/authelia/pull/789/files#diff-9e2351f62725f17ebbfe9b0db5b6510f234ae1f25fbc641a80f2c767b1695397R248) also indicates that `session.secret` is only used for Redis.

Should the docs be updated and/or this configuration option moved into the `redis` config block?

nightah:
This looks like it just fell out of sync with what actually already exists within the [`config.template.yml`](https://github.com/authelia/authelia/blob/695cd5bf8f5d964ffff715c1ea14c240299c6899/config.template.yml#L291).

Fixes #1411.

### DIFF
--- a/docs/configuration/session.md
+++ b/docs/configuration/session.md
@@ -22,7 +22,7 @@ session:
   # The name of the session cookie. (default: authelia_session).
   name: authelia_session
 
-  # The secret to encrypt the session cookie.
+  # The secret to encrypt the session data. This is only used with Redis.
   # Secret can also be set using a secret: https://docs.authelia.com/configuration/secrets.html
   secret: unsecure_session_secret
 


Ornias1993 (approved):
Yeah, that looks more like it 👍

james-d-elliott (approved):
LGTM

PR #1450, opened 2020-11-11
[MISC] Append log file instead of overwriting

### LINKED ISSUE #1294: Logs overwritten in log file after reboot
Hi there,

First of all, thank you for this great projet!

I am trying to use a log file instead of the docker stdout in order to parse the log with fail2ban. Everything works great except when I reboot the Authelia container. After restarting, logs are written at the beggining of the file, overwriting existing lines.
Here an example
Image sans titre

I don't know if this is a normal behavior or if I'm missing something.

Here some info :
Docker version 19.03.12
Athelia container : authelia/authelia:latest
log_level: debug
log_file_path: /config/log/authelia.log

nightah:
If Authelia is restarted when a `log_file_path` is defined, upon restart the log file is overwritten.

This change ensures that the log file will be appended to. Users will need to ensure that they rotate/truncate this over time especially if running in `debug` or `trace`.

Fixes #1294.

### DIFF
--- a/config.template.yml
+++ b/config.template.yml
@@ -5,7 +5,7 @@
 # The host and port to listen on
 host: 0.0.0.0
 port: 9091
-# tls_key: /config/ssl/key.pem
+# tls_key: /config/ssl/key.pem
 # tls_cert: /config/ssl/cert.pem
 
 # Configuration options specific to the internal http server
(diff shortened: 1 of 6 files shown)


Ornias1993:
"This change ensures that the log file will be appended to. Isers will need to ensure that they rotate/truncate this over time especially if running in debug or trace."

This is going to cause issues in the long run. Certainly when no documentation is changed.

james-d-elliott (approved):
LGTM

nightah:
> This is going to cause issues in the long run. Certainly when no documentation is changed.

Intent was to also add some docs, which now are included!

Ornias1993 (approved):

PR #1465, opened 2020-11-16
[CI] Fix development workflow

nightah:
Since merging #1135 and utilising Go templating to enable/disable the rememeber me and password reset features these have stopped working in the development workflow.

During frontend development if someone wants to enable/disable these features they should modify the `.env.development` file accordingly.


james-d-elliott (approved):
LGTM

PR #1467, opened 2020-11-16
[CI] Add ability to customise the chromedriver port

nightah:
The development workflow expects chromedriver to be run on the host on port `4444`.
There is currently no mechanism to modify this behaviour at runtime, so if another service is running on `4444` tests will just fail silently.

This change introduces the `CHROMEDRIVER_PORT` environment variable which can be utilised to set a custom port.


james-d-elliott (approved):
LGTM

PR #1472, opened 2020-11-18
[CI] Collect coverage from frontend during integration tests

nightah:
This change will allow us to collect frontend code coverage from our Selenium based integration tests.

Given that the frontend is embedded into the Go binary and the integration tests run with a compiled binary in Docker this poses some issues with the instrumented code and the ability for it to run in this manner.

To fix this we need to relax Authelia's CSP for the integration tests. This is achieved by setting the env variable `ENVIRONMENT` to `dev`.


james-d-elliott (approved):
LGTM

PR #1478, opened 2020-11-21
[DOCS] Make HAProxy regex case insensitive

ntimo:
This makes the regex for the host header check case insensitive, because other wise an attacker could send a HOST header with an uppercase letter and would by this be able to circumvent Authelia.


nightah:
Thanks for the PR @ntimo. Could you please also update the examples in https://github.com/authelia/authelia/blob/master/internal/suites/example/compose/haproxy/haproxy.cfg?

Once this is complete we can look to merge this in.

ntimo:
Hey @nightah I have updated the haproxy.cfg 👍

nightah (approved):

PR #1487, opened 2020-11-24
[MISC] Consistently utilise correct logging interface

nightah:
This change aims to utilise the correct logging interface consistently.

The only instances where stdlib log is utilised is for tests and when commands that Authelia supports; for example certificate generation, password hashing and config validation.


james-d-elliott (approved):
LGTM

PR #1489, opened 2020-11-24
[FEATURE] Create dedicated health endpoint

nightah:
This change points the Docker containers healthcheck to the dedicated `/api/health` endpoint and also includes support for Authelia running with a path prefix.


james-d-elliott (approved):
LGTM

PR #1506, opened 2020-12-03
[CI] Adjust reviewdog filtermode for linting

nightah:
This will ensure that linter errors are picked up for the entire codebase instead of just against the default of [added/modified lines](https://github.com/reviewdog/reviewdog#added-default).


james-d-elliott (approved):
LGTM

Issue #361, opened 2019-04-25
Feature Request: implement password policy


davama:
Hello,

Wondering if possible to include some form of password policy?
Our ldap backend has a ppolicy for password complexity. 
The password reset feature allows you to put any password you want, such as “password1”.

Some complexity ideas:

Min length = X
At least 1 upper/lower
At least 1 digit
At least 1 special

Thank you,
Dave

clems4ever:
It's definitely a great idea. However one has to make sure that no password is directly updated in LDAP because otherwise the policy cannot be enforced or can diverge.

This is a good idea anyway. I'll definitely take it at some point if nobody has done it before.
Thanks for the suggestion.

davama:
Thank you for the positive response!
:+1:

clems4ever:
That could be a starting point: https://github.com/dropbox/zxcvbn

mkey-cdx:
This is a must have feature that would prevent Authelia from weakening efforts to set a password policy.

I would not be afraid of potential password policy drift between Authelia and LDAP in this case because Authelia cannot respect LDAP password policy, and even bypass it. It is therefore better to enforce a policy at application level anyway, instead of leaving the infrastructure with an obvious security hole.

As a containment, I would suggest to add the possibility to remove the _Forgot password?_ link entirely.

tlvenn:
@clems4ever there is actually a go port of this lib: https://github.com/nbutton23/zxcvbn-go

james-d-elliott:
I'm taking a look at this feature due to it's link to 387. @tlvenn really appreciate the link to that. I think I may fork it and fix some of the issues (or look for other public forks). Considering it's a security product and looks unmaintained (open issues with no replies over a year old) I don't think direct use of it would be smart. If the maintainer ever wanted I'd just PR it back into the maintainers repo.

james-d-elliott:
Nightah pointed me to this: https://github.com/trustelem/zxcvbn

davama:
Hello,
Just wondering if any update on this :smile: 

Thank you for the support!!

davama:
Hello,
Sorry if i sound like a broken record....
Just hoping this hasnt been forgotten.
Thank you very much always for the awesome support!!!
Best,
Dave

nightah:
If it's still here as an open issue, it hasn't been forgotten.

[...]

PR #1508, opened 2020-12-03
[MISC] Catch OpenLDAP ppolicy error

nightah:
Further to the discussion over at #361, this change now ensures that OpenLDAP password complexity errors are caught and appropriately handled.

This change also includes the PasswordComplexity test suite in the LDAP integration suite. This is because a ppolicy has been setup and enforced.

### DIFF
--- a/internal/handlers/const.go
+++ b/internal/handlers/const.go
@@ -41,7 +41,7 @@ const unableToRegisterSecurityKeyMessage = "Unable to register your security key
 const unableToResetPasswordMessage = "Unable to reset your password."
 const mfaValidationFailedMessage = "Authentication failed, please retry later."
 
-const ldapPasswordComplexityCode = "0000052D"
+var ldapPasswordComplexityErrors = []string{"0000052D", "LDAP Result Code 19 \"Constraint Violation\": Password fails quality checking policy"}
 
 const testInactivity = "10"
 const testRedirectionURL = "http://redirection.local"
--- a/internal/handlers/const.go
+++ b/internal/handlers/const.go
@@ -41,7 +41,9 @@ const unableToRegisterSecurityKeyMessage = "Unable to register your security key
 const unableToResetPasswordMessage = "Unable to reset your password."
 const mfaValidationFailedMessage = "Authentication failed, please retry later."
 
-var ldapPasswordComplexityErrors = []string{"0000052D", "LDAP Result Code 19 \"Constraint Violation\": Password fails quality checking policy"}
+const ldapPasswordComplexityCode = "0000052D."
+
+var ldapPasswordComplexityErrors = []string{"LDAP Result Code 19 \"Constraint Violation\": 0000052D: Constraint violation - check_password_restrictions: the password is too short. It should be equal or longer than 3 characters!", "LDAP Result Code 19 \"Constraint Violation\": Password fails quality checking policy"}
 
 const testInactivity = "10"
 const testRedirectionURL = "http://redirection.local"
--- a/internal/handlers/const.go
+++ b/internal/handlers/const.go
@@ -43,7 +43,8 @@ const mfaValidationFailedMessage = "Authentication failed, please retry later."
 
 const ldapPasswordComplexityCode = "0000052D."
 
-var ldapPasswordComplexityErrors = []string{"LDAP Result Code 19 \"Constraint Violation\": 0000052D: Constraint violation - check_password_restrictions: the password is too short. It should be equal or longer than 3 characters!", "LDAP Result Code 19 \"Constraint Violation\": Password fails quality checking policy"}
+var ldapPasswordComplexityCodes = []string{"0000052D"}
+var ldapPasswordComplexityErrors = []string{"LDAP Result Code 19 \"Constraint Violation\": Password fails quality checking policy"}
 
 const testInactivity = "10"
 const testRedirectionURL = "http://redirection.local"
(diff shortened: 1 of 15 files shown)


internal/handlers/handler_reset_password_step2.go
```diff
@@ -31,8 +31,8 @@ func ResetPasswordPost(ctx *middlewares.AutheliaCtx) {
 
 	if err != nil {
 		switch {
-		case strings.Contains(err.Error(), ldapPasswordComplexityCode):
-			ctx.Error(fmt.Errorf("%s", err), ldapPasswordComplexityCode)
+		case utils.IsStringInSlice(err.Error(), ldapPasswordComplexityErrors):
```
james-d-elliott, line 34: We may need to think about this more, strings.Contains checks if one string contains another, utils.IsStringInSlice checks if the string is a 100% match for any string in the slice. I'm thinking something like this:

```go
// IsStringInSliceContains checks if a single string is in an array of strings.
func IsStringInSliceContains(a string, list []string) (inSlice bool) {
	for _, b := range list {
		if strings.Contains(a, b) {
			return true
		}
	}

	return false
}
```

james-d-elliott (dismissed):
So the only thing. Think we're introducing a regression if we choose the current method.

internal/handlers/handler_reset_password_step2.go
nightah: Yep good catch, not only this but it looks like the cases in the frontend need to be revisited and fixed.

nightah: Addressed in https://github.com/authelia/authelia/pull/1508/commits/0ec0c5d61d6fc6f2d56e6634f292bce326952603.

web/src/services/Client.ts
```diff
@@ -4,8 +4,8 @@ import { ServiceResponse, hasServiceError, toData } from "./Api";
 export async function PostWithOptionalResponse<T = undefined>(path: string, body?: any) {
     const res = await axios.post<ServiceResponse<T>>(path, body);
 
-    if (res.status !== 200 || hasServiceError(res)) {
-        throw new Error(`Failed POST to ${path}. Code: ${res.status}.`);
+    if (res.status !== 200 || hasServiceError(res)[0]) {
+        throw new Error(`Failed POST to ${path}. Code: ${res.status}. Message: ${hasServiceError(res)[1]}`);
```
nightah, line 8: @clems4ever any thoughts as to if you're okay with handling bubbling the error up this way?
Is there another specific way you want this handled?

internal/handlers/handler_reset_password_step2.go
```diff
@@ -31,7 +31,7 @@ func ResetPasswordPost(ctx *middlewares.AutheliaCtx) {
 
 	if err != nil {
 		switch {
-		case strings.Contains(err.Error(), ldapPasswordComplexityCode):
+		case utils.IsStringInSliceContains(err.Error(), ldapPasswordComplexityErrors):
 			ctx.Error(fmt.Errorf("%s", err), ldapPasswordComplexityCode)
```
james-d-elliott, line 35:
```suggestion
                case utils.IsStringInSliceContains(err.Error(), ldapPasswordComplexityCodes):
                        ctx.Error(fmt.Errorf("%s", err), ldapPasswordComplexityCode)
		case utils.IsStringInSlice(err.Error(), ldapPasswordComplexityErrors):
			ctx.Error(fmt.Errorf("%s", err), ldapPasswordComplexityCode)
```

So this was a bit more what I had in mind. Mainly think this will mean we can match error codes (0000052D), which should capture most errors. But we can also match specific error messages exactly.

internal/handlers/const.go
```diff
@@ -41,7 +41,9 @@ const unableToRegisterSecurityKeyMessage = "Unable to register your security key
 const unableToResetPasswordMessage = "Unable to reset your password."
 const mfaValidationFailedMessage = "Authentication failed, please retry later."
 
-const ldapPasswordComplexityCode = "0000052D"
+const ldapPasswordComplexityCode = "0000052D."
+
+var ldapPasswordComplexityErrors = []string{"LDAP Result Code 19 \"Constraint Violation\": 0000052D: Constraint violation - check_password_restrictions: the password is too short. It should be equal or longer than 3 characters!", "LDAP Result Code 19 \"Constraint Violation\": Password fails quality checking policy"}
```
james-d-elliott, line 46:
```suggestion
var ldapPasswordComplexityErrors = []string{"LDAP Result Code 19 \"Constraint Violation\": Password fails quality checking policy"}
var ldapPasswordComplexityCodes = []string{"0000052D"}
```

To be considered with the other part of my review.

web/src/services/Client.ts
james-d-elliott: I have a feeling this is not necessary to worry about. The only time this would theoretically be presented to an attacker is if they already compromised the users email right?

james-d-elliott (commented):
Mostly a what do you think? I'm happy to go ahead if you think it's not valuable.

web/src/services/Api.ts
```diff
@@ -58,7 +58,7 @@ export function toData<T>(resp: AxiosResponse<ServiceResponse<T>>): T | undefine
 export function hasServiceError<T>(resp: AxiosResponse<ServiceResponse<T>>) {
     const errResp = toErrorResponse(resp);
     if (errResp && errResp.status === "KO") {
-        return true;
+        return [true, errResp.message];
```
clems4ever, line 61: I advise you to use a dict instead of a tuple, that way at the caller you can refer to a name instead of [0] which is more user friendly for other developers.

clems4ever (dismissed):

james-d-elliott (approved):
LGTM

PR #1511, opened 2020-12-04
[MISC] Refactor and address most errcheck linter ignores

nightah:
This is mostly a quality of life change.
When we first implemented the errcheck linter we ignored a number of items in our legacy codebase with intent to revisit down the track.

There's a only a few leftover TODO's, these have been left for the team to discuss on Matrix and agree on the way forward.

### DIFF
--- a/internal/handlers/handler_firstfactor.go
+++ b/internal/handlers/handler_firstfactor.go
@@ -95,7 +95,8 @@ func FirstFactorPost(msInitialDelay time.Duration, delayEnabled bool) middleware
 
 		if err != nil {
 			ctx.Logger.Debugf("Mark authentication attempt made by user %s", bodyJSON.Username)
-			ctx.Providers.Regulator.Mark(bodyJSON.Username, false) //nolint:errcheck // TODO: Legacy code, consider refactoring time permitting.
+
+			_ = ctx.Providers.Regulator.Mark(bodyJSON.Username, false)
 
 			handleAuthenticationUnauthorized(ctx, fmt.Errorf("Error while checking password for user %s: %s", bodyJSON.Username, err.Error()), authenticationFailedMessage)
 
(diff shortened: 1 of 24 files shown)


james-d-elliott (approved):
LGTM

internal/handlers/handler_firstfactor.go
```diff
@@ -95,7 +95,8 @@ func FirstFactorPost(msInitialDelay time.Duration, delayEnabled bool) middleware
 
 		if err != nil {
 			ctx.Logger.Debugf("Mark authentication attempt made by user %s", bodyJSON.Username)
-			ctx.Providers.Regulator.Mark(bodyJSON.Username, false) //nolint:errcheck // TODO: Legacy code, consider refactoring time permitting.
+
+			_ = ctx.Providers.Regulator.Mark(bodyJSON.Username, false)
```
clems4ever, line 99: Why not handle the error here?

internal/handlers/handler_firstfactor.go
```diff
@@ -104,7 +105,8 @@ func FirstFactorPost(msInitialDelay time.Duration, delayEnabled bool) middleware
 
 		if !userPasswordOk {
 			ctx.Logger.Debugf("Mark authentication attempt made by user %s", bodyJSON.Username)
-			ctx.Providers.Regulator.Mark(bodyJSON.Username, false) //nolint:errcheck // TODO: Legacy code, consider refactoring time permitting.
+
+			_ = ctx.Providers.Regulator.Mark(bodyJSON.Username, false)
```
clems4ever, line 109: why not handle the error here?

web/tsconfig.json
```diff
@@ -17,7 +17,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react",
+    "jsx": "react-jsx",
```
clems4ever, line 20: ah yeah, I've seen that one at work XD.

clems4ever (approved):
Only a few comments but I'm ok for merging it like this since it already improves the situation a lot.

internal/handlers/handler_firstfactor.go
nightah: I had issues when running integration tests and I thought it was related.

After some testing it appears that it's not so I'll handle these too.

nightah: Addressed in https://github.com/authelia/authelia/pull/1511/commits/15646c4bedd408da2fe42ddef2c07318c3a54c25.

internal/handlers/handler_firstfactor.go
nightah: Addressed in https://github.com/authelia/authelia/pull/1511/commits/15646c4bedd408da2fe42ddef2c07318c3a54c25.

PR #1533, opened 2020-12-14
[DOCS] Add a note on regexps in ACLs.

### LINKED ISSUE #1523: Escape character error in regular expressions of resources (Access Control Policies)
Hi there,

It seems Authelia doesn't support escape character in regex :
`time="2020-12-10T16:39:31Z" level=error msg="Error malformed yaml: line 34: found unknown escape character"`

The related `configuration.yml` part with the rule trying to match a string including `?`:
```
...
access_control:
  rules:
  - domain: secure.example.com
      resources:
        - "^/auth\?.*$"
      policy: one_factor
...
```

clems4ever:
Fixes #1523


james-d-elliott (approved):
LGTM. I think we can eventually consider methods to check query param, but I don't think it's necessary in the use case of the user.

Issue #450, opened 2019-11-18
Create a swagger description to document API


clems4ever:
It would be helpful to have a comprehensive documentation of the endpoints to help configure Authelia correctly in real life environments.

james-d-elliott:
https://github.com/swaggo/swag

PR #1534, opened 2020-12-14
[MISC] Make the unexisting endpoints in the API return 404 instead of…

### LINKED ISSUE #1520: Returning login page (200) for all paths is dangerous
This seems dangerous as a single typo in the reverse-proxy config could send requests, for example, to `/api/verifi`, authelia would reply with a 200 and the original request would go through.

From what I gather (haven't looked at the code):

 - `/static/*` serves files, or a 404
 - `/api/verify` works as expected
 - `/*` serves the login HTML

So in case of a typo, `/api/x` hits `/*` which returns a 200, except if x is one of the other existing endpoints like state or info, but for a typo that is extremely unlikely.
An admin would likely quickly notice their server is entirely unprotected, but it's better to not make this situation possible.

I propose only returning the login page for `/`, and returning an error by default.

clems4ever:
… 200.

Fixes #1520

### DIFF
--- a/internal/server/index.go
+++ b/internal/server/index.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 	"text/template"
 
 	"github.com/valyala/fasthttp"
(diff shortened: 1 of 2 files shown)


james-d-elliott (approved):
LGTM

nightah:
Merging this will cause issues for the PR I'm working on to resolve #450. The PR will utilise the same indexhandler with a different path for assets.

I'm planning to serve out the api docs/swaggerui from `/api/` and the JavaScript for swaggerui has its own hash router. So unless we choose a different path to serve out the assets this will conflict.

james-d-elliott:
Yeah I saw that in Matrix. I'll leave it as is for now. We can figure out the solution later as this change is fairly small.

nightah:
Now that #1544 has been merged this needs to be reconsidered.

Our options are:

1. Apply this change and move swagger-ui from being served at `/api/`.
2. Define all of the frontend paths and drop the NotFound handler.
3. Do nothing.

@clems4ever, @james-d-elliott probably worth collating thoughts to move this ahead.

PR #1537, opened 2020-12-15
[DOCS] Update hash-password example with single quotes

### LINKED ISSUE #1536: Doc: Passwords passed to hash-password must be quoted
Otherwise:

```bash
$ sudo docker run authelia/authelia:latest authelia hash-password Some!1
sudo docker run authelia/authelia:latest authelia hash-password Someexit
```

bbros-dev:
Closes #1536

### DIFF
--- a/docs/configuration/authentication/file.md
+++ b/docs/configuration/authentication/file.md
@@ -90,9 +90,10 @@ you could generate a 16 byte salt and provide it with the `--salt` flag.
 Example: `authelia hash-password --salt abcdefghijklhijl`. For argon2id the salt must 
 always be valid for base64 decoding (characters a through z, A through Z, 0 through 9, and +/).
 
+Passwords passed to `hash-password` must always be single quoted to prevent parameter subtitution.
 For instance to generate a hash with the docker image just run:
 
-    $ docker run authelia/authelia:latest authelia hash-password yourpassword
+    $ docker run authelia/authelia:latest authelia hash-password 'yourpassword'
     Password hash: $argon2id$v=19$m=65536$3oc26byQuSkQqksq$zM1QiTvVPrMfV6BVLs2t4gM+af5IN7euO0VB6+Q8ZFs
 
 Full CLI Help Documentation:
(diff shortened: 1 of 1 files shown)


docs/configuration/authentication/file.md
```diff
@@ -90,9 +90,10 @@ you could generate a 16 byte salt and provide it with the `--salt` flag.
 Example: `authelia hash-password --salt abcdefghijklhijl`. For argon2id the salt must 
 always be valid for base64 decoding (characters a through z, A through Z, 0 through 9, and +/).
 
+Passwords passed to `hash-password` must always be single quoted to prevent parameter subtitution.
```
nightah, line 93:
```suggestion
Passwords passed to `hash-password` should be single quoted if using special characters to prevent parameter substitution.
```

nightah (changes requested):
Just a small nit, otherwise I'm okay with this.

docs/configuration/authentication/file.md
bbros-dev: Hmm, I've come lean the other way and I now think it better to remove the reference to parameter substitution. Generalyl I think people would be surprised to find that `!` is one of the characters that triggers some kind of substitution.  What are all the others?  People will freeze wondering what character combinations must be quoted and what can be left un-quoted, I did - after wasting hours trying to workout why the password some!1 never worked..... consistency is simplicity. 

Single quote everything and drop all mention of special character and character combinations that may or may not be expanded.

At the moment always quote is a documentation admonition. Eventually the code should enforce it, but that is a low priority, and lower still with a document instruction in place?

clems4ever: Hello @bbros-dev , quotes are interpreted by the shell not authelia tool so there is nothing we can enforce with code unfortunately.

The version of the doc proposed by @nightah makes sense to me since it gives more details on the why it should be single quoted. I think it's the best we can do for users on that one.

nightah: I view the docs as an opportunity to educate instead just giving instructions for the user to blindly follow.

Because the example in there is still surrounded by a single quote it means if they decide to just copy paste the command they likely will not run into issues, but by telling the user that only special characters strictly would require you to use single quotes it gives them an understanding of why that suggestion is there.

It could just be me but I don't think that suggesting "special characters" should confuse a user, while I may know which characters do and don't need to be quoted the intent of the wording "special characters" is to advise anything that isn't alphanumeric, if you don't believe that's clear I'm happy for you to change the wording to explicitly suggest that.

nightah (approved):
LGTM.

PR #1546, opened 2020-12-17
[BUGFIX] Set server name in TLS options for LDAP.

james-d-elliott:
* fixes issue with LDAP validation of the host when insecureskipverify is disabled


james-d-elliott:
Replaced by 1557

Issue #1517, opened 2020-12-08
LDAP groups for user always empty


zmiguel:
Hi, I'm unable to setup authelia using groups permission because the ldap groups for a user are always empty

**My Setup** *(docker)*

*osixia/openldap:latest* - as OpenLDAP server
*wheelybird/ldap-user-manager:v1.4* - as OpenLDAP Manager

**Authelia config**

```
ldap:
    implementation: custom
    url: ldap://openldap
    skip_verify: true
    start_tls: false
    minimum_tls_version: TLS1.2

    base_dn: dc=zmiguel,dc=me

    # An additional dn to define the scope to all users.
    additional_users_dn: ou=people

    users_filter: (&({username_attribute}={input})(objectClass=person))

    # An additional dn to define the scope of groups.
    additional_groups_dn: ou=groups
    
    groups_filter: (&(member={dn})(objectclass=groupOfNames))

    # The attribute holding the name of the group
    group_name_attribute: cn

    user: cn=admin,dc=zmiguel,dc=me
    password: <admin_password>
```
If using 
```
rules:
    - domain: "mydomain.example"
      subject: "group:users"
      policy: one_factor
```
I always get access denied because Authelia can't find the groups for my user.

If I remove `subject: "group:users"` it works because authelia ignores the groups

**Authelia Logs**

```
authelia             | time="2020-12-08T21:30:51Z" level=trace msg="LDAP client starts a session over raw TCP"
authelia             | time="2020-12-08T21:30:51Z" level=trace msg="Computed user filter is (&(uid=zmiguel)(objectClass=person))"
authelia             | time="2020-12-08T21:30:51Z" level=trace msg="LDAP client starts a session over raw TCP"
authelia             | time="2020-12-08T21:30:51Z" level=debug msg="Credentials validation of user zmiguel is ok" method=POST path=/api/firstfactor remote_ip=redacted
authelia             | time="2020-12-08T21:30:51Z" level=debug msg="Mark authentication attempt made by user zmiguel" method=POST path=/api/firstfactor remote_ip=redacted
authelia             | time="2020-12-08T21:30:51Z" level=trace msg="LDAP client starts a session over raw TCP"
authelia             | time="2020-12-08T21:30:51Z" level=trace msg="Computed user filter is (&(uid=zmiguel)(objectClass=person))"
[...]

PR #1549, opened 2020-12-17
Add config example for LDAP groupOfUniqueNames group structure

zmiguel:
Based on #1517 

Adding a comment to the LDAP group section of the config file to tell users they need to use a different group_filter if their LDAP groups use the `groupOfUniqueNames` structure instead of the expected `groupOfNames` structure.

Not doing so result in Authelia not finding the groups for the user and not authorizing group-based access.

### DIFF
--- a/config.template.yml
+++ b/config.template.yml
@@ -161,6 +161,7 @@ authentication_backend:
     # - {mail_attribute} is a placeholder replaced by what is configured in `mail_attribute`.
     # - DON'T USE - {0} is an alias for {input} supported for backward compatibility but it will be deprecated in later versions, so please don't use it.
     # - DON'T USE - {1} is an alias for {username} supported for backward compatibility but it will be deprecated in later version, so please don't use it.
+    # If your groups use the `groupOfUniqueNames` structure use this instead: (&(uniquemember={dn})(objectclass=groupOfUniqueNames))
     groups_filter: (&(member={dn})(objectclass=groupOfNames))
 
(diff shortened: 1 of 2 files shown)


james-d-elliott (changes requested):
LGTM, which LDAP standard/backend are you using? RFC4519?

Also just need to adjust this file too (same change as the config template):

https://github.com/authelia/authelia/blob/master/docs/configuration/authentication/ldap.md

zmiguel:
> LGTM, which LDAP standard/backend are you using? RFC4519?

I'm using OpenLDAP configured to use RFC2307BIS_SCHEMA

I initialized my LDAP database using the _[wheelybird/ldap-user-manager](https://hub.docker.com/r/wheelybird/ldap-user-manager)_ docker container that uses `groupOfUniqueNames` by default for groups.

This pull request is to provide more information to new users that might be using the same setup on their docker deployments that they need to change the group filter so authelia correctly picks up the groups for the given username.

james-d-elliott (approved):
LGTM

james-d-elliott:
I was just curious for the future, it's nice to know what our users are using so we can setup LDAP implementation templates like this: https://www.authelia.com/docs/configuration/authentication/ldap.html#implementation

Thanks for contributing! If you were interested in contributing in a future PR we could assist either via GitHub, Matrix, or Discord (see badges on https://github.com/authelia/authelia to join Matrix/Discord).

PR #1551, opened 2020-12-20
[FEATURE] Display TOTP secret on device registration

### LINKED ISSUE #1509: Feature request: make TOTP secret explicit during registration
Hi everyone !

I would like to request adding the TOTP secret in a little text box near the QRcode during the registration process.
I personally prefer to register from my computer, and simply copy-pasting a sequence of characters is easier than scanning a QRcode.

It was done in v3, and it should be doable to implement it again:

> We also used to do this with Authelia v3, I don't see why we couldn't do it again, please lodge a new issue with that specific request and we can address accordingly.

_Originally posted by @nightah in https://github.com/authelia/authelia/issues/1505#issuecomment-738001209_

What do you guys think ? :)

Edit: just found out that the link of the QRcode image contains the secret, so it's even less of a hassle now since I don't need to scan the picture.

nightah:
This change provides the TOTP secret which allows users to copy and utilise for password managers and other applications.

Screenshot_2020-12-20-03-21-51

**EDIT:** I've made a slight modification to the view, refer to my [comment below](https://github.com/authelia/authelia/pull/1551#issuecomment-750701251).

Closes #1509.

### DIFF
--- a/internal/suites/action_totp.go
+++ b/internal/suites/action_totp.go
@@ -2,6 +2,7 @@ package suites
 
 import (
 	"context"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


nightah:
I've removed the inline icon and added icons under text field which allow copying of the secret key and the whole OTP URL.

Here's what it looks like now:

Screenshot_2020-12-24-12-40-23

james-d-elliott (approved):
LGTM

james-d-elliott (approved):

james-d-elliott (approved):
LGTM

PR #1564, opened 2020-12-28
[BUGFIX] Static Session Expiration Key

### LINKED ISSUE #1059: [K8S] DaemonSet causes remember me to not be honored (4.19+)
The upgrade to the session provider in 4.19 causes a strange interaction only when Authelia is running in multiple pods (using redis). The remember me duration is not honored, instead the cookie expires based on the session expiration value.

It is confirmed that this interaction only happens with multiple pod architecture like a deployment with multiple pods or a daemonset replicated on all worker nodes (more than one).

james-d-elliott:
* keys for session expiration are random for each instance of Authelia
* this is caused by upstream setting it to a random value
* using a temporary bugfix fork of github.com/fasthttp/session to resolve locally

Should theoretically fix #1059 if accepted and merged. Requires a bugfix fork of github.com/fasthttp/session until the proper fix is determined upstream.


nightah (approved):
LGTM.

PR #1565, opened 2020-12-29
[CI] Add linting option for frontend and enforce styling

nightah:
We now extend the default React Eslint configuration and enforce styling with prettier for all of our frontend code.

An example of the linter output can be found [here](https://github.com/authelia/authelia/runs/1621966777).


james-d-elliott (approved):
LGTM

PR #1575, opened 2021-01-01
[DOC] Correct private IP range

Kristof-Mattei:
This PR updates the trusted IP range in the nginx example.

The example only trusts the correct IP from the Private Internets (as described here: https://tools.ietf.org/html/rfc1918) however one of the ranges was too large. 

The correct ranges are:
-  `10.0.0.0/8`
-  `172.16.0.0/12`
-  `192.168.0.0/16`

Source: https://tools.ietf.org/html/rfc1918#section-3

Happy New Year!


james-d-elliott (approved):
LGTM

james-d-elliott:
Nice catch.

PR #1595, opened 2021-01-13
[MISC] Validate all sections of ACLs on startup

nightah:
This change ensures that all sections of the `access_control` key are validated on startup.


james-d-elliott (approved):
LGTM

PR #1604, opened 2021-01-15
[MISC] Fix OTP countdown circle

nightah:
On the method selection page previously the OTP option had a circle which gradually reduced in size and eventually reset, this was a intended as a visual indicator for a time based OTP.

Currently this is broken and in turn we see the following:
Screenshot_2021-01-15-03-36-01

With this fixed implemented we see the following:
Peek 2021-01-15 15-39


james-d-elliott (approved):
LGTM

PR #1607, opened 2021-01-16
[MISC] Add missing CLI suite test 

nightah:
* Add missing test for `authelia version` command in CLI suite.
* Standardise logger calls.
* Swap CSP switch order to ensure correct CSP is applied in specific scenarios.


james-d-elliott (approved):
LGTM

PR #1632, opened 2021-01-25
feat(authorizer): request method acl rules

james-d-elliott:
If merged should fix #648 and #1624

PR #1640, opened 2021-01-28
perf(authorizer): preload access control lists

### LINKED ISSUE #648: Failing http-post requests because of automatic preflight-requests

### LINKED ISSUE #1339: Multi-tenant app support?
Hi, I'm researching SSO solutions for a few applications we develop. They are multi-tenant "enabled". This means when a user makes a request for "foo.example.org", the application is configured to only fetch records for "foo" just for that request.
Users can belong to one or more tenants.

I was looking at the access control documentation but I think the thing I had in mind is not possible? E.g.:

- domain: "(*).example.com"
  subject:
       - "group:$1"
  policy: one_factor

So the tenant id would be a capture group that can be referenced later in the group.

Thanks in advance.

### LINKED ISSUE #1624: infinite redirects
[configuration.yml.txt](https://github.com/authelia/authelia/files/5853533/configuration.yml.txt)
[docker-compose.yml.txt](https://github.com/authelia/authelia/files/5853534/docker-compose.yml.txt)
[middlwares.yml.txt](https://github.com/authelia/authelia/files/5853535/middlwares.yml.txt)

**Issue:** When browsing to a website protected by authelia, it infinitely redirects to itself resulting in a very long URL / Header. This of course errors chrome out.
**Environment**: Docker / Traefik 2 / Authelia / middleware files / secrets

Within the configuration file, if I add authelia.example.com with bypass, this fixes the issue.
However in the examples, I do not see this done - yet I assume it works as intended?

What lead me here was my original config has authelia set to bypass and whenever I went to a site protected by authelia it would let me regardless of what user I used. In my files I have plexpy restricted to subject: "user:bob" however I can access the site with another user account.
I thought possibly this bypass I set for authelia was allowing this behavior, since technically the redirect URL contains authelia in it.

### LINKED ISSUE #1765: BUG: ACL rules with a subject above bypass rules without a subject get skipped
Currently with rules like the following the second rule is totally skipped. 

```yaml
default_policy: deny
rules:
  - domain: foo.bar.com
    policy: one_factor
    subject: "group:admins"
  - domain: baz.bar.com
    resources:
      - "^/super/secret.*$"
    policy: one_factor
    subject: "group:admins"
  - domain: baz.bar.com
    policy: bypass
```

james-d-elliott:
~~Additional code review in #1632.~~

Fixes #1339. Fixes #648. Fixes #1624. Fixes #1765.

To Do:

- [x] schema validation: invalid methods


internal/utils/strings_test.go
```diff
@@ -50,42 +58,62 @@
...
-	s := IsStringInSliceContains(a, b)
-	assert.False(t, s)
+	slice := []string{"abc", "onetwothree"}
+
+	assert.False(t, IsStringInSliceContains(a, slice))
+}
+
+func TestShouldFindStringInSliceFold(t *testing.T) {
+	a := "xyz"
+	b := "ABC"
```
autheliabot, line 105: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>string `ABC` has 2 occurrences, make it a constant (goconst)

nightah (approved):
Broadly this LGTM, couple small nits.

PR #1680, opened 2021-02-01
refactor(suites): simplify kubernetes suite

nightah:
This PR achieves the following goals:
* Utilise upstream version of kind instead of a patched version which allows binding to networks other than the default "kind"
* Utilises the registry cache which is setup one level above the kind cluster

The former point was required to successfully run our integration tests in a Kubernetes environment, however this is now possible without running a patched version of kind.

The second point is because DockerHub has introduced rate limiting for container downloads. If there are a large number of CI jobs nodes may occasionally be rejected due to the Kubernetes suite not pulling down from the registry cache.


james-d-elliott (approved):
LGTM

PR #1684, opened 2021-02-02
test(handlers): add health check to coverage build

nightah:
This change ensures that the integration test container also utilises the health check script to ensure coverage for this area of code.


james-d-elliott (approved):
LGTM

PR #1708, opened 2021-02-08
Fix invalid path in Getting Started

Duponin:
I wanted to try `Authelia` but getting started is currently invalid so quick and simple commit to fix it.

I tried to find which commit made the `Getting Started` invalid but can't find it in the history:
``` shell
$ git log --full-history -1 -- authelia/compose/local/setup.sh
```

Duponin

PS: edit at line 34 looks to be a Github bug, it only appears when I do the Pull Request (change is done in web editor)

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -13,7 +13,7 @@ nav_order: 2
 These commands are intended to be run sequentially:
 
 - `git clone https://github.com/authelia/authelia.git`
-- `cd authelia/compose/local`
+- `cd compose/local`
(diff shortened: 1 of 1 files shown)


nightah:
I'm not entirely sure how this resulted in an invalid path for you, cloning a repository does not also change your directory into said clone.

If we change the instructions per your modifications this actually results in an invalid path and failing to change directory.

Duponin:
Oh it's supposed to be the cloned project?
I thought it was in the project directory (the `cd` was implied in my head).
Then don't mind my PR, I'll close it.
My bad haha.

PR #1584, opened 2021-01-05
[FEATURE] Add theme support

nightah:
This change allows users to select a theme for Authelia on start-up.

The default will continue to be the existing theme which is known as `light`.
 ~~Three~~ Two new options are now also provided:
* `dark`
* `grey`
* ~~`custom`~~

~~The `custom` theme allows users to specify a primary and secondary hex color code to be utilised to style the portal.~~

Closes #113.

Co-authored-by: BankaiNoJutsu <lbegert@gmail.com>

PR #1723, opened 2021-02-13
fix(web): fix otp input box with themes

### LINKED ISSUE #1720: No indication when typing TOTP on OTP page
Using v4.26.1

Looks like this was introduced in v4.26.0

On v2.25.2
image

After:
image

Is that normal? I can see it being highlighted as you type, but faintly (especially if using dark/grey theme).
Any change to show at least a * dot ?

Thanks!

nightah:
#1584 introduced a regression due to the way layouts are addressed with [CSS normalization](https://material-ui.com/components/css-baseline/#layout).

This PR ensures that the OTP input box will remain a `content-box` to avoid any text display and box sizing issues caused by padding.

The OTP input has also been changed to only accept numbers.

Fixes #1720.


james-d-elliott (approved):
LGTM!

PR #1733, opened 2021-02-19
feat: go:embed static assets

nightah:
Go 1.16 introduced the ability to embed files within a generated binary directly with the go tool chain. This simplifies our dependencies and the significantly improves the development workflow for future developers.

**Key points to note:**

Due to the inability to embed files that do not reside within the local package we need to duplicate our `config.template.yml` within `internal/configuration`.

To avoid issues with the development workflow empty mock files have been included within `internal/server/public_html`. These are substituted with the respective generated files during the CI/CD and build workflows.


james-d-elliott (approved):
LGTM

PR #1762, opened 2021-02-24
fix(configuration): lower argon2id default memory requirements

nightah:
The current default hashing value of 1024MB (1GB) is far too aggressive to cover all use cases.
Reducing this number and encouraging users to to read the documentation and tune will result in less issues and a better user experience.


james-d-elliott (approved):
LGTM

PR #1768, opened 2021-02-26
feat(session): add redis sentinel provider

james-d-elliott:
This adds an implementation that should work with redis sentinel in a fail over configuration.

### DIFF
--- a/config.template.yml
+++ b/config.template.yml
@@ -109,7 +109,7 @@ authentication_backend:
 
     # The url to the ldap server. Scheme can be ldap or ldaps in the format (port optional) <scheme>://<address>[:<port>].
     url: ldap://127.0.0.1
-    
+
     # Use StartTLS with the LDAP connection.
     start_tls: false
 
@@ -125,7 +125,7 @@ authentication_backend:
 
     # The base dn for every entries.
     base_dn: dc=example,dc=com
-    
+
     # The attribute holding the username of the user. This attribute is used to populate
     # the username in the session information. It was introduced due to #561 to handle case
     # insensitive search queries.
@@ -138,7 +138,7 @@ authentication_backend:
     # them, we instead advise to use the attributes mentioned above (sAMAccountName and uid) to follow
     # https://www.ietf.org/rfc/rfc2307.txt.
     # username_attribute: uid
-    
+
     # An additional dn to define the scope to all users.
     additional_users_dn: ou=users
 
@@ -159,7 +159,7 @@ authentication_backend:
 
     # An additional dn to define the scope of groups.
     additional_groups_dn: ou=groups
-    
+
     # The groups filter used in search queries to find the groups of the user.
     # - {input} is a placeholder replaced by what the user inputs in the login form.
     # - {username} is a placeholder replace by the username stored in LDAP (based on `username_attribute`).
@@ -270,8 +270,8 @@ access_control:
         - 10.0.0.1
 
     - domain:
-      - secure.example.com
-      - private.example.com
+        - secure.example.com
+        - private.example.com
       policy: two_factor
 
     - domain: singlefactor.example.com
@@ -326,7 +326,7 @@ session:
   # The name of the session cookie. (default: authelia_session).
   name: authelia_session
 
-  # The secret to encrypt the session data. This is only used with Redis.
+  # The secret to encrypt the session data. This is only used with Redis / Redis Sentinel.
   # Secret can also be set using a secret: https://docs.authelia.com/configuration/secrets.html
   secret: insecure_session_secret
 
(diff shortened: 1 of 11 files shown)


config.template.yml
```diff
@@ -348,19 +355,64 @@ session:
...
+    # This provides specific options to Redis Sentinel, sentinel_name must be defined (Master Name).
+    # high_availability:
+      # Sentinel Name / Master Name
+      # sentinel_name: mysentinel
+
+      # Specific password for Redis Sentinel. The node username and password is configured above.
+      # sentinel_password: sentinel_specific_pass
+
+      # The additional nodes to pre-seed the redis provider with (for sentinel).
+      # The first node in this list will always be the one in session.redis.host/port if specified.
```
nightah, line 403:
```suggestion
      # The first node in this list should always be the one in session.redis.host/port if specified.
```

docs/configuration/session.md
```diff
@@ -48,11 +48,59 @@ session:
...
+    # This provides specific options to Redis Sentinel, sentinel_name must be defined (Master Name).
+    high_availability:
+      # Sentinel Name / Master Name
+      sentinel_name: mysentinel
+
+      # Specific password for Redis Sentinel. The node username and password is configured above.
+      sentinel_password: sentinel_specific_pass
+
+      # The additional nodes to pre-seed the redis provider with (for sentinel).
+      # The first node in this list will always be the one in session.redis.host/port if specified.
```
nightah, line 92:
```suggestion
      # The first node in this list should always be the one in session.redis.host/port if specified. 
```

docs/configuration/session.md
```diff
@@ -74,4 +122,9 @@ host: "[fd00:1111:2222:3333::1]"
 
 ## Loading a password from a secret instead of inside the configuration
 
-Password can also be defined using a [secret](../secrets.md).
\ No newline at end of file
+Password can also be defined using a [secret](../secrets.md).
+
+## Redis Sentinel
+
+When using Redis Sentinel, the host specified in the main redis section is added (it will be the first node) to the
```
nightah, line 129: I'm actually confused by this statement, does the user need to ensure that the first node is what appears in the redis configuration or is this done automatically?
I think the wording needs to be more clear and our specific instructions in this case.

internal/configuration/config.template.yml
```diff
@@ -348,19 +355,64 @@ session:
...
+    # This provides specific options to Redis Sentinel, sentinel_name must be defined (Master Name).
+    # high_availability:
+      # Sentinel Name / Master Name
+      # sentinel_name: mysentinel
+
+      # Specific password for Redis Sentinel. The node username and password is configured above.
+      # sentinel_password: sentinel_specific_pass
+
+      # The additional nodes to pre-seed the redis provider with (for sentinel).
+      # The first node in this list will always be the one in session.redis.host/port if specified.
```
nightah, line 403:
```suggestion
      # The first node in this list should always be the one in session.redis.host/port if specified.
```

internal/configuration/validator/const.go
```diff
@@ -169,6 +185,11 @@ var specificErrorKeys = map[string]string{
 	"authentication_backend.file.hashing.parallelism":          "config key incorrect: authentication_backend.file.hashing should be authentication_backend.file.password",
 }
 
+const errFmtSessionSecretRedisProvider = "The session secret is not set but must be set when using the %s session provider"
+const errFmtSessionRedisPortRange = "The port must be between 1 and 65535 the %s session provider"
```
nightah, line 189:
```suggestion
const errFmtSessionRedisPortRange = "The port must be between 1 and 65535 for the %s session provider"
```

This reads a little odd without the "for".

internal/suites/example/compose/redis/docker-compose.yml
```diff
@@ -1,8 +1,9 @@
 version: '3'
 services:
   redis:
```
nightah, line 3: I do not believe this file will be required anymore if you're are using the redis-sentiel compose stack.

nightah (changes requested):
Only a couple small nits, I have to admit I'd never heard of redis sentinel before this PR.

Did you find during your research that Redis Sentinel is generally favoured over their Redis cluster?

james-d-elliott:
> Did you find during your research that Redis Sentinel is generally favoured over their Redis cluster?

The main difference in my mind from my research is Redis Cluster is objectively better with large amounts of data where you need to split it from a single memory blob; and Redis Sentinel is objectively better with small amounts of data where RAM will never become a concern (i.e. our session data is tiny) but you want more redundancy than a single node. 

I have actually used redis as a replacement for a traditional relational database before when designing a chat bot with user settings, without any fancy JSON etc. I did mix it with a SQL database for logging, but I feel like if a company needs a fast DB and they're thoughtful with their design Redis can actually handle a lot of traditional database tasks, albeit with extra setup work.

nightah:
Oh also forgot to mention the only other thing I would expect to see is actually utilising a test case in the HighAvailability suite to bring down redis nodes and ensuring that the session is still preserved.

docs/configuration/session.md
james-d-elliott: It's technically optional. I ensure either the host is specified or the nodes section has a single entry. I'll reword it.

internal/suites/example/compose/redis/docker-compose.yml
james-d-elliott: I have merged them all into redis.

james-d-elliott:
> 
> 
> Oh also forgot to mention the only other thing I would expect to see is actually utilising a test case in the HighAvailability suite to bring down redis nodes and ensuring that the session is still preserved.

Yep! It was on my mind. I just wanted to make sure the original tests work first to get a baseline of everything working with Redis Sentinel. That way I'm not chasing my tail.

internal/suites/example/compose/redis/templates/master.conf
```diff
@@ -0,0 +1,2013 @@
+# Redis configuration file example.
```
james-d-elliott, line 1: Add note about source of this file.

internal/suites/example/compose/redis/templates/sentinel.conf
```diff
@@ -0,0 +1,342 @@
+# Example sentinel.conf
```
james-d-elliott, line 1: Add a note about source of this file.

internal/suites/example/compose/redis/templates/slave.conf
```diff
@@ -0,0 +1,2013 @@
+# Redis configuration file example.
```
james-d-elliott, line 1: Add a note about source of this file.

nightah (approved):
LGTM.

PR #1797, opened 2021-03-10
ci: restore dependabot rules

james-d-elliott:
Restores the dependabot rules in buildkite for the purpose of security fixes which are still handled by dependabot.


nightah (approved):
LGTM.

PR #1806, opened 2021-03-11
fix(server): send 404 on missing api endpoints instead of 200

### LINKED ISSUE #1520: Returning login page (200) for all paths is dangerous

james-d-elliott:
Returns a 404 instead of 200 on bad API endpoints.
Co-authored-by: Clément Michaud <clement.michaud34@gmail.com>

Fixes #1520
Closes #1534

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -53,7 +53,7 @@ func StartServer(configuration schema.Configuration, providers middlewares.Provi
 	}
 
 	r.GET("/static/{filepath:*}", embeddedFS)
-	r.GET("/api/{filepath:*}", embeddedFS)
+	r.ANY("/api/{filepath:*}", embeddedFS)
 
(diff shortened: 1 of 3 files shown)


nightah (approved):
LGTM.

PR #1809, opened 2021-03-12
refactor(session): use github.com/fasthttp/session/v2 instead of github.com/authelia/session/v2

james-d-elliott:
Reverts to the upstream library instead of our maintenance fork.


nightah (approved):
LGTM.

PR #1819, opened 2021-03-12
docs(authorization): document changed resources behavior

### LINKED ISSUE #1817: Resources ACL settings no longer working per documentation in 4.27.x
It seems that resource (path) based ACLs are not entirely working in 4.27.x any longer. I have verified that such configurations work with the `4.26.2` docker image tag, but not in either of `4.27.0` or `4.27.1`

An example ACL configuration which works in 4.26.2 but not later versions:

```
    - domain: "myhost.example.com"
      policy: bypass
      resources:
        - '^/api/v[0-9]+$'
```

This results in 302 redirects to authelia in current versions, vs. a proper bypass in `4.26.2` and below.

However, changing the resource value to the following does work:

`'^/api/v[0-9]+.*$'` (Adding a greedy `.*` at the end)

An HTTP request example for the application in question looks similar to the following:

```
GET /api/v2?apikey=decafc0ffeebadecafc0ffeebadddecafc0f&other_param=yes&one_more_param=action
```

I wonder if this was introduced in 4dce8f94962d3bd0099bbb202f76696a551d099b for the ACL preload work (#1640) ? Is it possible that URL query parameters are now also being evaluated in resources, where they were not before? If so, that may just require a documentation update.

james-d-elliott:
I missed documenting this change, but prior to 4.27.0 the query param was never considered when matching resources. But that's no longer the case.

Fixes #1817

### DIFF
--- a/docs/configuration/access-control.md
+++ b/docs/configuration/access-control.md
@@ -102,8 +102,9 @@ A rule can define multiple regular expressions for matching the path of the reso
 similar to the list of domains. If any one of them matches, the resource criteria of
 the rule matches.
 
-Note that regular expressions can be used to match a given path. However, they do not match
-the query parameters in the URL, only the path.
+Note that regular expressions can be used to match a given path. However prior to 4.27.0, they do not match
+the query parameters in the URL, only the path. If you're upgrading to 4.27.0+ you may have to alter some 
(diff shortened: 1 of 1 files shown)


nightah (approved):
LGTM.

PR #1835, opened 2021-03-16
docs: fix haproxy examples for /api/verify?auth=basic

nightah:
The previous examples did not appropriately pass through the `WWW-Authenticate` header and 401 when the user was unauthenticated, therefore not resulting in a basic auth login prompt.


aHVzY2g (approved):

james-d-elliott (approved):
LGTM

PR #1863, opened 2021-03-28
fix: exit setup.sh if missing pre-requisites

Duckwhale:
Running the script with missing prerequisites
doesn't fail, but instead warns the user
that return statements aren't allowed here.

---

I didn't have ``docker-compose`` installed, but the script continued to run instead of stopping immediately. Seems like an oversight?

### DIFF
--- a/compose/local/setup.sh
+++ b/compose/local/setup.sh
@@ -16,12 +16,12 @@ echo "Checking for pre-requisites"
 
 if [[ ! -x "$(command -v docker)" ]]; then
   echo "You must install Docker on your machine";
-  exit
+  exit 1
 fi
 
 if [[ ! -x "$(command -v docker-compose)" ]]; then
   echo "You must install Docker Compose on your machine";
-  exit
+  exit 1
 fi
 
 echo "Pulling Authelia docker image for setup"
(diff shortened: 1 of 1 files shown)


compose/local/setup.sh
```diff
@@ -16,12 +16,12 @@ echo "Checking for pre-requisites"
 
 if [[ ! -x "$(command -v docker)" ]]; then
   echo "You must install Docker on your machine";
-  return
+  exit
```
nightah, line 19:
```suggestion
  exit 1
```

compose/local/setup.sh
```diff
@@ -16,12 +16,12 @@ echo "Checking for pre-requisites"
...
 fi
 
 if [[ ! -x "$(command -v docker-compose)" ]]; then
   echo "You must install Docker Compose on your machine";
-  return
+  exit
```
nightah, line 24:
```suggestion
  exit 1
```

nightah (changes requested):
This was purely an oversight, just a small nit and happy to accept.
Thanks for the PR @Duckwhale!

nightah (approved):
LGTM!

PR #1871, opened 2021-03-30
fix: remove health checks on compose examples

### LINKED ISSUE #1864: Following the "Getting started" documentation doesn't lead to the expected result
I've never used Authelia before and was simply trying to check out the demo. The setup script seems to have run successfully and the containers are ``up``. However, accessing either of the three subdomains described in the "Getting started" guide simply results in a ``404: Not Found`` . This is after the faux SSL warning from my browser, so it appears the host/port is indeed correct.

Alas, I'm not sure what should have happened instead... but... surely this isn't the desired result? As a dumb user who has no idea how Authelia works and wants to give it a try, I would expect to see something (anything, really) after the script tells me to check out the redirected``example.com`` URLs.

PS: This issue isn't just to ask what might've gone wrong, but also a suggestion to improve the documentation:

* If there is supposed to be some visible UI, add a paragraph with a screenshot to tell people what they should be seeing
* If there's any other way to verify the setup went through successfully, e.g., by checking the API endpoint is available or similar, that should also be noted
* If the expected outcome is "invisible", this is not ideal, but should also be noted
* If additional steps need to be taken before visible results manifest, all of them should be listed

Rationale:

The tutorial is the most critical part of any project; it'll be read by virtually every new user and most of them will be completely clueless... just like me :D It should work immediately and instructions should be provided that lead to the desired outcome, as well provisions included for any part where things could go wrong.

If it doesn't, lots of people will leave and find another project that "just works" (TM), never to be seen again (and obviously never supporting this one in any way), or they might open lazy support tickets, putting additional strain on the development team.

nightah:
Traefik does not add routes for containers via the Docker provider if the health check does not return healthy, this causes inadvertent user experience issues when attempting the pre-made compose examples.

This change removes the health checks for said examples and also ensures that Traefik logs are written to stdout so a user can view them within the Docker container logs.

Fixes #1864.

### DIFF
--- a/compose/lite/docker-compose.yml
+++ b/compose/lite/docker-compose.yml
@@ -24,6 +24,8 @@ services:
     expose:
       - 9091
     restart: unless-stopped
+    healthcheck:
+      disable: true
     environment:
       - TZ=Australia/Melbourne
 
@@ -74,7 +76,6 @@ services:
       - '--certificatesResolvers.letsencrypt.acme.httpChallenge.entryPoint=http'
       - '--log=true'
       - '--log.level=DEBUG'
-      - '--log.filepath=/var/log/traefik.log'
 
   secure:
     image: containous/whoami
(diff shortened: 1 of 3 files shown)


james-d-elliott (approved):
LGTM

PR #1883, opened 2021-04-01
fix: update authelia.service dependencies

Veganizer:
Authelia is throwing error if it configures to starts before mysql server.
If Authelia starts after multi-user.target it will start after mysql server, because the mysql server is starting after network.target.


nightah:
The intent of the the base systemd unit is actually for the most minimalistic version of Authelia.

Ideally a user should be modifying the dependencies to suit how they plan to deploy it, that's partly why @james-d-elliott didn't make a PR to fix this.

Having said that, this change as a default seems to be a reasonable compromise. I would still recommend that users tune their dependencies so it's specific to their deployment.

nightah (approved):
LGTM!

PR #1898, opened 2021-04-09
fix(validator): misleading warning for empty acl domains

james-d-elliott:
This fixes misleading errors for ACL rules with an empty list of domains. This also enables admins to have a default policy with zero ACL rules as long as the default policy is not deny or bypass. It also adds a rule number to all ACL rule related log messages which is the position in the YAML list plus 1. Lastly it adds comprehensive per rule HIT/MISS logging when Authelia trace logging is enabled. This trace logging includes the rule number.


internal/authorization/authorizer.go
```diff
@@ -35,17 +48,44 @@ func (p *Authorizer) IsSecondFactorEnabled() bool {
...
-			return rule.Policy
+			return &rule.Policy
 		}
 	}
 
-	logger.Tracef("No matching rule for subject %s and url %s... Applying default policy.", subject.String(), object.String())
+	return nil
+}
+
+func getRequiredLevelTraceFunc(log *logrus.Logger, rules []*AccessControlRule, subject Subject, object Object) *Level {
```
clems4ever, line 79: I think it's better to factorize the code instead of duplicating it. Maybe we can have a bool in the args of this internal function and call the function with true or false depending on the case. Otherwise,  by duplicating the code, we will start diverging too much and we can already see that some parts of that function are not covered by tests yet by codecov so that could be an issue in the long run.

internal/authorization/access_control_rule.go
```diff
@@ -32,6 +33,7 @@ func NewAccessControlRule(rule schema.ACLRule, networksMap map[string][]*net.IPN
 
 // AccessControlRule controls and represents an ACL internally.
 type AccessControlRule struct {
+	ID        int
```
clems4ever, line 36: I'd call it `Position` instead of ID because my fear is that one day we put references to ACL in the DB and people might think it's a primary key whereas it's not. If we decide to agree on Position, then we should propagate through out all functions and update argument names in the function signatures.

internal/configuration/validator/access_control.go
```diff
@@ -68,61 +68,76 @@ func ValidateAccessControl(configuration schema.AccessControlConfiguration, vali
 
 // ValidateRules validates an ACL Rule configuration.
 func ValidateRules(configuration schema.AccessControlConfiguration, validator *schema.StructValidator) {
-	for _, r := range configuration.Rules {
-		if len(r.Domains) == 0 {
-			validator.Push(fmt.Errorf("No access control rules have been defined"))
+	if len(configuration.Rules) == 0 {
+		defaultPolicy := strings.ToLower(configuration.DefaultPolicy)
```
clems4ever, line 72: I'd rather not open the door to case insensitive enums (like aLLoW or deNy or whatever) in the config but rather throw an error if it does not exactly match. It should not be our problem if people don't care about case sensitivity imo. Let's keep control on that. What do you think?

internal/configuration/validator/access_control.go
```diff
@@ -68,61 +68,76 @@ func ValidateAccessControl(configuration schema.AccessControlConfiguration, vali
...
-		if r.Policy == bypassPolicy && len(r.Subjects) != 0 {
-			validator.Push(fmt.Errorf(errAccessControlInvalidPolicyWithSubjects, r.Domains, r.Subjects))
+		if rule.Policy == bypassPolicy && len(rule.Subjects) != 0 {
+			validator.Push(fmt.Errorf(errAccessControlInvalidPolicyWithSubjects, ruleID, rule.Domains, rule.Subjects))
 		}
 	}
 }
 
-func validateNetworks(r schema.ACLRule, configuration schema.AccessControlConfiguration, validator *schema.StructValidator) {
+func validateNetworks(id int, r schema.ACLRule, configuration schema.AccessControlConfiguration, validator *schema.StructValidator) {
```
clems4ever, line 109:
```suggestion
func validateNetworks(rulePosition int, r schema.ACLRule, configuration schema.AccessControlConfiguration, validator *schema.StructValidator) {

File: internal/configuration/validator/access_control.go [hunk 5]
```diff
@@ -68,61 +68,76 @@ func ValidateAccessControl(configuration schema.AccessControlConfiguration, vali
...
 			if !IsNetworkGroupValid(configuration, network) {
-				validator.Push(fmt.Errorf("Network %s for domain: %s is not a valid network or network group", r.Networks, r.Domains))
+				validator.Push(fmt.Errorf("Network %s for rule #%d domain: %s is not a valid network or network group", r.Networks, id, r.Domains))
 			}
 		}
 	}
 }
 
-func validateResources(r schema.ACLRule, validator *schema.StructValidator) {
+func validateResources(id int, r schema.ACLRule, validator *schema.StructValidator) {
```
clems4ever, line 119:
```suggestion
func validateResources(rulePosition int, r schema.ACLRule, validator *schema.StructValidator) {

File: internal/configuration/validator/const.go [hunk 6]
```diff
@@ -30,7 +32,7 @@ const (
 	testTLSCert       = "/tmp/cert.pem"
 	testTLSKey        = "/tmp/key.pem"
 
-	errAccessControlInvalidPolicyWithSubjects = "Policy [bypass] for domain %s with subjects %s is invalid. It is " +
+	errAccessControlInvalidPolicyWithSubjects = "Policy [bypass] for rule %d domain %s with subjects %s is invalid. It is " +
```
clems4ever, line 35:
```suggestion
errAccessControlInvalidPolicyWithSubjects = "Policy [bypass] for rule #%d domain %s with subjects %s is invalid. It is " +

clems4ever (dismissed):

internal/configuration/validator/access_control.go
james-d-elliott: Sounds fine to me.

internal/authorization/authorizer.go
james-d-elliott: I'm actually really glad you got to see this before it was merged for this reason. I was really unsure if what I was doing was a good idea.

The intent behind it was purely optimization since this section of code is the most hit section in day to day activities. I'm happy to defer to you regarding if you think it's necessary. My gut was leaning towards it's not since it's a very low cost check.

internal/configuration/validator/const.go
james-d-elliott: Nice spot!

internal/authorization/authorizer.go
clems4ever: Glad that was helpful.
If I had to choose between two rather similar functions with two similar set of tests or one factorized implementation with only one extended set of tests, I'd go for the latter but it's only my opinion. I try to factorize when it makes it avoids duplicating the tests as well.

james-d-elliott: Yeah that's what I applied in one of the later commits. I'm fine with it after bench-marking it. A 100 item ACL only takes about 12,000ns to parse now on a modern system.

nightah (approved):
LGTM.

PR #1904, opened 2021-04-11
refactor(authentication): add trace logs for the user/group baseDN

james-d-elliott:
This logs the baseDN for user and group searching on startup. Additionally refactors the location of a few log messages, and exposes the logger to the provider to reduce calls to logging.Logger().


nightah (approved):
LGTM.

PR #1907, opened 2021-04-12
docs: Add a page listing supported proxies.

clems4ever:
I'm adding this page because after doing a quick review of the doc, it was not easy to find out this basic information.


james-d-elliott (approved):
LGTM

PR #1912, opened 2021-04-14
fix: removed deprecated smtp/ldap options

james-d-elliott:
This removes the deprecated options from 4.25. This includes the LDAP filters which allow {0} or {1} placeholders. The new aliases are documented. Additionally it refactors the keys validator to use uniform messages for most replaced keys.

Fix because they should have been removed in 4.28.0, but will be removed in 4.28.1 instead. Also it's noteworthy that the coverage has decreased, it appears it's because there were a lot of previously covered lines completely removed mostly. I couldn't see any other negative hits though if you see one let me know.


nightah (changes requested):
Just a couple small changes but otherwise LGTM.

nightah (approved):
LGTM!

PR #1919, opened 2021-04-14
ci: include version in the name of tar.gz artifacts

### LINKED ISSUE #1918: Please add version to the authelia-public_html tarball
It is a common practice for tarballs to have a version in their file name, not only in the URL name.

Thank you,
Yuri

clems4ever:
Fix #1918


james-d-elliott (approved):
LGTM

PR #1921, opened 2021-04-15
fix(suites): ensure k8s suite utilises the registry cache

nightah:
The annotation was previously pointing to the incorrect registry so the registrycache was not being utilised.


james-d-elliott (approved):
LGTM

PR #1943, opened 2021-04-24
feat: bloc floc

james-d-elliott:
This change adds a Response Headers Middleware and adds a header which should BLoC FLoC when users visit Authelia instances.

### DIFF
--- a/internal/middlewares/headers.go
+++ b/internal/middlewares/headers.go
@@ -1,0 +1,11 @@
+package middlewares
+
+import "github.com/valyala/fasthttp"
+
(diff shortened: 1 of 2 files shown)


james-d-elliott:
It has been decided we wont be implementing this due to misleading information being spread about FLoC. Basically unless the javascript makes a specific Browser API call in the Chrome browser, FLoC is never used. As we do not call this API, it does not track Authelia. Read more here: https://seirdy.one/2021/04/16/permissions-policy-floc-misinfo.html

PR #1973, opened 2021-05-04
refactor: use fatal log message on oidc init error

james-d-elliott:
Instead of using panic() when an unknown error was detected on OP initialization, use logger.Fatalf. This provides a more useful output to the user.


nightah (approved):
LGTM.

PR #1976, opened 2021-05-05
fix(cmd): retry clean tag logic for dockerhub

nightah:
This change will ensure that if the curl command for the cleaning of Docker tags on DockerHub fails it will be reattempted up to 2 more times (total of 3) with a 10 second sleep between each attempt.

The clean tag logic itself within curl attempts to execute the http request upto 3 times so this will ensure a maximum of 9 attempts.


james-d-elliott (approved):
LGTM

PR #1982, opened 2021-05-06
refactor: forcibly use latest go-sqlite3

james-d-elliott:
This is necessary to force the latest version of go-sqlite3 due to an issue with the release lifecycle upstream and due to deep dependencies.


nightah (approved):
LGTM.

PR #1983, opened 2021-05-06
fix(handlers): include the jwk key id in the jwt for validation

### LINKED ISSUE #1979: OIDC Invalid Signature while verifying ID Token
### Description

I set up a test client to try OIDC on https://openidconnect.net/

The workflow works fine but at step 3: Verify User Token the validation of the JWT's signature fails.

From https://openidconnect.net/:
This token is cryptographically signed with the RS256 algorithim. We'll use the public key of the OpenID Connect server to validate it. In order to do that, we'll fetch the public key from https://authelia/api/oidc/jwks, which is found in the discovery document or configuration menu options.

### Expected Behaviour

Successful validation of ID Token

### Reproduction Steps

1. Request an access code from client that is configured in config
2. Turn that access code into an access token, by having the client make a request to your token endpoint
3. Verify the ID Token

### Additional Information

At https://jwt.io/ the id token has also an invalid signature

james-d-elliott:
This is so the sig key used to sign the JWT can be verified using the JWKS endpoint.

Fixes #1979

### DIFF
--- a/internal/handlers/oidc.go
+++ b/internal/handlers/oidc.go
@@ -100,7 +100,10 @@ func newDefaultOIDCSession(ctx *middlewares.AutheliaCtx) (session *openid.Defaul
 			Extra:       make(map[string]interface{}),
 		},
 		Headers: &jwt.Headers{
-			Extra: make(map[string]interface{}),
+			Extra: map[string]interface{}{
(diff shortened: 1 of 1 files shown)


nightah (approved):
LGTM.

PR #1991, opened 2021-05-08
feat(oidc): add additional config options and refactoring

james-d-elliott:
This is a  experimental branch, which implements the following changes;

- Refactors several areas out of the handlers package and moves them into the oidc package
- Implements a key manager that handles JWKs and RSA keys which will allow key rotation in the future
- Key ID is now the SHA256 hexidecimal hash of the RSA pubkey
- Implements and adds conf opts for:
  - client debug mode setting
  - response modes
  - lifetime for: id_tokens, access_tokens, refresh_tokens, authorize_codes
- Adds a lot of unit tests
- Adds checks to validator for scopes, response_types, etc


internal/configuration/validator/configuration_test.go
```diff
@@ -158,14 +158,24 @@
 	assert.EqualError(t, validator.Errors()[0], "Provide a JWT secret using \"jwt_secret\" key")
 }
 
+func TestShouldRaiseErrorWithBadExternalURL(t *testing.T) {
+	validator := schema.NewStructValidator()
+	config := newDefaultConfig()
+	config.ExternalURL = "abc"
```
autheliabot, line 164: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>string `abc` has 2 occurrences, make it a constant (goconst)

internal/handlers/oidc.go
```diff
@@ -23,87 +19,20 @@
...
-	oidcSession.Claims.Extra = extra
-	oidcSession.Claims.Subject = userSession.Username
-	oidcSession.Claims.Audience = ar.GetGrantedAudience()
-
-	return oidcSession, err
 }
 
-func newDefaultOIDCSession(ctx *middlewares.AutheliaCtx) (session *openid.DefaultSession, err error) {
-	issuer, err := ctx.ForwardedProtoHost()
+type OpenIDSession struct {
```
autheliabot, line 33: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>exported type `OpenIDSession` should have comment or be unexported (golint)

internal/handlers/handler_oidc_wellknown.go
```diff
@@ -7,63 +7,58 @@
...
+				"code token id_token",
+				"none",
+			},
+		}
 	}
 
 	ctx.SetContentType("application/json")
-
-	if err := json.NewEncoder(ctx).Encode(configuration); err != nil {
+	if err := json.NewEncoder(ctx).Encode(*ctx.Providers.OpenIDConnect.WellKnown); err != nil {
```
autheliabot, line 61: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>if statements should only be cuddled with assignments (wsl)

internal/oidc/keys_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	webKey, err := manager.GetActiveWebKey()
+	assert.NoError(t, err)
+	assert.NotNil(t, webKey)
+
+	keySet := manager.GetKeySet()
+	assert.NotNil(t, keySet)
+
+	assert.Equal(t, kid, manager.GetActiveKeyID())
+
```
autheliabot, line 53: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>unnecessary trailing newline (whitespace)

internal/oidc/keys_test.go
```diff
@@ -0,0 +1,54 @@
...
+}
```
autheliabot, line 54: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>block should not end with a whitespace (or comment) (wsl)

internal/oidc/keys_test.go
```diff
@@ -0,0 +1,54 @@
...
+	key, wk, err := manager.AddActiveKeyData(exampleIssuerPrivateKey)
+	require.NoError(t, err)
+	require.NotNil(t, key)
+	require.NotNil(t, wk)
+	require.NotNil(t, manager.strategy)
+	require.NotNil(t, manager.Strategy())
+
+	thumbprint, err := wk.Thumbprint(crypto.SHA256)
+	assert.NoError(t, err)
+	kid := fmt.Sprintf("%x", thumbprint)
```
autheliabot, line 26: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>assignments should only be cuddled with other assignments (wsl)

config.template.yml
```diff
@@ -575,6 +580,12 @@ notifier:
     #   --- KEY START
     #   --- KEY END
 
+    ## The lifespans configure the expiration for these token types.
+    # access_token_lifespan: 1h
+    # authorize_code_lifespan: 1h
```
clems4ever, line 585: I think this should be set to something like 30 seconds by default and in comments . Authorization codes do not need more in my opinion.

config.template.yml
```diff
@@ -575,6 +580,12 @@ notifier:
...
+    # id_token_lifespan: 1h
+    # refresh_token_lifespan: 720h
```
clems4ever, line 587: We should probably advertise something smaller: https://docs.apigee.com/api-platform/antipatterns/oauth-long-expiration

james-d-elliott: Yep I agree, currently the defaults are exactly what we have configured in 4.29.1 (by not configuring it). As I unwrap more of fosite I'm discovering more of these things.

config.template.yml
```diff
@@ -611,4 +625,10 @@ notifier:
         ## It's not recommended to define this unless you know what you're doing.
         # response_types:
         # - code
+
+        ## Response Modes configures which response modes this client supoprts.
```
clems4ever, line 624:
```suggestion
        ## Response Modes configures which response modes this client supports.

File: docs/configuration/identity-providers/oidc.md [hunk 10]
```diff
@@ -136,6 +136,26 @@ The private key in DER base64 encoded PEM format used to encrypt the [OpenID Con
 
 Can also be defined using a [secret](../secrets.md) which is the recommended for containerized deployments.
 
+### access_token_lifespan
+
+The maximum lifetime of an access token.
+
+### authorize_code_lifespan
+
+The maximum lifetime of an authorize code.
```
clems4ever, line 145: We should probably add some recommendations for each of them to help people.

internal/configuration/validator/identity_providers.go
```diff
@@ -82,17 +96,96 @@ func validateOIDCClients(configuration *schema.OpenIDConnectConfiguration, valid
...
+		configuration.Clients[c].Scopes = schema.DefaultOpenIDConnectClientConfiguration.Scopes
+		return
+	}
+
+	if !utils.IsStringInSlice("openid", configuration.Clients[c].Scopes) {
+		configuration.Clients[c].Scopes = append(configuration.Clients[c].Scopes, "openid")
+	}
+
+	for _, scope := range configuration.Clients[c].Scopes {
+		if !utils.IsStringInSlice(scope, validScopes) {
```
clems4ever, line 106: I think the concept of valid scope is limited to ID tokens but does not apply to access tokens. To me it should be ok to request a custom scope from an app and attach it to the access token. Maybe there is something to consider here regarding this array.

internal/configuration/validator/identity_providers.go
```diff
@@ -82,17 +96,96 @@ func validateOIDCClients(configuration *schema.OpenIDConnectConfiguration, valid
...
+	for _, responseMode := range configuration.Clients[c].ResponseModes {
+		if !utils.IsStringInSlice(responseMode, validOIDCResponseModes) {
+			validator.Push(fmt.Errorf(
+				errFmtOIDCServerClientInvalidResponseMode,
+				configuration.Clients[c].ID, responseMode, strings.Join(validOIDCResponseModes, "', '")))
+		}
+	}
+}
+
+func validateOIDCClientRequestURIs(client schema.OpenIDConnectClientConfiguration, validator *schema.StructValidator) {
```
clems4ever, line 159: As discussed, I think this should be added in a dedicated PR.

internal/handlers/handler_firstfactor.go
```diff
@@ -174,6 +174,7 @@ func FirstFactorPost(msInitialDelay time.Duration, delayEnabled bool) middleware
 		userSession.Groups = userDetails.Groups
 		userSession.Emails = userDetails.Emails
 		userSession.AuthenticationLevel = authentication.OneFactor
+		userSession.Authenticated = time.Now().Unix()
```
clems4ever, line 177: On full 2FA environments the user would be considered authenticated only after 2FA stage.

internal/handlers/handler_oidc_userinfo.go
```diff
@@ -0,0 +1,79 @@
...
+	fositejwt "github.com/ory/fosite/token/jwt"
+
+	"github.com/authelia/authelia/internal/middlewares"
+	"github.com/authelia/authelia/internal/oidc"
+)
+
+func oidcUserinfo(ctx *middlewares.AutheliaCtx, rw http.ResponseWriter, req *http.Request) {
+	session := newOpenIDSession("")
+
+	tokenType, ar, err := ctx.Providers.OpenIDConnect.Fosite.IntrospectToken(req.Context(), fosite.AccessTokenFromRequest(req), fosite.AccessToken, session)
```
clems4ever, line 21: Are you sure we are not mixing token introspection and userinfo endpoints?

internal/configuration/validator/identity_providers.go
clems4ever: I got my answer in the Discovery RFC: `Servers MAY choose not to advertise some supported scope values even when this parameter is used, although those defined in [OpenID.Core] SHOULD be listed, if supported.`
So this support list only relates to OIDC not OAuth2.

internal/oidc/client.go
```diff
@@ -31,6 +68,21 @@ func (c InternalClient) GetID() string {
 	return c.ID
 }
 
+// GetConsentRequestBody returns the proper consent request body for this session.OIDCWorkflowSession.
+func (c InternalClient) GetConsentRequestBody(session *session.OIDCWorkflowSession) ConsentGetResponseBody {
```
clems4ever, line 72: Request and response are mixed up here.

internal/oidc/client.go
```diff
@@ -73,3 +125,64 @@ func (c InternalClient) IsPublic() bool {
...
+//
+// Implements part of fosite.OpenIDConnectClient.
+func (c InternalClient) GetResponseModes() []fosite.ResponseModeType {
+	return c.ResponseModes
+}
+
+// GetRequestURIs returns the valid request URIs for this client.
+//
+// Implements part of fosite.OpenIDConnectClient.
+func (c InternalClient) GetRequestURIs() []string {
```
clems4ever, line 139: Should be in a dedicated PR as discussed.

internal/oidc/client.go
```diff
@@ -73,3 +125,64 @@ func (c InternalClient) IsPublic() bool {
...
+//
+// Implements part of fosite.OpenIDConnectClient.
+func (c InternalClient) GetJSONWebKeysURI() string {
+	return c.JSONWebKeysURI
+}
+
+// GetRequestObjectSigningAlgorithm returns the object sig alg.
+//
+// Implements part of fosite.OpenIDConnectClient.
+func (c InternalClient) GetRequestObjectSigningAlgorithm() string {
```
clems4ever, line 164: should be in a dedicated PR as discussed.

internal/oidc/keys.go
```diff
@@ -0,0 +1,204 @@
...
+	manager = new(KeyManager)
+	manager.keys = map[string]*rsa.PrivateKey{}
+	manager.keySet = new(jose.JSONWebKeySet)
+
+	return manager
+}
+
+// KeyManager keeps track of all of the active/inactive rsa keys and provides them to services requiring them.
+// It additionally allows us to add keys for the purpose of key rotation in the future.
+type KeyManager struct {
```
clems4ever, line 42: Please move the struct on top of the file.

internal/oidc/provider.go
```diff
@@ -1,28 +1,24 @@
...
 
 // OpenIDConnectProvider for OpenID Connect.
 type OpenIDConnectProvider struct {
-	privateKeys map[string]*rsa.PrivateKey
-
-	Fosite fosite.OAuth2Provider
-	Store  *OpenIDConnectStore
+	Fosite    fosite.OAuth2Provider
+	Store     *OpenIDConnectStore
+	Writer    *herodot.JSONWriter
```
clems4ever, line 16: Do we really need to introduce this dependency and expose this Writer publicly out of this package?
I'd rather avoid depending on a new lib to serialize JSON unless there is a deep reason behind which would need to be explicited in a comment.

internal/handlers/const.go
```diff
@@ -58,12 +58,12 @@
 
 // OIDC constants.
 const (
-	oidcWellKnownPath  = "/.well-known/openid-configuration"
 	oidcJWKsPath       = "/api/oidc/jwks"
 	oidcAuthorizePath  = "/api/oidc/authorize"
 	oidcTokenPath      = "/api/oidc/token" //nolint:gosec // This is not a hard coded credential, it's a path.
 	oidcIntrospectPath = "/api/oidc/introspect"
 	oidcRevokePath     = "/api/oidc/revoke"
+	oidcUserinfoPath   = "/api/oidc/userinfo"
```
autheliabot, line 66: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>`oidcUserinfoPath` is unused (deadcode)

internal/handlers/handler_oidc_userinfo.go
james-d-elliott: I am fairly certain. I believe in order to obtain userinfo you need to do introspection.

internal/session/user_session.go
```diff
@@ -12,3 +15,49 @@
 		LastActivity:        0,
 	}
 }
+
+// SetOneFactor sets the expected property values for one factor authentication.
+func (s *UserSession) SetOneFactor(now time.Time, details *authentication.UserDetails, keepMeLoggedIn bool) {
+	if s.FirstFactorAuthn == 0 {
+		s.FirstFactorAuthn = now.Unix()
+	}
+	s.LastActivity = now.Unix()
```
autheliabot, line 24: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>assignments should only be cuddled with other assignments (wsl)

internal/handlers/handler_firstfactor.go
james-d-elliott: Fixed.

internal/oidc/client.go
james-d-elliott: Resolved!

internal/oidc/client.go
james-d-elliott: Removed.

internal/oidc/client.go
james-d-elliott: Removed.

internal/oidc/keys.go
james-d-elliott: Done.

config.template.yml
james-d-elliott: I've done this. Did 60s, and documented it.

config.template.yml
james-d-elliott: I adjusted this to 90m and documented it. Since people can tweak it there should be no issues. I also added a link to those docs.

docs/configuration/identity-providers/oidc.md
james-d-elliott: Done.

internal/utils/strings.go
```diff
@@ -1,12 +1,29 @@
...
+// IsStringAbsURL checks a string can be parsed as a URL and that is IsAbs and if it can't it returns an error
+// describing why.
+func IsStringAbsURL(input string) (err error) {
+	parsedURL, err := url.Parse(input)
+	if err != nil {
+		return fmt.Errorf("could not parse '%s' as a URL", input)
+	}
+
+	if !parsedURL.IsAbs() {
+		return fmt.Errorf("url '%s' is not absolute", input)
```
nightah, line 21: I think this error might confuse non-technical folk.

nightah (changes requested):
Mostly looks good, picked up a couple small things in here.

Couple things I noticed that haven't been called out in the review commentary:

* I think your first dot point in the PR body needs to change
* There was a large regression for coverage on `internal/session/user_session.go` by virtue of the refactor, this should probably be investigated

internal/handlers/handler_oidc_wellknown.go
```diff
@@ -7,11 +7,11 @@ import (
 	"github.com/valyala/fasthttp"
 
 	"github.com/authelia/authelia/internal/middlewares"
+	"github.com/authelia/authelia/internal/oidc"
 )
 
 func oidcWellKnown(ctx *middlewares.AutheliaCtx) {
-	var configuration WellKnownConfigurationJSON
-
+	// TODO (james-d-elliott): append the server.path here for path based installs. Also check other instances in OIDC.
```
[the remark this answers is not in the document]
james-d-elliott: So the well known file sends  URL's for the expected endpoints. The way it is used currently the endpoint for well-known will be `https://example.com/authelia/.well-known/openid-configuration`, but the external urls it advertise will be similar to `https://example.com/api/oidc/authorize` when they should be `https://example.com/authelia/api/oidc/authorize`. I have yet to test if this is the case, but can't see how it wouldn't be.

internal/configuration/validator/const.go
```diff
@@ -12,14 +12,28 @@
...
-	errFileHashing  = "config key incorrect: authentication_backend.file.hashing should be authentication_backend.file.password"
-	errFilePHashing = "config key incorrect: authentication_backend.file.password_hashing should be authentication_backend.file.password"
-	errFilePOptions = "config key incorrect: authentication_backend.file.password_options should be authentication_backend.file.password"
+	errFmtOIDCServerClientRedirectURI = "OIDC Client with ID '%s' redirect URI %s has an invalid scheme '%s', " +
+		"should be http or https"
+	errFmtOIDCServerClientRedirectURICantBeParsed = "OIDC Client with ID '%s' has an invalid redirect URI '%s' " +
+		"could not be parsed: %v"
+	errFmtOIDCServerClientInvalidPolicy = "OIDC Client with ID '%s' has an invalid policy '%s', " +
+		"should be either 'one_factor' or 'two_factor'"
+	errFmtOIDCServerClientInvalidSecret = "OIDC Client with ID '%s' has an empty secret"
```
autheliabot, line 21: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>G101: Potential hardcoded credentials (gosec)

internal/configuration/validator/const.go
```diff
@@ -43,12 +59,19 @@ const (
...
+	errAccessControlInvalidPolicyWithSubjects = "Policy [bypass] for rule #%d domain %s with subjects %s is invalid. " +
+		"It is not supported to configure both policy bypass and subjects. For more information see: " +
 		"https://www.authelia.com/docs/configuration/access-control.html#combining-subjects-and-the-bypass-policy"
 )
 
 var validLoggingLevels = []string{"trace", "debug", "info", "warn", "error"}
+var validScopes = []string{"openid", "email", "profile", "groups", "offline_access"}
+var validOIDCGrantTypes = []string{"implicit", "refresh_token", "authorization_code", "password", "client_credentials"}
+var validOIDCResponseTypes = []string{
+	"code", "code id_token", "id_token", "token id_token", "token", "token id_token code",
```
clems4ever, line 71: what about all permutations of `token id_token code`?

internal/oidc/keys.go
```diff
@@ -0,0 +1,195 @@
...
+func (m KeyManager) GetActivePrivateKey() (key *rsa.PrivateKey, err error) {
+	if key, ok := m.keys[m.activeKeyID]; ok {
+		return key, nil
+	}
+
+	return nil, errors.New("failed to retrieve active private key")
+}
+
+// AddActiveKeyData adds a rsa.PublicKey given the key in the PEM string format, then sets it to the active key.
+func (m *KeyManager) AddActiveKeyData(data string) (key *rsa.PrivateKey, webKey *jose.JSONWebKey, err error) {
```
clems4ever, line 88: I would rename this method to AddActivePrivateKeyData to make it clear that the data should be a private key.

internal/oidc/keys.go
```diff
@@ -0,0 +1,195 @@
...
+		Algorithm: "RS256",
+		Use:       "sig",
+	}
+
+	keyID, err := wk.Thumbprint(crypto.SHA1)
+	if err != nil {
+		return nil, err
+	}
+
+	strKeyID := strings.ToLower(fmt.Sprintf("%x", keyID)[0:6])
```
clems4ever, line 112: please add a boundary check just in case. You can never be too careful.

internal/oidc/keys.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil, fmt.Errorf("key id %s already exists", strKeyID)
+	}
+
+	// TODO: Add Mutex here when implementing key rotation.
+	wk.KeyID = strKeyID
+	m.keySet.Keys = append(m.keySet.Keys, wk)
+	m.keys[strKeyID] = key
+	m.activeKeyID = strKeyID
+
+	if m.strategy == nil {
```
clems4ever, line 124: I would avoid the lazy assignment if possible and rather assign the strategy when the manager is created unless I missed something.

internal/oidc/store.go
```diff
@@ -14,51 +14,30 @@ import (
 )
 
 // NewOpenIDConnectStore returns a new OpenIDConnectStore using the provided schema.OpenIDConnectConfiguration.
-func NewOpenIDConnectStore(configuration *schema.OpenIDConnectConfiguration) (store *OpenIDConnectStore) {
-	store = &OpenIDConnectStore{}
+func NewOpenIDConnectStore(configuration *schema.OpenIDConnectConfiguration) (store *OpenIDConnectStore, err error) {
+	store = &OpenIDConnectStore{
+		memory: &storage.MemoryStore{
+			IDSessions:             make(map[string]fosite.Requester),
```
clems4ever, line 20: why instantiating differently compared to the others below?

internal/session/provider.go
```diff
@@ -77,6 +77,8 @@ func NewProvider(configuration schema.SessionConfiguration, certPool *x509.CertP
 func (p *Provider) GetSession(ctx *fasthttp.RequestCtx) (UserSession, error) {
 	store, err := p.sessionHolder.Get(ctx)
 
+	store.GetSessionID()
```
clems4ever, line 80: I don't understand this line. Is it some kind of side effect? If it is it's worth either renaming it or documenting it with a comment.

internal/session/types.go
```diff
@@ -36,6 +36,8 @@ type UserSession struct {
 	KeepMeLoggedIn      bool
 	AuthenticationLevel authentication.Level
 	LastActivity        int64
+	FirstFactorAuthn    int64
```
clems4ever, line 39: I would append Timestamp. Something like FirstFactorAuthnTimestamp and same for the second factor timestamp.

internal/session/types.go
```diff
@@ -70,4 +72,5 @@ type OIDCWorkflowSession struct {
 	TargetURI                  string
 	AuthURI                    string
 	RequiredAuthorizationLevel authorization.Level
+	Created                    int64
```
clems4ever, line 75: Append Timestamp as well. It should be clear what this variable holds by simply reading the name if there is no dedicated type.

internal/session/user_session.go
```diff
@@ -12,3 +15,50 @@ func NewDefaultUserSession() UserSession {
...
+
+	s.Username = details.Username
+	s.DisplayName = details.DisplayName
+	s.Groups = details.Groups
+	s.Emails = details.Emails
+}
+
+// SetTwoFactor sets the expected property values for two factor authentication.
+func (s *UserSession) SetTwoFactor(now time.Time) {
+	if s.SecondFactorAuthn == 0 {
```
clems4ever, line 38: why checking? why not simply always assigning the current timestamp?

internal/session/user_session.go
```diff
@@ -12,3 +15,50 @@ func NewDefaultUserSession() UserSession {
...
+// AuthenticatedAt returns the unix timestamp this session authenticated successfully at the given level.
+func (s UserSession) AuthenticatedAt(level authorization.Level) (authenticatedAt time.Time) {
+	switch level {
+	case authorization.OneFactor:
+		return time.Unix(s.FirstFactorAuthn, 0)
+	case authorization.TwoFactor:
+		return time.Unix(s.SecondFactorAuthn, 0)
+	}
+
+	if s.SecondFactorAuthn != 0 {
```
clems4ever, line 55: In my opinion and unless I missed something, this function should raise an error if we are not matching a known state (what is covered by the switch block). It's better to advertise the admin when something is not happening as expected but for sure we should not let erroneous cases execute silently with best-effort values. It might be ok in a non-security oriented component but here in my opinion we should be on the safe side if that makes sense. What do you think?

clems4ever (changes requested):
Looks good. Few comments on the latest revision.

internal/session/provider.go
james-d-elliott: Unintentional, must have been debugging something. Whoops!

internal/oidc/store.go
james-d-elliott: Oh I'm not sure why we did that, that's preexisting, let me change it and see if there was a  reason, but doubt it.

internal/oidc/keys.go
james-d-elliott: This was done because it should be set when we assign the active key. However the nil check is not useful I don't think.

internal/configuration/validator/const.go
james-d-elliott: I removed this. I think we need to carefully check the spec before we implement this. Your comment actually triggered this thought. I believe there are only certain orders/combinations that are technically accepted by the spec, though it would be good to validate this one way or another.

internal/oidc/keys.go
james-d-elliott: Can you take a look at this adjustment? I think it's adequate.

config.template.yml
```diff
@@ -595,6 +595,10 @@ notifier:
     ## Enables additional debug messages.
     # enable_client_debug_messages: false
 
+    ## SECURITY NOTICE: It's not recommended to change this option, and highly discouraged to have it above 8 for
```
nightah, line 598:
```suggestion
    ## SECURITY NOTICE: It's not recommended to change this option, and highly discouraged to have it below 8 for
```

I think this is what you're meaning to say here?

internal/configuration/config.template.yml
```diff
@@ -595,6 +595,10 @@ notifier:
     ## Enables additional debug messages.
     # enable_client_debug_messages: false
 
+    ## SECURITY NOTICE: It's not recommended to change this option, and highly discouraged to have it above 8 for
```
nightah, line 598:
```suggestion
    ## SECURITY NOTICE: It's not recommended to change this option, and highly discouraged to have it below 8 for
```

internal/configuration/validator/const.go
```diff
@@ -25,6 +25,8 @@ const (
 		"must be one of: '%s'"
 	errFmtOIDCServerClientInvalidResponseMode = "OIDC Client with ID '%s' has an invalid response mode '%s', " +
 		"must be one of: '%s'"
+	errFmtOIDCServerInsecureParameterEntropy = "SECURITY ISSUE: OIDC minimum parameter entropy is configured to an " +
+		"unsafe value, it should be above 8 but it's configured to %d."
```
nightah, line 29:
```suggestion
		"unsafe value, it should be above 8 but it's configured to %d."
```
If I'm wrong about the first one then this should be adjusted too, otherwise this is good.

nightah (changes requested):
Small nit in the configuration notes.
Also should we be including this in the docs too?

internal/configuration/validator/const.go
james-d-elliott: You're not, it should be above 8, nice catch.

georglauterbach:
Nice to see this being worked on for #2126. When this is merged, I should be able to test the user info endpoint with a test image for GitLab.

internal/configuration/validator/const.go
clems4ever: Ok great.

internal/oidc/keys.go
clems4ever: I'm fine with the current implementation but I'm wondering what prevents us to store the strategy in each key instead of at the key manager level? Maybe an improvement for a future review.

internal/session/provider.go
clems4ever: np

clems4ever (approved):
LGTM

internal/handlers/handler_oidc_introspect.go
```diff
@@ -7,14 +7,7 @@ import (
...
 func oidcIntrospect(ctx *middlewares.AutheliaCtx, rw http.ResponseWriter, req *http.Request) {
-	oidcSession, err := newDefaultOIDCSession(ctx)
-
-	if err != nil {
-		ctx.Logger.Errorf("Error occurred in NewDefaultOIDCSession: %+v", err)
-		ctx.Providers.OpenIDConnect.Fosite.WriteIntrospectionError(rw, err)
-
-		return
-	}
+	oidcSession := newOpenIDSession("")
```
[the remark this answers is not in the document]
james-d-elliott: It's just refactoring it as discussed.. I will however see what's involved in doing an introspect endpoint request.

PR #1996, opened 2021-05-10
fix(oidc): implement required subject types supported meta

james-d-elliott:
The well known openid-configuration should have the subject_types_supported meta values.


nightah (approved):
LGTM.

PR #2008, opened 2021-05-16
Fix: Add missing lines for bulding frontend

payamnaghdy:
-Build front-end and copy files to internal/server/public_html
	-Copy api to internal/server/public_html

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,14 @@
+# ========================================
+# ===== Build image for the frontend =====
+# ========================================
+FROM node:16-alpine AS builder-frontend
+
+WORKDIR /node/src/app
+COPY web .
+
+# Install the dependencies and build
+RUN yarn install --frozen-lockfile && INLINE_RUNTIME_CHUNK=false yarn coverage
+
 # =======================================
 # ===== Build image for the backend =====
 # =======================================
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -20,7 +31,8 @@ COPY internal internal
 
 # Prepare static files to be embedded in Go binary
 RUN rm -rf internal/server/public_html
-COPY public_html internal/server/public_html
+COPY --from=builder-frontend /node/src/app/build internal/server/public_html
```
76090, line 34: Information from you about this is going good.

76090: N What are you doing for your family for the holidays? P You can save money on your account balance and your family if you want to be in a better connection.

76090 (approved):

nightah:
This is intended because the frontend is built and cached/utilised in the latter steps within our CI/CD environment:

https://github.com/authelia/authelia/blob/3ff50ae9794c5e43e672aa9047fe07107fc877e3/cmd/authelia-scripts/cmd_build.go#L107-L128

https://github.com/authelia/authelia/blob/3ff50ae9794c5e43e672aa9047fe07107fc877e3/.buildkite/pipeline.sh#L35-L43

If for whatever reason you want/need to manually build an Authelia container you can download and extract the frontend assets from [GitHub](https://github.com/authelia/authelia/releases/download/v4.29.2/authelia-v4.29.2-public_html.tar.gz) for the version you're intending to build and extract that in the root of the Authelia directory before you run the Docker build (this is how it works within our CI/CD).

PR #2016, opened 2021-05-21
fix(handlers): align response status codes for the verify endpoint

james-d-elliott:
This aligns all response status codes on the /api/verify endpoint when an error occurs, making it impossible to determine the actual reason for the failure.


clems4ever:
LGTM

clems4ever (approved):
LGTM

landloos:
Thought I'd just report that this broke my Authelia. I just get a blank page instead of the login page from this version (4.29.3) onwards. If I roll back to 4.29.2 it works. I use the official docker images.

nightah:
This was due to a library bump, the issue has already been resolved in `4.29.4`.

landloos:
Thanks. I'll go and update to that and check again.

PR #2046, opened 2021-05-30
feat(web): implement automatic theme switch for light/dark

AlexGustafsson:
The `auto` theme will automatically switch between `dark` and `light` depending on user preference. This allows for automatic dark mode.


nightah:
Thanks for this PR @AlexGustafsson, for this to work you will also need to make updates at the following location:

https://github.com/authelia/authelia/blob/1c510edbf31b314aaed5b9723b4d4deaabfac9e1/internal/configuration/validator/theme.go#L12-L14

Also when you get an opportunity could you please read out [commit message guidelines](https://www.authelia.com/docs/contributing/commitmsg-guidelines.html).

AlexGustafsson (commented):
> Thanks for this PR @AlexGustafsson, for this to work you will also need to make updates at the following location:
> 
> https://github.com/authelia/authelia/blob/1c510edbf31b314aaed5b9723b4d4deaabfac9e1/internal/configuration/validator/theme.go#L12-L14
> 
> Also when you get an opportunity could you please read our [commit message guidelines](https://www.authelia.com/docs/contributing/commitmsg-guidelines.html).

I'll take a look. Is it acceptable to force push in this case in order to correct the history?

nightah:
No need to correct the history, just add another commit with the update.
We will end up squashing when it gets committed to `master` anyway.

docs/configuration/theme.md
```diff
@@ -31,3 +31,5 @@ There are currently 3 available themes for Authelia:
 * light (default)
 * dark
 * grey
+
+To enable automatic switching between the light and dark theme you can set the theme to `auto`.
```
clems4ever, line 35: Can you please document what criteria is the automatic switch based on. From what I understand it's based on 'prefers-color-scheme' media query.

web/src/App.tsx
```diff
@@ -37,16 +37,23 @@ function Theme() {
...
             return themes.Light;
     }
 }
 
 const App: React.FC = () => {
     const [notification, setNotification] = useState(null as Notification | null);
-
+    const [theme, setTheme] = useState(Theme());
+    if (getTheme() === "auto") {
+        window.matchMedia("(prefers-color-scheme: dark)").addEventListener("change", (e) => {
```
clems4ever, line 51: The listener should be a callback created with useCallback and setTheme should be a dependency of this callback I think.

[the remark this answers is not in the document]
clems4ever: Sounds like reasonable proposals.

clems4ever (changes requested):

docs/configuration/theme.md
nightah: Indeed which I believe relies on the OS or application setting for light/dark mode.

AlexGustafsson: Fixed. Should it be even more clear / technical? If so, perhaps it would make sense to link to https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme.

web/src/App.tsx
AlexGustafsson: I'm very new to React so I might be wrong, but it seems as if the `useCallback` hook will call a function if a dependency is changed. In this case, the value used is not a stateful dependency, so I don't think `useCallback` would be applicable.

docs/configuration/theme.md
```diff
@@ -31,3 +31,5 @@ There are currently 3 available themes for Authelia:
 * light (default)
 * dark
 * grey
+
+To enable automatic switching between themes, you can set `theme` to `auto`. The theme will be set to either "dark" or "light" depending on users' system preference.
```
nightah, line 35:
```suggestion
To enable automatic switching between themes, you can set `theme` to `auto`. The theme will be set to either `dark` or `light` depending on the users system preference.
```

AlexGustafsson: Fixed.

docs/configuration/theme.md
james-d-elliott: Yep I agree with linking. The general style of our docs is explain reasonably and for complex topics add something like this:

To read more technical details see the [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)

docs/configuration/theme.md
```diff
@@ -31,3 +31,5 @@ There are currently 3 available themes for Authelia:
 * light (default)
 * dark
 * grey
+
+To enable automatic switching between themes, you can set `theme` to `auto`. The theme will be set to either `dark` or `light` depending on the user's system preference.
```
james-d-elliott, line 35:
```suggestion
To enable automatic switching between themes, you can set `theme` to `auto`. The theme will be set to either `dark` or `light` depending on the user's system preference which is determined using media queries. To read more technical details about the media queries used, read the [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).
```

james-d-elliott (commented):
Thanks again for your contributions! Additionally something like this I think covers what everyone has suggested. What does everyone think? @clems4ever @nightah

docs/configuration/theme.md
nightah: SGTM.

clems4ever: SGTM to me too. That's what I had in mind.

web/src/App.tsx
clems4ever: The component function will be executed everytime the component needs to be rendered (which is out of your control) and in general in React you want to control when the callbacks are called so in general the calls should not be in the rendered section at all but rather in some kind of callback. In your case, you're right, you want to execute setTheme only once (no dep change) so I was wrong, you need a useEffect instead of useCallback with setTheme in the dependency list.

clems4ever (changes requested):

web/src/App.tsx
AlexGustafsson: Thanks for the clarification, I see the issue now.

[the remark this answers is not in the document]
AlexGustafsson: Fixed.

docs/configuration/theme.md
AlexGustafsson: Fixed.

nightah:
@AlexGustafsson this LGTM just need you to resolve the conflict from `master` so this can be merged.

nightah:
Looks like resolving the merge conflict has resulted in a small linting error: https://github.com/authelia/authelia/runs/2846063162

You can make all of the theme names a const to resolve that.

AlexGustafsson:
> @AlexGustafsson this LGTM just need you to resolve the conflict from `master` so this can be merged.

Well, it's resolved for the most part now. There's still one conflicting line in `internal/configuration/validator/theme.go`:

```diff
-validThemes := regexp.MustCompile("light|dark|grey")
+validThemes := regexp.MustCompile("light|dark|grey|auto")
```

I don't know how to fix it without rebasing off of master. Do you usually merge such things manually or do you want me to rebase this time so that the merge will be "clean"?

nightah:
> > @AlexGustafsson this LGTM just need you to resolve the conflict from `master` so this can be merged.
> 
> @nightah Well, it's resolved for the most part now. There's still one conflicting line in `internal/configuration/validator/theme.go`:
> 
> ```diff
> -validThemes := regexp.MustCompile("light|dark|grey")
> +validThemes := regexp.MustCompile("light|dark|grey|auto")
> ```
> 
> I don't know how to fix it without rebasing off of master. Do you usually merge such things manually or do you want me to rebase this time so that the merge will be "clean"?

It's okay I've resolved that merge conflict just need that last linting issue resolved, if you prefer I can fix it.

Up to you though!

AlexGustafsson:
@nightah Fix it if you like so that you get the constants where you'd want them.

nightah:
Looks like it was an issue because of the merge conflict, no change required!

nightah (approved):
LGTM!

Thanks for your efforts on this @AlexGustafsson.

nightah:
@all-contributors please add @AlexGustafsson for code and docs.

Issue #1709, opened 2021-02-09
Use config for hash-password


ThinkChaos:
It would be nice to be able to pass `--config` to hash-password to avoid having to duplicate the info in the config on the CLI.

nightah:
This would be possible to do and not overly difficult, I envisage that you would need to pass in both the config and the specific username that you're passing the hash in so it knows what to replace.

ThinkChaos:
Sorry I wasn't clear, I was thinking of reading the config.yml's `authentication_backend.file.password` section to avoid having to pass `-i`, `-k`, `-m`, etc. to `authelia hash-password`.

AlexGustafsson:
@ThinkChaos could you take a look at #2047 to see if that would fit your use case?

I might be mistaken, but I think you should be able to test it out using Docker once the pipeline passes:

```sh
docker run authelia/authelia:PR2047 hash-password --help
docker run authelia/authelia:PR2047 hash-password --config ./my-config.yml 'password'
```

ThinkChaos:
I can confirm your PR works, I just had to adjust the command to:

    docker run -v "$PWD"/authelia/data:/data:ro authelia/authelia:PR2047 -- authelia hash-password --config /data/configuration.yml badpassword

PR #2047, opened 2021-05-30
feat: add config flag to hash-password tool

AlexGustafsson:
Solves #1709 by implementing a `--config` flag for the `hash-password` which parses the config and validates it just as it would at run-time. The values specified in the config replaces those specified as parameters.


AlexGustafsson:
I'm not that happy with the workaround for not creating the configuration file if it was not found. Another option would be to create a new function like `ReadWithoutCreating` and just duplicate a lot of code. Optimally, the existing `Read` function wouldn't handle the creation of the configuration file at all.

Also, the way it's implemented now, the flags will be overwritten by the configuration file. It was mostly done to change as little code as possible to introduce the feature, but it might be worth revisiting so that flags overwrite the values from the configuration file instead.

james-d-elliott:
> 
> 
> I'm not that happy with the workaround for not creating the configuration file if it was not found. Another option would be to create a new function like `ReadWithoutCreating` and just duplicate a lot of code. Optimally, the existing `Read` function wouldn't handle the creation of the configuration file at all.
> 
> Also, the way it's implemented now, the flags will be overwritten by the configuration file. It was mostly done to change as little code as possible to introduce the feature as well as the described use case didn't mention using both a config file and flags. But it might be worth revisiting so that flags overwrite the values from the configuration file instead.

Maybe move the code to create the config to a function that calls the original function at the end.

AlexGustafsson:
> Maybe move the code to create the config to a function that calls the original function at the end.

It's a bit of a more intrusive change, but is it alright if I instead separate concerns so we have a `Read` function and a `Create` function? Then, each use of `configuration.Read` would be changed so that the `Create` function is called before if necessary (only the server start right now?). Your suggested change works, but it feels a bit bad to rename the function `CreateAndRead` and then introduce a `Read` when we may as well have `Create` and `Read` separated.

james-d-elliott:
> 
> 
> > Maybe move the code to create the config to a function that calls the original function at the end.
> 
> It's a bit of a more intrusive change, but is it alright if I instead separate concerns so we have a `Read` function and a `Create` function? Then, each use of `configuration.Read` would be changed so that the `Create` function is called before if necessary (only the server start right now?). Your suggested change works, but it feels a bit bad to rename the function `CreateAndRead` and then introduce a `Read` when we may as well have `Create` and `Read` separated.

Sounds fine to me as long as we're not duplicating code.

internal/configuration/generator_test.go
```diff
@@ -0,0 +1,40 @@
+package configuration
+
+import (
+	"os"
+	"runtime"
+	"testing"
+
```
autheliabot, line 7: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>File is not `goimports`-ed with -local github.com/authelia/authelia (goimports)

cmd/authelia/main.go
```diff
@@ -27,7 +27,14 @@
 //nolint:gocyclo // TODO: Consider refactoring/simplifying, time permitting.
 func startServer() {
 	logger := logging.Logger()
+	created, err := configuration.GenerateIfNotFound(configPathFlag)
+	if err != nil {
```
autheliabot, line 31: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>only one cuddle assignment allowed before if statement (wsl)

cmd/authelia/main.go
```diff
@@ -27,7 +27,14 @@
...
+		logger.Error(err)
+	}
+	if created {
```
autheliabot, line 34: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>if statements should only be cuddled with assignments (wsl)

cmd/authelia/main.go
james-d-elliott: It's just asking you to add a new line before the if.

cmd/authelia/main.go
james-d-elliott: It's just asking you to add a new line before the if.

james-d-elliott (changes requested):
Mostly looks like a nice PR thanks for contributing!

cmd/authelia/main.go
```diff
@@ -27,6 +27,13 @@ var configPathFlag string
 //nolint:gocyclo // TODO: Consider refactoring/simplifying, time permitting.
 func startServer() {
 	logger := logging.Logger()
+	created, err := configuration.GenerateIfNotFound(configPathFlag)
```
clems4ever, line 30: I don't understand how this relates to the problem mentioned in the title of the PR. Should probably be part of a dedicated PR where we want to generate a default configuration if none is provided.

It's a good idea to generate the configuration but I would rather let Authelia fails if the file is not found and rather provide this feature using a new documented CLI command. I'd rather make it explicit that something is wrong when starting up Authelia rather than let the user figure out very late that the config was not at the right place and another one has been generated "silently" in the meantime (even a log line saying the config has been generated means silent for me because I make the assumption not everyone read logs carefully and I wan't to avoid that toil).

clems4ever (changes requested):

cmd/authelia/main.go
james-d-elliott: How it relates is if a user tries to hash-password and gives an invalid path it creates a configuration file there. This behavior is used with the container on the config volume on startup.

AlexGustafsson: Perhaps it makes sense to handle these things in separate PRs - close this and create a new one once the configuration issue has been resolved in a new PR? That way we can easily focus on getting each part done in a way that works for everyone.

james-d-elliott: Yeah that makes sense to me @AlexGustafsson If you just do the basic portion of getting the hash-password to use the config if provided I will do that work. I'm actually doing work to implement a new config lib which should make all of this a lot cleaner.

AlexGustafsson: @james-d-elliott could you ping me here once that's merged or ready for merge? I'll fix it then.

james-d-elliott: Absolutely! Should be done this weekend, but will ping you. It's a fairly hefty refactor of configuration, but I can help merge the changes if you have issues either way.

ThinkChaos:
I tested this PR as mentioned in #1709 and it works fine with the caveats already mentioned.

internal/commands/hash.go
```diff
@@ -19,6 +21,7 @@ func init() {
 	HashPasswordCmd.Flags().IntP("parallelism", "p", schema.DefaultPasswordConfiguration.Parallelism, "[argon2id] set the parallelism param")
 	HashPasswordCmd.Flags().IntP("key-length", "k", schema.DefaultPasswordConfiguration.KeyLength, "[argon2id] set the key length param")
 	HashPasswordCmd.Flags().IntP("salt-length", "l", schema.DefaultPasswordConfiguration.SaltLength, "set the auto-generated salt length")
+	HashPasswordCmd.Flags().String("config", "", "Configuration file")
```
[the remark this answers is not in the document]
james-d-elliott: Marking this resolved as this is more applicable in the refactor.

nightah:
@AlexGustafsson sorry for the long wait on this, we've not merged the major configuration refactor if you want to rebase your changes.

There also were significant changes so please don't hesitate to reach out if you need help.

nightah:
@AlexGustafsson do you have time and are you happy to rebase this from `master`? Or would you prefer for us to finish this off?

james-d-elliott (approved):
LGTM

nightah:
Thanks for this PR @AlexGustafsson!
One thing to consider in future is [signing your commits](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification/signing-commits) with a GPG key.

james-d-elliott:
@allcontributors please add @AlexGustafsson for code

PR #2048, opened 2021-05-31
fix(docs): Update link to Lite bundle

fhemberger:


clems4ever:
Thanks for the fix @fhemberger!

PR #2050, opened 2021-05-31
feat(configuration): add error and warn log levels

james-d-elliott:
This is so levels like warn and error can be used to exclude info or warn messages. Additionally there is a reasonable refactoring of logging moving the log config options to the logging key because there are a significant number of log options now. This also decouples the expvars and pprof handlers from the log level, and they are now configured by server.enable_expvars and server.enable_pprof at any logging level.

Closes #2049.

### DIFF
--- a/internal/configuration/schema/logging.go
+++ b/internal/configuration/schema/logging.go
@@ -1,5 +1,6 @@
 package schema
 
+// LoggingConfiguration represents the logging configuration.
 type LoggingConfiguration struct {
 	Level      string `mapstructure:"level"`
 	Format     string `mapstructure:"format"`
@@ -7,6 +8,7 @@ type LoggingConfiguration struct {
 	KeepStdout bool   `mapstructure:"keep_stdout"`
 }
 
+// DefaultLoggingConfiguration is the default logging configuration.
 var DefaultLoggingConfiguration = LoggingConfiguration{
 	Level:  "info",
 	Format: "text",
--- a/internal/configuration/schema/logging.go
+++ b/internal/configuration/schema/logging.go
@@ -1,0 +1,13 @@
+package schema
+
+type LoggingConfiguration struct {
+	Level      string `mapstructure:"level"`
+	Format     string `mapstructure:"format"`
+	FilePath   string `mapstructure:"file_path"`
+	KeepStdout bool   `mapstructure:"keep_stdout"`
+}
+
+var DefaultLoggingConfiguration = LoggingConfiguration{
+	Level:  "info",
+	Format: "text",
+}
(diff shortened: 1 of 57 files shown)


internal/configuration/schema/logging.go
```diff
@@ -0,0 +1,13 @@
+package schema
+
+type LoggingConfiguration struct {
```
autheliabot, line 3: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>exported type `LoggingConfiguration` should have comment or be unexported (golint)

internal/configuration/schema/logging.go
```diff
@@ -0,0 +1,13 @@
...
+	Level      string `mapstructure:"level"`
+	Format     string `mapstructure:"format"`
+	FilePath   string `mapstructure:"file_path"`
+	KeepStdout bool   `mapstructure:"keep_stdout"`
+}
+
+var DefaultLoggingConfiguration = LoggingConfiguration{
```
autheliabot, line 10: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>exported var `DefaultLoggingConfiguration` should have comment or be unexported (golint)

internal/configuration/validator/const.go
```diff
@@ -66,19 +72,28 @@ var validKeys = []string{
...
-	"log_format",
-	"log_file_path",
 	"default_redirection_url",
 	"theme",
 	"tls_key",
 	"tls_cert",
 	"certificates_directory",
 
+	// Logging keys.
+	"log_level",
```
nightah, line 82: Is it worthwhile putting deprecate/TODO around these keys?

internal/configuration/validator/logging.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if configuration.Logging.Level == "" {
+		configuration.Logging.Level = schema.DefaultLoggingConfiguration.Level
+	}
+
+	if configuration.Logging.Format == "" {
+		configuration.Logging.Format = schema.DefaultLoggingConfiguration.Format
+	}
+
+	if configuration.Logging.FilePath == "" && configuration.Logging.KeepStdout {
```
nightah, line 23: This set of configuration (which I suspect would be incorrectly applied by a user) would result in no logging.

I think we can strip the second part of the condition here and if `FilePath = ""` we should `log.Fatal` and exit so the user can setup their logging appropriately.

Mainly because KeepStdout doesn't actually do anything unless the FilePath is not blank:

https://github.com/authelia/authelia/blob/4cfda7eeceec77799bb4ccc7b783940426413365/internal/logging/logger.go#L33-L53

nightah (changes requested):
Good idea to refactor and move most of the default configurations into their own validator.

Only a couple small nits but otherwise this is good.

nightah (approved):
LGTM!

PR #2057, opened 2021-06-02
feat: password modify feature

Freebien:
Any connected user will be able to modify his own password when already connected.
The user will have to provide his actual password and the new one.

The user has to send a POST request to /api/user/password

```json
{
    "old_password": "oldPassword",
    "password": "password"
}
```

### DIFF
--- a/internal/handlers/handler_user_password.go
+++ b/internal/handlers/handler_user_password.go
@@ -8,7 +8,8 @@ import (
 )
 
 type updatePasswordRequestBody struct {
-	Password string `json:"password"`
+	OldPassword string `json:"old_password"`
+	Password    string `json:"password"`
 }
 
 // UserPasswordPost handler for resetting passwords.
@@ -19,9 +20,16 @@ func UserPasswordPost(ctx *middlewares.AutheliaCtx) {
 	err := ctx.ParseBody(&requestBody)
 
 	if err != nil {
-		ctx.Error(err, unableToResetPasswordMessage)
+		ctx.Error(err, unableToUpdatePasswordMessage)
 		return
 	}
+
+	ok, err := ctx.Providers.UserProvider.CheckUserPassword(userSession.Username, requestBody.OldPassword)
+	if !ok || err != nil {
+		ctx.Error(err, unableToUpdatePasswordMessage)
+		return
+	}
+
 	err = ctx.Providers.UserProvider.UpdatePassword(userSession.Username, requestBody.Password)
 
 	if err != nil {
@@ -31,7 +39,7 @@ func UserPasswordPost(ctx *middlewares.AutheliaCtx) {
 		case utils.IsStringInSliceContains(err.Error(), ldapPasswordComplexityErrors):
 			ctx.Error(fmt.Errorf("%s", err), ldapPasswordComplexityCode)
 		default:
-			ctx.Error(fmt.Errorf("%s", err), unableToResetPasswordMessage)
+			ctx.Error(fmt.Errorf("%s", err), unableToUpdatePasswordMessage)
 		}
 
 		return
--- a/internal/handlers/handler_user_password.go
+++ b/internal/handlers/handler_user_password.go
@@ -1,0 +1,43 @@
+package handlers
+
+import (
+	"fmt"
+
+	"github.com/authelia/authelia/internal/middlewares"
+	"github.com/authelia/authelia/internal/utils"
+)
+
+type updatePasswordRequestBody struct {
+	Password string `json:"password"`
+}
+
+// UserPasswordPost handler for resetting passwords.
+func UserPasswordPost(ctx *middlewares.AutheliaCtx) {
+	userSession := ctx.GetSession()
+
+	var requestBody updatePasswordRequestBody
+	err := ctx.ParseBody(&requestBody)
+
+	if err != nil {
+		ctx.Error(err, unableToResetPasswordMessage)
+		return
+	}
+	err = ctx.Providers.UserProvider.UpdatePassword(userSession.Username, requestBody.Password)
+
+	if err != nil {
+		switch {
+		case utils.IsStringInSliceContains(err.Error(), ldapPasswordComplexityCodes):
+			ctx.Error(fmt.Errorf("%s", err), ldapPasswordComplexityCode)
+		case utils.IsStringInSliceContains(err.Error(), ldapPasswordComplexityErrors):
+			ctx.Error(fmt.Errorf("%s", err), ldapPasswordComplexityCode)
+		default:
+			ctx.Error(fmt.Errorf("%s", err), unableToResetPasswordMessage)
+		}
+
+		return
+	}
+
+	ctx.Logger.Debugf("Password of user %s has been modified", userSession.Username)
+
+	ctx.ReplyOK()
+}
--- a/api/openapi.yml
+++ b/api/openapi.yml
@@ -322,6 +322,31 @@ paths:
           description: Forbidden
       security:
         - authelia_auth: []
+  /api/user/password:
+    post:
+      tags:
+        - Password Update
+      summary: Password Update
+      description: >
+        This endpoint permits the update of the user's password.
+
+        It verifies the user's old password and then updates it.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/handlers.updatePasswordRequestBody'
+      responses:
+        "200":
+          description: Successful Operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/middlewares.OkResponse'
+      security:
+        - authelia_auth: []
+
   /api/secondfactor/totp/identity/start:
     post:
       tags:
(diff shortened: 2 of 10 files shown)


james-d-elliott:
Hey thanks for your contributions! This one looks awesome. Just a small note the scope should be a single scope, I'd recommend handlers, but it's really which ever scope you think is most applicable.

internal/handlers/handler_user_password.go
```diff
@@ -0,0 +1,51 @@
...
+		return
+	}
+
+	ok, err := ctx.Providers.UserProvider.CheckUserPassword(userSession.Username, requestBody.OldPassword)
+	if !ok || err != nil {
+		ctx.Error(err, unableToUpdatePasswordMessage)
+		return
+	}
+
+	err = ctx.Providers.UserProvider.UpdatePassword(userSession.Username, requestBody.Password)
```
james-d-elliott, line 33: I would prefer to do it this way, though it would require extending the interface, and coding both providers to implement a method which does it. If you'd like some help then let me know. The advantage is we can make this change as the actual LDAP user in the future doing it this way I believe. 

```suggestion
	err = ctx.Providers.UserProvider.ChangePassword(userSession.Username, requestBody.OldPassword, requestBody.Password)
```

For reference `ldap.NewPasswordModifyRequest` which is from the library we use takes this exact syntax.

james-d-elliott (changes requested):
Let me know what you think, if you're not confident in doing this I can either help you or make a commit afterwards.

nightah:
> Hey thanks for your contributions! This one looks awesome. Just a small note the scope should be a single scope, I'd recommend handlers, but it's really which ever scope you think is most applicable.

Yep just to add to this, single scope if one package is clearly more affected otherwise you can omit the scope entirely.

nightah:
So I've just had a play with this PR and noticed that I'll never saw the `Update Password` button.
This is because based on your current changes only users which do not have any `two_factor` ACLs defined will reach the `AuthenticatedRoute`:

https://github.com/authelia/authelia/blob/ae2bfa36388f3f38e4c4bda3caf224da8a7053bd/web/src/views/LoginPortal/LoginPortal.tsx#L100-L102

To make this apply to all the respective areas you'll need to make a couple more changes, here's a patch:

```patch
Index: web/src/views/LoginPortal/SecondFactor/SecondFactorForm.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/web/src/views/LoginPortal/SecondFactor/SecondFactorForm.tsx b/web/src/views/LoginPortal/SecondFactor/SecondFactorForm.tsx
--- a/web/src/views/LoginPortal/SecondFactor/SecondFactorForm.tsx	(revision e02eb187908ef76f931f8ce1a9ce92ac9a75e1b1)
+++ b/web/src/views/LoginPortal/SecondFactor/SecondFactorForm.tsx	(date 1622687257398)
@@ -15,6 +15,7 @@
     SecondFactorPushRoute,
     SecondFactorU2FRoute,
     SecondFactorRoute,
+    UpdatePasswordRoute,
 } from "../../../Routes";
 import { initiateTOTPRegistrationProcess, initiateU2FRegistrationProcess } from "../../../services/RegisterDevice";
 import { AuthenticationLevel } from "../../../services/State";
@@ -88,6 +89,10 @@
         history.push(SignOutRoute);
     };
 
+    const handleUpdatePasswordClick = () => {
+        history.push(UpdatePasswordRoute);
+    };
+
     return (
         <LoginLayout id="second-factor-stage" title={`Hi ${props.userInfo.display_name}`} showBrand>
             <MethodSelectionDialog
@@ -101,6 +106,10 @@
                 <Grid item xs={12}>
                     <Button color="secondary" onClick={handleLogoutClick} id="logout-button">
                         Logout
+                    </Button>
+                    {" | "}
+                    <Button color="secondary" onClick={handleUpdatePasswordClick} id="update-password-button">
+                        Update Password
                     </Button>
                     {" | "}
                     <Button color="secondary" onClick={handleMethodSelectionClick} id="methods-button">
Index: web/src/views/LoginPortal/AuthenticatedView/AuthenticatedView.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/web/src/views/LoginPortal/AuthenticatedView/AuthenticatedView.tsx b/web/src/views/LoginPortal/AuthenticatedView/AuthenticatedView.tsx
--- a/web/src/views/LoginPortal/AuthenticatedView/AuthenticatedView.tsx	(revision e02eb187908ef76f931f8ce1a9ce92ac9a75e1b1)
+++ b/web/src/views/LoginPortal/AuthenticatedView/AuthenticatedView.tsx	(date 1622686801191)
@@ -18,6 +18,7 @@
     const handleLogoutClick = () => {
         history.push(SignOutRoute);
     };
+
     const handleUpdatePasswordClick = () => {
         history.push(UpdatePasswordRoute);
     };

```

nightah:
This is really well written @Freebien thanks a lot for your contribution.

Other then the aforementioned feedback, the only other element I would like addressed to see this merged is factorising the `UpdatePasswordView`, there's a lot of code duplication from `ResetPasswordStep2`.

Let me know if you're comfortable with this change or if you need some help with this.

You're also more than welcome to join our Matrix and/or Discord channels to collaborate with the team.

Freebien:
> This is really well written @Freebien thanks a lot for your contribution.
> 
> Other then the aforementioned feedback, the only other elements I would like addressed to see this merged is factorising the `UpdatePasswordView`, there's a lot of code duplication from `ResetPasswordStep2`, some tests would also be nice but we're happy to help there if you're not comfortable.
> 
> You're also more than welcome to join our Matrix and/or Discord channels to collaborate with the team.

Thanks a lot :)

I'll try to do some other things, I'm rather comfy with Go, but not that much with React, i'll try to do my best. I'm on Discord so we could have a chat if needed.

clems4ever:
Hello @Freebien , thanks for this contribution. It's a good proposal indeed. The only thing I'm not comfortable with yet with the current implementation is that the only requirement to update a password is to know the previous password. It seems a bit too weak requirement to me and maybe we should validate the identity of the user through email again to validate this sensitive operation.

For a 2FA setup, the current proposal is not too bad because a hacker can change a password while knowing the previous one but will still be blocked by 2nd factor to get access. However the user won't notice it... For single-factor setups though, I tend to think that we want to go back to an email to validate the password change, it would be a kind of second factor to validate this sensitive operation. So I propose we always send an email to the user once the password change is requested and it is only possible to change it after the link is clicked.
I'm also thinking that maybe sending back an email to the user when the password has successfully been changed can be interesting (maybe for an additional PR) because it could be a kind of attack detection in some cases. It could be done both for forgotten passwords or password change btw.

I'd be glad to hear what you guys think about my concern and proposal?

Freebien:
Hi !

This is a good idea and should be added in the configuration I guess. Some time, you have only one account to access all your services on your environment. For example, you have a mail service, and an internal git and both of those are linked to the same LDAP. the password will be the same and it will not be a bigger security measure, since you could have access to the mail, if it's not secured enough.

Also if the person is only using a password authorization, a hacker already have access to every services.

I guess we could add some mail configuration for every steps of every type of password reset stuff.

What do you think ?

clems4ever:
> Hi !
> 
> This is a good idea and should be added in the configuration I guess. Some time, you have only one account to access all your services on your environment. For example, you have a mail service, and an internal git and both of those are linked to the same LDAP. the password will be the same and it will not be a bigger security measure, since you could have access to the mail, if it's not secured enough.
> 
> Also if the person is only using a password authorization, a hacker already have access to every services.
> 
> I guess we could add some mail configuration for every steps of every type of password reset stuff.
> 
> What do you think ?

Thanks for your feedback @Freebien .

Authelia aims to be a fortress protecting sensitive assets and if users are not able to secure the rest of their environment, that's not Authelia's fault and they won't be able to state that it was because Authelia was the weak point. Also Authelia  is meant to be used by a single user, families as well as companies with many users and therefore when we design something we should also think about the bigger objectives in order for Authelia not to be turned into a snowflake handling all potential edge cases (for instance one single user in the DB).

That being said, I was not thinking about adding more configuration or do something fancy, rather use the notifier as it is by reusing the identity verification middleware. Everything already exists, I propose we just reuse it in your PR.

Let me know what you think.
Thanks.

nightah:
To make sure this doesn't fall to the wayside, why don't we make this option only available to a user if they have performed 2FA?

For users/setups that do not have any 2FA, the user can use the Forgot my Password to reset their password which will trigger the notifier to validate the sensitive action.

In a case where the user does know their password and wants to change let's make sure we force them to provide their 2FA before allowing the sensitive action?

Thoughts @clems4ever and @james-d-elliott?

clems4ever:
In my opinion we should always go back to confirmation with email whatever the number of factors. It's the standard identity validation steps used for many sensitive actions today, I suggest we stick to it. The configuration regarding the mail client is already mandatory and the identity verification workflow exists so I propose we just reuse them and we make the experience for users consistent.

piengeng:
is this stale or on-going?

PterX:
any update here?

james-d-elliott:
The user in question has abandoned it, so yes there is an update, it's being closed. However we have a flow that's being released that could also be used to ensure this can be securely achieved.

PR #2059, opened 2021-06-03
build: optimize docker builds

james-d-elliott:
This change is so that each of the COPY/RUN steps occurs in a single layer which should theoretically decrease build times.


nightah (approved):
LGTM.

PR #2072, opened 2021-06-08
refactor(configuration): use key log instead of logging

james-d-elliott:
This change makes the refactor from log_level etc to logging.level a refactor to log.level instead. This will be cleaner in the future with things like CLI args for configuration. Additionally it adds a command in the docs to checkout the latest tag for the repo.


nightah (approved):
LGTM.

PR #2078, opened 2021-06-10
ci(golangci-lint): replace golint with revive linter

nightah:
Remove deprecated `golint` linter and replace with `revive` linter.

Also fix outstanding issues due to upgraded linters.


james-d-elliott (approved):
LGTM

PR #2083, opened 2021-06-15
feat(handlers): synology password complexity err on reset

james-d-elliott:
This responds to the client with the correct error when used with Synology LDAP servers.


nightah (approved):
LGTM.

PR #2092, opened 2021-06-17
build(deps): replace jwt-go

james-d-elliott:
This replaces github.com/dgrijalva/jwt-go and github.com/form3tech-oss/jwt-go with github.com/golang-jwt/jwt which will be the maintained package going forward.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,14 +8,13 @@ require (
 	github.com/Workiva/go-datastructures v1.0.53
 	github.com/asaskevich/govalidator v0.0.0-20200907205600-7a23bdc65eef
 	github.com/deckarep/golang-set v1.7.1
-	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/duosecurity/duo_api_golang v0.0.0-20201112143038-0e07e9f869e3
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -36,6 +35,6 @@ require (
 )
 
 replace (
-	github.com/dgrijalva/jwt-go => github.com/form3tech-oss/jwt-go v3.2.1+incompatible
+	github.com/golang-jwt/jwt => github.com/golang-jwt/jwt v3.2.1+incompatible
```
oxisto, line 38: I think you can get rid of the replacement here and directly specify `github.com/golang-jwt/jwt v3.2.1+incompatible` in line 17 since all import paths are already adjusted and seem to work.

james-d-elliott: Thanks for the feedback! Regarding this we have deps that use those JWT libs as well, and we're aiming to force the use of github.com/golang-jwt/jwt in those as well.

nightah (approved):
LGTM.

go.mod
james-d-elliott: Looks like you're 100% accurate, for some reason fosite's use of the other jwt-go lib doesn't cause this to show up in our go.sum.

PR #2100, opened 2021-06-19
refactor(web): use absolute imports with aliases

nightah:
Refactors all of the TS/JS frontend to utilise absolute imports along with import aliases.
Each of the paths within `src` are represented with their own alias:

* `@assets`
* `@components`
* `@constants` (new)
* `@hooks`
* `@layouts`
* `@models`
* `@services`
* `@themes`
* `@utils`
* `@views`

`Routes.ts` and `constant.ts` have been relocated to the constants directory for consistency.


clems4ever (approved):
LGTM

PR #2101, opened 2021-06-19
refactor: drop cgo requirement for sqlite

nightah:
Replace github.com/mattn/go-sqlite3 with modernc.org/sqlite which drops our CGO requirement.


nightah:
@yurivict this change will have implications for the FreeBSD ports build as the SQLite package does not support FreeBSD yet.

You *should* be able to perform the following command to do a straight replace if continuing to build from source:

```bash
go mod edit -replace=github.com/gosqlite/sqlite=github.com/mattn/go-sqlite3@v1.14.7
```

Is there any chance you can test and confirm?

I'm also keen to hear your thoughts on how you we should best handle this, we could explore the option of building FreeBSD binaries and distributing those so you don't need to build from source if that's preferred?

clems4ever (approved):
This library has much less stars according to gitlab so it might not be as maintained (secure?) as the other one but given there is no big change to swap one for the other, let's give it a try and see if it's problematic in the future.

nightah:
> This library has much less stars according to gitlab so it might not be as maintained (secure?) as the other one but given there is no big change to swap one for the other, let's give it a try and see if it's problematic in the future.

I don't think that's going to be a huge concern it's actively developed and the lead developer is very active with Golang/Gopher related stuff.

I'd hold off on merging this either way until we've decided what we will do moving forward with the FreeBSD builds.

clems4ever:
Ok, good to know. I'm ok to merge whenever you're ready.

PR #2116, opened 2021-06-24
refactor: drop qemu binary requirement

nightah:
QEMU binaries no longer need to be baked into containers.


james-d-elliott (approved):
LGTM

PR #2117, opened 2021-06-25
refactor: cra build path

nightah:
The `authelia-scripts` helper currently performs steps to move files around in different stages of development and CI/CD.

We now utilise the `BUILD_PATH` environment variable to adjust the output directory for the web frontend from the default of `./web/build/` simplifying the helper somewhat.

Additionally we no longer build the Go binary in the unit test stage of our CI/CD as this is not necessary.


james-d-elliott (approved):
LGTM

PR #2124, opened 2021-06-28
feat(authentication): use the passwordmodify exop for pwd resets with ldap

ArsenArsen:
This was tested for OpenLDAP, I am unable to test for ActiveDirectory.

### DIFF
--- a/internal/authentication/ldap_connection_factory.go
+++ b/internal/authentication/ldap_connection_factory.go
@@ -15,6 +15,7 @@ type LDAPConnection interface {
 
 	Search(searchRequest *ldap.SearchRequest) (*ldap.SearchResult, error)
 	Modify(modifyRequest *ldap.ModifyRequest) error
+	PasswordModify(pwdModifyRequest *ldap.PasswordModifyRequest) error
 	StartTLS(config *tls.Config) error
 }
 
@@ -46,6 +47,12 @@ func (lc *LDAPConnectionImpl) Search(searchRequest *ldap.SearchRequest) (*ldap.S
 // Modify modifies an ldap object.
 func (lc *LDAPConnectionImpl) Modify(modifyRequest *ldap.ModifyRequest) error {
 	return lc.conn.Modify(modifyRequest)
+}
+
+// PasswordModify modifies an ldap objects password
+func (lc *LDAPConnectionImpl) PasswordModify(pwdModifyRequest *ldap.PasswordModifyRequest) error {
+	_, err := lc.conn.PasswordModify(pwdModifyRequest)
+	return err
 }
 
 // StartTLS requests the LDAP server upgrades to TLS encryption.
--- a/internal/authentication/ldap_connection_factory.go
+++ b/internal/authentication/ldap_connection_factory.go
@@ -49,7 +49,7 @@ func (lc *LDAPConnectionImpl) Modify(modifyRequest *ldap.ModifyRequest) error {
 	return lc.conn.Modify(modifyRequest)
 }
 
-// PasswordModify modifies an ldap objects password
+// PasswordModify modifies an ldap objects password.
 func (lc *LDAPConnectionImpl) PasswordModify(pwdModifyRequest *ldap.PasswordModifyRequest) error {
 	_, err := lc.conn.PasswordModify(pwdModifyRequest)
 	return err
--- a/internal/authentication/ldap_user_provider_test.go
+++ b/internal/authentication/ldap_user_provider_test.go
@@ -710,8 +710,9 @@ func TestShouldUpdateUserPassword(t *testing.T) {
 	)
 
 	err := ldapClient.checkServer()
-	err := ldapClient.UpdatePassword("john", "password")
+	require.NoError(t, err)
 
+	err = ldapClient.UpdatePassword("john", "password")
 	require.NoError(t, err)
 }
 
--- a/internal/authentication/ldap_user_provider_test.go
+++ b/internal/authentication/ldap_user_provider_test.go
@@ -677,7 +677,7 @@ func TestShouldUpdateUserPassword(t *testing.T) {
 					},
 				},
 			},
-		}, nil)
+		}, nil),
 
 		mockConn.EXPECT().
 			Search(gomock.Any()).
--- a/internal/authentication/ldap_user_provider_test.go
+++ b/internal/authentication/ldap_user_provider_test.go
@@ -678,8 +678,8 @@ func TestShouldUpdateUserPassword(t *testing.T) {
 				},
 			},
 		}, nil),
-		mockConn.EXPECT()
-			.Close(),
+		mockConn.EXPECT().
+			Close(),
 
 		mockFactory.EXPECT().
 			DialURL(gomock.Eq("ldap://127.0.0.1:389"), gomock.Any()).
--- a/internal/authentication/ldap_user_provider_test.go
+++ b/internal/authentication/ldap_user_provider_test.go
@@ -665,19 +665,19 @@ func TestShouldUpdateUserPassword(t *testing.T) {
 
 		mockConn.EXPECT().
 			Search(NewExtendedSearchRequestMatcher("(objectClass=*)", "", ldap.ScopeBaseObject, ldap.NeverDerefAliases, false, []string{ldapSupportedExtensionAttribute})).
-		Return(&ldap.SearchResult{
-			Entries: []*ldap.Entry{
-				{
-					DN: "",
-					Attributes: []*ldap.EntryAttribute{
-						{
-							Name:   ldapSupportedExtensionAttribute,
-							Values: []string{ldapOIDPasswdModifyExtension},
+			Return(&ldap.SearchResult{
+				Entries: []*ldap.Entry{
+					{
+						DN: "",
+						Attributes: []*ldap.EntryAttribute{
+							{
+								Name:   ldapSupportedExtensionAttribute,
+								Values: []string{ldapOIDPasswdModifyExtension},
+							},
 						},
 					},
 				},
-			},
-		}, nil),
+			}, nil),
 
 		mockFactory.EXPECT().
 			DialURL(gomock.Eq("ldap://127.0.0.1:389"), gomock.Any()).
--- a/internal/authentication/ldap_user_provider_test.go
+++ b/internal/authentication/ldap_user_provider_test.go
@@ -678,7 +678,15 @@ func TestShouldUpdateUserPassword(t *testing.T) {
 				},
 			},
 		}, nil),
+		mockConn.EXPECT()
+			.Close(),
 
+		mockFactory.EXPECT().
+			DialURL(gomock.Eq("ldap://127.0.0.1:389"), gomock.Any()).
+			Return(mockConn, nil),
+		mockConn.EXPECT().
+			Bind(gomock.Eq("cn=admin,dc=example,dc=com"), gomock.Eq("password")).
+			Return(nil),
 		mockConn.EXPECT().
 			Search(gomock.Any()).
 			Return(&ldap.SearchResult{
--- a/internal/authentication/ldap_user_provider_test.go
+++ b/internal/authentication/ldap_user_provider_test.go
@@ -662,6 +662,23 @@ func TestShouldUpdateUserPassword(t *testing.T) {
 		mockConn.EXPECT().
 			Bind(gomock.Eq("cn=admin,dc=example,dc=com"), gomock.Eq("password")).
 			Return(nil),
+
+		mockConn.EXPECT().
+			Search(NewExtendedSearchRequestMatcher("(objectClass=*)", "", ldap.ScopeBaseObject, ldap.NeverDerefAliases, false, []string{ldapSupportedExtensionAttribute})).
+		Return(&ldap.SearchResult{
+			Entries: []*ldap.Entry{
+				{
+					DN: "",
+					Attributes: []*ldap.EntryAttribute{
+						{
+							Name:   ldapSupportedExtensionAttribute,
+							Values: []string{ldapOIDPasswdModifyExtension},
+						},
+					},
+				},
+			},
+		}, nil)
+
 		mockConn.EXPECT().
 			Search(gomock.Any()).
 			Return(&ldap.SearchResult{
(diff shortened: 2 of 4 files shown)


internal/authentication/ldap_connection_factory.go
```diff
@@ -48,6 +49,12 @@
 	return lc.conn.Modify(modifyRequest)
 }
 
+// PasswordModify modifies an ldap objects password
```
autheliabot, line 52: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Comment should end in a period (godot)

james-d-elliott (commented):
Hey thanks for choosing to contribute! Could you explain the use case as well as the result prior to this change and after?

ArsenArsen:
> Hey thanks for choosing to contribute! Could you explain the use case as well as the result prior to this change and after?

Sure thing, prior to this change, password modification happened using the usual replace operation LDAP provides, however, that causes passwords to be stored plaintext. There's two solutions to this, and a third that technically breaks the LDAP standard:
- use PasswordModify, leaving securing the password to the server (the preferred solution, supported by the LDAP specs and any decent implementation, and the solution that's expected to be used)
- hashing the passwords manually in LDAP format, duplicating that work and increasing its load on the admin (since the security parameters would now have to be configured in two places, the LDAP server and Authelia), and
- configuring your LDAP server to always hash your userPassword field, this breaks LDAPs data model technically, and is an unnecessary hack

With this patch applied, the password is stored in hashed form, keeping it secure, in the way LDAP expects it to be done.

davama:
Had a similar issue with authelia storing `userPassword` in a less secure way. Solution was have openldap backend always hash. See [link](https://github.com/authelia/authelia/issues/361#issuecomment-744482335)

Hope it's useful

ArsenArsen:
> Had a similar issue with authelia storing `userPassword` in a less secure way. Solution was have openldap backend always hash. See [link](https://github.com/authelia/authelia/issues/361#issuecomment-744482335)
> 
> Hope it's useful

I already mentioned this solution as the third point in my comment above.

clems4ever:
Hey, thanks for this contribution! It looks good to me. @james-d-elliott , @nightah , would you have an environment with an AD to test against. That way we would be able to include the change in the PR? I don't have any setup to test on my side for now.

> EDIT: I'd also like to add: should a page where a users password can be changed without a password reset be added, the correct way to implement it would be to issue a passwordmodify with oldPassword set to what the user provided.

This is planned once we start implementing the user interface. The change password view is definitely in the backlog.

ArsenArsen:
> This is planned once we start implementing the user interface. The change password view is definitely in the backlog.

Wonderful! I'd just like to, ahead of time, note the potential footgun here: on that form, check whether the user-provided old password field is empty, and if it is, fail; if you were to pass an empty password on, you could just bypass the password comparison, and that's no good!

james-d-elliott:
> Hey, thanks for this contribution! It looks good to me. @james-d-elliott , @nightah , would you have an environment with an AD to test against.

Our integration tests actually test this against SMBAD I believe, however I will manually test it against MSAD.

nightah:
> > Hey, thanks for this contribution! It looks good to me. @james-d-elliott , @nightah , would you have an environment with an AD to test against.
> 
> Our integration tests actually test this against SMBAD I believe, however I will manually test it against MSAD.

Yep @james-d-elliott is spot on. The `ActiveDirectory` integration suite spins up Samba AD and should be a good start.

james-d-elliott:
@ArsenArsen can you please run the following code against your OpenLDAP implementation and report back if it shows the capabilities/controls? We're looking for `1.3.6.1.4.1.4203.1.11.1` which indicates it supports it. After talking with @nightah there are potential implementations that do not support this (ActiveDirectory is one, but we have no idea what other implementations don't support this exop, so we should check it before using it). 

You will need to modify the `conn, err := ldap.DialURL("ldaps://dc.example.com", dialOpt)` line, and the `err = conn.Bind("CN=user,DC=example,DC=com", "password")` line. If nothing comes back you may also need to uncomment the `# attributes = []string{}` line.

```go
package main
import (
	"crypto/tls"
	"fmt"
	
	"github.com/go-ldap/ldap/v3"
)
func main() {
	dialOpt := ldap.DialWithTLSConfig(&tls.Config{InsecureSkipVerify: true})
	conn, err := ldap.DialURL("ldaps://dc.example.com", dialOpt)
	if err != nil {
		panic(err)
	}
	err = conn.Bind("CN=user,DC=example,DC=com", "password")
	if err != nil {
		panic(err)
	}
	baseDN := ""
	scope := ldap.ScopeBaseObject
	deref := ldap.NeverDerefAliases
	sizeLimit := -1
	timeLimit := 30
	typesOnly := false
	filter := "(objectClass=*)"
	attributes := []string{"supportedControl", "supportedCapabilities"}
        # all attributes
        # attributes = []string{}
	controls := []ldap.Control{}
	searchRequst := ldap.NewSearchRequest(baseDN, scope, deref, sizeLimit, timeLimit, typesOnly, filter, attributes, controls)
	result, err := conn.Search(searchRequst)
	if err != nil {
		panic(err)
	}
	for _, entry := range result.Entries {
		for _, attribute := range entry.Attributes {
			fmt.Printf("%s: %v\n", attribute.Name, attribute.Values)
		}
	}
}
```

james-d-elliott:
I believe #2133 will provide a sufficient check of the extension OIDs in order to accurately determine if the client should/should not use passwdModify exop.

ArsenArsen:
```ldif
arsen@ldapserver:~$ ldapsearch -w password -D 'cn=authelia,ou=programs,dc=fh' -b '' -s base supportedControl supportedCapabilities supportedExtension | grep 1.3.6.1.4.1.4203.1.11.1
supportedExtension: 1.3.6.1.4.1.4203.1.11.1
arsen@ldapserver:~$
```

nightah:
> EDIT: I think there should be more than a TRACE message to notify the administrator, since the password would be stored in plaintext, that's quite dangerous

That's not necessarily true. AD for example stores passwords as hashes and not in plaintext and as you've already called out in the previous comments OpenLDAP can also enforce hashing on the attribute if written to in plain text.

Realistically the purpose of #2133 is to be coupled with this change, instead of determining whether the exop should be utilised based on the our `implementation` option it should be based on if it's available.

james-d-elliott:
> 
> 
> ```
> arsen@ldapserver:~$ ldapsearch -w password -D 'cn=authelia,ou=programs,dc=fh' -b '' -s base supportedControl supportedCapabilities supportedExtension | grep 1.3.6.1.4.1.4203.1.11.1
> supportedExtension: 1.3.6.1.4.1.4203.1.11.1
> arsen@ldapserver:~$
> ```
> 
> EDIT: I think there should be more than a TRACE message to notify the administrator, since the password would be stored in plaintext, that's quite dangerous

We can realistically add a log message specifically for instances that are *not* Active Directory and *do not* support passwdModify exop. Thoughts @nightah @ArsenArsen?

nightah:
We could but we also don't know for certain that those instances are not performing their own hashing or definitively storing password in plain text.

I think if the exop is available we should be using it. In instances where it's not available we shouldn't assume said behaviour. Though I'm not averse to warning a user, I'd prefer to not send out unwarranted noise.

ArsenArsen:
> > EDIT: I think there should be more than a TRACE message to notify the administrator, since the password would be stored in plaintext, that's quite dangerous
> 
> That's not necessarily true. AD for example stores passwords as hashes and not in plaintext and as you've already called out in the previous comments OpenLDAP can also enforce hashing on the attribute if written to in plain text.

Yes, but the implementation cannot inform whether this is done, leaving the task to the administrator.
An option to use replace instead of password modify should be added to the config, and an outright failure should happen if that is not set and it is discovered that the exop is not available.

None of this applies to AD, naturally, so that should stay unmodified.

james-d-elliott:
> 
> 
> We could but we also don't know for certain that those instances are not performing their own hashing or definitively storing password in plain text.
> 
> I think if the exop is available we should be using it. In instances where it's not available we shouldn't assume said behaviour. Though I'm not averse to warning a user, I'd prefer to not send out unwarranted noise.

Yep I tend to agree.

nightah:
> Yes, but the implementation cannot inform whether this is done, leaving the task to the administrator.

Exactly, I think there's many tasks that are the responsibility of the administrator.

> An option to use replace instead of password modify should be added to the config, and an outright failure should happen if that is not set and it is discovered that the exop is not available.

Personally I don't agree, I think the onus is on the administrator to make sure that they have made conscious decisions about all the components they're utilising in their stack of which Authelia may be one.

We definitely should take a conservative approach by ensuring we're using the safest methods to integrate with ancillary services, which is why this PR is abosolutely a must. I however do not believe that we should be failing in the instance where the exop isn't available.

Again many LDAP implementations differ and even non-AD ones can perform hashing by default even without the exop and additional configuration like OpenDJ. With so many LDAP implementations out there in the wild a failure would potentially result in many support issues, when realistically it's upto the administrator to understand and make conscious decisions about how to implement their ancillary services.

james-d-elliott:
Hey just letting you know @ArsenArsen that we've merged the other PR and you can implement this using that. Our recommended approach is as follows:

```go
	switch {
	case p.supportExtensionPasswdModify:
		// Use PasswordModifyRequest
	case p.configuration.Implementation == schema.LDAPImplementationActiveDirectory:
		// Use utf16 ModifyRequest
	default:
		// Use standard ModifyRequest
	}
```

Additionally we are going to add a warning log message (warning, but do not fail and continue) on startup anytime the implementation isn't ActiveDirectory and it doesn't support the PasswordModify Extension. This is probably what you meant now that we read the conversation again and had a chance to discuss it as a team. Thanks for bringing it up!

ArsenArsen:
> Hey just letting you know @ArsenArsen that we've merged the other PR and you can implement this using that. Our recommended approach is as follows:

I shall rebase and refactor later today then, thanks!

> Additionally we are going to add a warning log message (warning, but do not fail and continue) on startup anytime the implementation isn't ActiveDirectory and it doesn't support the PasswordModify Extension. This is probably what you meant now that we read the conversation again and had a chance to discuss it as a team. Thanks for bringing it up!

I still think this is not strict enough when dealing with user data :P but as you wish

clems4ever:
> I still think this is not strict enough when dealing with user data :P but as you wish

As far as I'm concerned, I assume that the administrators are always considered involved when they have to configure Authelia and so they need to be mindful and read the doc, particularly when dealing with security components. Configuring an LDAP server properly when installing Authelia is part of the job as much as choosing the right values for session expiration but we don't stop Authelia when the values are off because we cannot really assume the security requirements involved in a setup. Making Authelia fail in the case described above would mean that some informed administrators would be deceived they can not configure their LDAP server the way they want and perhaps in an even more secure way than Authelia is able to provide, who knows.
The main LDAP implementations will not fall in that category anyway so we can assume that the majority should not even care about this problem and if the others do not see the warning in the logs and do not read the manual (because we however need to document that there is a security risk in the doc), then there is not much more we can do for them.

ArsenArsen:
Looks like the testcase should be adjusted to include pwd modify being available. Do we want to include another test and mock for the unavailable case?

james-d-elliott:
> 
> 
> Looks like the testcase should be adjusted to include pwd modify being available. Do we want to include another test and mock for the unavailable case?

You wont need another mock, you'll just need to execute the check server and expect the result correctly, see here: https://github.com/authelia/authelia/blob/master/internal/authentication/ldap_user_provider_test.go#L180-L208

I would just add a single test for each of the 3 scenarios when similar to that test except call  UpdatePassword as well. You can probably just replace that test and the one below it to get what you need.

internal/authentication/ldap_user_provider_test.go
```diff
@@ -659,6 +662,23 @@
...
+					DN: "",
+					Attributes: []*ldap.EntryAttribute{
+						{
+							Name:   ldapSupportedExtensionAttribute,
+							Values: []string{ldapOIDPasswdModifyExtension},
+						},
+					},
+				},
+			},
+		}, nil)
```
autheliabot, line 680: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>missing ',' before newline in argument list (typecheck)

internal/authentication/ldap_user_provider_test.go
```diff
@@ -659,6 +662,23 @@
 		mockConn.EXPECT().
 			Bind(gomock.Eq("cn=admin,dc=example,dc=com"), gomock.Eq("password")).
 			Return(nil),
+
+		mockConn.EXPECT().
+			Search(NewExtendedSearchRequestMatcher("(objectClass=*)", "", ldap.ScopeBaseObject, ldap.NeverDerefAliases, false, []string{ldapSupportedExtensionAttribute})).
+		Return(&ldap.SearchResult{
```
autheliabot, line 668: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>File is not `gofmt`-ed with `-s` (gofmt)

internal/authentication/ldap_user_provider_test.go
```diff
@@ -649,10 +649,38 @@
...
+					Attributes: []*ldap.EntryAttribute{
+						{
+							Name:   ldapSupportedExtensionAttribute,
+							Values: []string{ldapOIDPasswdModifyExtension},
+						},
+					},
+				},
+			},
+		}, nil),
+		mockConn.EXPECT()
```
autheliabot, line 681: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>missing ',' before newline in argument list (typecheck)

internal/authentication/ldap_user_provider_test.go
```diff
@@ -649,10 +649,38 @@
...
+			.Close(),
```
autheliabot, line 682: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>expected operand, found '.' (typecheck)

james-d-elliott (approved):
LGTM, can you merge master into this branch? Then we can merge it back into master.

james-d-elliott:
@ArsenArsen thanks for contributing the code including the tests and security improvements. 

cc @all-contributors

james-d-elliott:
@all-contributors please add @ArsenArsen for code, test, and security.

james-d-elliott:
@all-contributors please add @davama for userTesting

PR #2125, opened 2021-06-28
feat(web): provide autocomplete attributes

dakriy:
Add autocomplete fields to first factor and reset password pages for auto filling.

See:
https://developer.apple.com/documentation/security/password_autofill/enabling_password_autofill_on_an_html_input_element

### DIFF
--- a/web/src/views/LoginPortal/FirstFactor/FirstFactorForm.tsx
+++ b/web/src/views/LoginPortal/FirstFactor/FirstFactorForm.tsx
@@ -95,6 +95,7 @@ const FirstFactorForm = function (props: Props) {
                         onChange={(v) => setUsername(v.target.value)}
                         onFocus={() => setUsernameError(false)}
                         autoCapitalize="none"
+                        autoComplete="username"
                         onKeyPress={(ev) => {
                             if (ev.key === "Enter") {
                                 if (!username.length) {
@@ -124,6 +125,7 @@ const FirstFactorForm = function (props: Props) {
                         onChange={(v) => setPassword(v.target.value)}
                         onFocus={() => setPasswordError(false)}
                         type="password"
+                        autoComplete="current-password"
                         onKeyPress={(ev) => {
                             if (ev.key === "Enter") {
                                 if (!username.length) {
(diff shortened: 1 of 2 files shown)


james-d-elliott (commented):
Hey thanks for choosing to contribute! Looks like a very succinct change. Did you implement this in order to resolve a particular issue though?

For reference, full autocomplete docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

Also we should probably consider this before implementing it: https://portswigger.net/kb/issues/00500800_password-field-with-autocomplete-enabled

dakriy:
Hello!

Yeah, I made this to address a specific issue. I am using Authelia to protect some resources (love the project by the way), and we are implementing a Face ID authentication feature in an iOS application which requires the autocomplete fields to be set correctly. I made the PR as I'd rather not have to maintain a separate custom build of Authelia, and it's a relatively small and easy change.

As for the PCI compliance modification, would you rather that done in this PR or in a separate PR? I am open to adding it into this, although I'm not sure when I'd have the time to make that change, and as such, would probably take a few weeks at least. If I were to make the change (in this PR or in a separate one), I'd want some assistance speccing it out to the projects standards (things such as config naming and options etc).

james-d-elliott:
> 
> 
> Hello!
> 
> Yeah, I made this to address a specific issue. I am using Authelia to protect some resources (love the project by the way), and we are implementing a Face ID authentication feature in an iOS application which requires the autocomplete fields to be set correctly. I made the PR as I'd rather not have to maintain a separate custom build of Authelia, and it's a relatively small and easy change.
> 
> As for the PCI compliance modification, would you rather that done in this PR or in a separate PR? I am open to adding it into this, although I'm not sure when I'd have the time to make that change, and as such, would probably take a few weeks at least. If I were to make the change (in this PR or in a separate one), I'd want some assistance speccing it out to the projects standards (things such as config naming and options etc).

I have to talk to the other maintainers, but thinking we can probably just do it in another PR. I noticed this issue when doing research on autocomplete. Currently we don't set a value, so it's not currently PCI compliant (and we probably are not in other areas). So I think it'll be fine as is, but I'll get back to you.

clems4ever:
I'm fine with doing it in another PR with dedicated specs in it. This PR looks good to me.

clems4ever:
Thanks for the contribution by the way!

james-d-elliott:
Well, perfect timing. I manged to contact them both. We'll merge this, I'll create an issue and assign it to you @dakriy - you're welcome to pick it up in a few weeks when you get time, or I can once  I get through a few other things.

james-d-elliott (approved):

james-d-elliott:
Thanks for your code contributions. Let's add @dakriy for their recent contributions. cc @all-contributors

james-d-elliott:
@all-contributors please add @dakriy for code.

PR #2129, opened 2021-06-30
fix(suites): disable cgo for delve during development

lippl:
following #2101 I was receiving the following error when using the authelia scripts suite for developing
```
[00] # runtime/cgo
[00] cgo: exec gcc: exec: "gcc": executable file not found in $PATH
```
adding the CGO_ENABLED=0 before the dlv build command fixes the issue
must have been overlooked :)


nightah:
Definitely was overlooked @lippl thanks for the fix!

nightah (approved):
LGTM!

PR #2133, opened 2021-07-01
feat(authentiation): check ldap support for extended operations on startup

james-d-elliott:
This PR adds a startup check to the LDAP authentication backend. It additionally adds support for checking supportedExtension OIDs, currently only checking passwdModifyOID (1.3.6.1.4.1.4203.1.11.3). This can relatively easily be enhanced to add detection for other rootDSE capabilities like supportedControl and supportedCapabilities as necessary.


clems4ever (approved):
LGTM

PR #2139, opened 2021-07-02
fix(authentication): use passwdmodify oid instead of whoami oid

james-d-elliott:
This is the correct OID for the passwdModify Extended Operation.


nightah (approved):
LGTM!

PR #2141, opened 2021-07-02
refactor(authentication): log ldap warning on startup in rare condition

james-d-elliott:
This is so on startup administrators who have a LDAP server implementation that may not support password hashing by default are clearly warned. This only triggers if the disable password reset option is not enabled, we cannot find the extension OID for the Extended Password Modify Operation, and the implementation is not Active Directory. Active Directory has it's own method for this which doesn't advertise the extension.

### DIFF
--- a/internal/authentication/ldap_user_provider.go
+++ b/internal/authentication/ldap_user_provider.go
@@ -29,18 +29,19 @@ type LDAPUserProvider struct {
 }
 
 // NewLDAPUserProvider creates a new instance of LDAPUserProvider.
-func NewLDAPUserProvider(configuration schema.LDAPAuthenticationBackendConfiguration, certPool *x509.CertPool) (provider *LDAPUserProvider, err error) {
-	provider = newLDAPUserProvider(configuration, certPool, nil)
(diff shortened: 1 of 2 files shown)


internal/authentication/ldap_user_provider.go
```diff
@@ -29,18 +29,19 @@ type LDAPUserProvider struct {
...
 		return provider, err
 	}
 
-	if provider.supportExtensionPasswdModify {
-		provider.logger.Trace("LDAP Server does support passwdModifyOID Extension")
-	} else {
-		provider.logger.Trace("LDAP Server does not support passwdModifyOID Extension")
+	if !provider.supportExtensionPasswdModify && !configuration.DisableResetPassword &&
+		provider.configuration.Implementation != schema.LDAPImplementationActiveDirectory {
+		provider.logger.Warnf("Your LDAP server implementation may not support a method for password hashing " +
```
clems4ever, line 42: OPT: add a unit test to please codecov.

clems4ever (approved):
LGTM

PR #2171, opened 2021-07-12
feat(oidc): implement client type public

james-d-elliott:
This implements the public option for clients which allows using Authelia as an OpenID Connect Provider for cli applications and SPA's where the client secret cannot be considered secure.


clems4ever (commented):
From a pure code perspective this PR looks good but I'm wondering how it differs from what's described here: https://www.oauth.com/oauth2-servers/single-page-apps/.
I have never seen this public flag anywhere in the OAuth2 documentation so it seems a bit strange to me. Do you have more info about that?

Is this implementation in Fosite the way to implement the implicit grant flow as described here? https://datatracker.ietf.org/doc/html/rfc6749#section-1.3.2

clems4ever:
> From a pure code perspective this PR looks good but I'm wondering how it differs from what's described here: https://www.oauth.com/oauth2-servers/single-page-apps/.
> I have never seen this public flag anywhere in the OAuth2 documentation so it seems a bit strange to me. Do you have more info about that?
> 
> Is this implementation in Fosite the way to implement the implicit grant flow as described here? https://datatracker.ietf.org/doc/html/rfc6749#section-1.3.2

My bad, I found my answers in https://github.com/ory/fosite/issues/78 and https://datatracker.ietf.org/doc/html/rfc6749#section-2.1.

docs/configuration/identity-providers/oidc.md
```diff
@@ -283,6 +300,23 @@ required: yes
...
+<div markdown="1">
+type: bool
+{: .label .label-config .label-purple }
+default: false
+{: .label .label-config .label-blue }
+required: no
+{: .label .label-config .label-green }
+</div>
+
+The public option enables a client where the end user or a malicious application could reasonably obtain the [client id](#id)
```
clems4ever, line 313: I would add the following link https://datatracker.ietf.org/doc/html/rfc6749#section-2.1 somewhere around here to help people understand what public refers to in the RFC.

docs/configuration/identity-providers/oidc.md
```diff
@@ -283,6 +300,23 @@ required: yes
...
+and [client secret](#secret). This is particularly useful for SPA's and CLI tools. This option requires (and allows)
+setting the [client secret](#secret) to a blank string.
+
+In addition to the standard redirect URI requirements, the host must be `localhost` or `127.0.0.1`. Alternatively a
+redirect URI of exactly `urn:ietf:wg:oauth:2.0:oob` is valid.
```
clems4ever, line 318: Can you point to a documentation mentioning this URI? I found some links related to Google using that URI but nothing really described in a RFC.

clems4ever (changes requested):

docs/configuration/identity-providers/oidc.md
james-d-elliott: Basically I had a look at what Dex did in this area. However on closer inspection they allow any redirect URI for public clients now. I think forcing clients which redirect to `localhost`, `127.0.0.*`, or `urn:ietf:wg:oauth:2.0:oob` to be public is longeterm going to be a reasonable safeguard. However upon inspection we should ensure  the OOB URI is not used with the token response type: https://github.com/dexidp/dex/blob/master/server/oauth2.go#L557-L559

I think what we should do for now is:

- allow public clients which is required to support cli tools safely like argo
- prevent localhost redirect uri's on anything other than public clients
- not support the OOB urn

Thoughts @clems4ever ?

docs/configuration/identity-providers/oidc.md
james-d-elliott: Added, thanks for the link!

docs/configuration/identity-providers/oidc.md
clems4ever: So if fosite already does it, let's trust them imo. It's better that the RFC implementation remains fosite duty altogether.

clems4ever: But we can keep some part of the documentation regarding the fact that only OOB should be used though.

clems4ever (dismissed):
Adapt the doc according to the last comment I left and we are good to go from my point of view.

PR #2178, opened 2021-07-13
feat(authentication): ldap user groups attribute searching

james-d-elliott:
This allows administrators to configure an attribute that exists on user objects which contains a list of distinguishedName's of the users group memberships.

### DIFF
--- a/internal/configuration/validator/authentication.go
+++ b/internal/configuration/validator/authentication.go
@@ -120,6 +120,8 @@ func validateLDAPAuthenticationBackend(configuration *schema.LDAPAuthenticationB
 		setDefaultImplementationCustomLDAPAuthenticationBackend(configuration)
 	case schema.LDAPImplementationActiveDirectory:
 		setDefaultImplementationActiveDirectoryLDAPAuthenticationBackend(configuration)
+	case schema.LDAPImplementationFreeIPA:
+		setDefaultImplementationFreeIPALDAPAuthenticationBackend(configuration)
 	default:
 		validator.Push(fmt.Errorf("authentication backend ldap implementation must be blank or one of the following values `%s`, `%s`", schema.LDAPImplementationCustom, schema.LDAPImplementationActiveDirectory))
 	}
@@ -173,7 +175,7 @@ func validateLDAPURL(ldapURL string, validator *schema.StructValidator) (finalUR
 	return parsedURL.String(), parsedURL.Hostname()
 }
 
-func validateLDAPRequiredParameters(configuration *schema.LDAPAuthenticationBackendConfiguration, validator *schema.StructValidator) {
+func validateLDAPRequiredParameters(configuration *schema.LDAPAuthenticationBackendConfiguration, validator *schema.StructValidator) { //nolint: gocyclo
 	// TODO: see if it's possible to disable this check if disable_reset_password is set and when anonymous/user binding is supported (#101 and #387)
 	if configuration.User == "" {
 		validator.Push(errors.New("Please provide a user name to connect to the LDAP server"))
@@ -207,36 +209,13 @@ func validateLDAPRequiredParameters(configuration *schema.LDAPAuthenticationBack
 		}
 	}
 
-	if configuration.GroupsFilter == "" {
+	switch {
+	case configuration.GroupsFilter != "" && configuration.GroupsAttribute != "":
+		validator.Push(fmt.Errorf(errFmtLDAPBothGroupsFilterAndGroupsAttributeSet))
+	case configuration.GroupsFilter == "" && configuration.GroupsAttribute == "":
 		validator.Push(errors.New("Please provide a groups filter with `groups_filter` attribute"))
-	} else if !strings.HasPrefix(configuration.GroupsFilter, "(") || !strings.HasSuffix(configuration.GroupsFilter, ")") {
+	case configuration.GroupsFilter != "" && (!strings.HasPrefix(configuration.GroupsFilter, "(") || !strings.HasSuffix(configuration.GroupsFilter, ")")):
 		validator.Push(errors.New("The groups filter should contain enclosing parenthesis. For instance cn={input} should be (cn={input})"))
-	}
-}
-
-func setDefaultImplementationActiveDirectoryLDAPAuthenticationBackend(configuration *schema.LDAPAuthenticationBackendConfiguration) {
-	if configuration.UsersFilter == "" {
-		configuration.UsersFilter = schema.DefaultLDAPAuthenticationBackendImplementationActiveDirectoryConfiguration.UsersFilter
-	}
-
-	if configuration.UsernameAttribute == "" {
-		configuration.UsernameAttribute = schema.DefaultLDAPAuthenticationBackendImplementationActiveDirectoryConfiguration.UsernameAttribute
-	}
-
-	if configuration.DisplayNameAttribute == "" {
-		configuration.DisplayNameAttribute = schema.DefaultLDAPAuthenticationBackendImplementationActiveDirectoryConfiguration.DisplayNameAttribute
-	}
-
-	if configuration.MailAttribute == "" {
-		configuration.MailAttribute = schema.DefaultLDAPAuthenticationBackendImplementationActiveDirectoryConfiguration.MailAttribute
-	}
-
-	if configuration.GroupsFilter == "" {
-		configuration.GroupsFilter = schema.DefaultLDAPAuthenticationBackendImplementationActiveDirectoryConfiguration.GroupsFilter
-	}
-
-	if configuration.GroupNameAttribute == "" {
-		configuration.GroupNameAttribute = schema.DefaultLDAPAuthenticationBackendImplementationActiveDirectoryConfiguration.GroupNameAttribute
 	}
 }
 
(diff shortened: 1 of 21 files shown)


internal/configuration/validator/authentication.go
```diff
@@ -120,6 +120,8 @@ func validateLDAPAuthenticationBackend(configuration *schema.LDAPAuthenticationB
 		setDefaultImplementationCustomLDAPAuthenticationBackend(configuration)
 	case schema.LDAPImplementationActiveDirectory:
 		setDefaultImplementationActiveDirectoryLDAPAuthenticationBackend(configuration)
+	case schema.LDAPImplementationFreeIPA:
+		setDefaultImplementationFreeIPALDAPAuthenticationBackend(configuration)
 	default:
 		validator.Push(fmt.Errorf("authentication backend ldap implementation must be blank or one of the following values `%s`, `%s`", schema.LDAPImplementationCustom, schema.LDAPImplementationActiveDirectory))
```
Alexsaphir, line 126: Add this option too `schema.LDAPImplementationFreeIPA`

Alexsaphir: Sorry it's for the line 126

james-d-elliott: Nice spot!

Alexsaphir:
So for now the log in is working.
I'm having issue with the group membership. I may not have understand your way.

FreeIPA log give me the following search : 

```
SRCH base="cn=groups,cn=accounts,dc=alexsaphir,dc=com" 
scope=2 
filter="(|(distinguishedName=cn=ipausers,cn=groups,cn=accounts,dc=alexsaphir,dc=com)...
```
I think that you wanted to replace distinguishedName by it's value, but i don't think LDAP support a filter with the dn as attribute. It could be replaced by an empty base, a filter cn=..., dn=...

Could you quickly explain to me what you do for the group membership, because I'm not fluent in Go to be able to follow your thought.

Thanks

james-d-elliott:
> 
> 
> So for now the log in is working.
> I'm having issue with the group membership. I may not have understand your way.
> 
> FreeIPA log give me the following search :
> 
> ```
> SRCH base="cn=groups,cn=accounts,dc=alexsaphir,dc=com" 
> scope=2 
> filter="(|(distinguishedName=cn=ipausers,cn=groups,cn=accounts,dc=alexsaphir,dc=com)...
> ```
> 
> I think that you wanted to replace distinguishedName by it's value, but i don't think LDAP support a filter with the dn as attribute. It could be replaced by an empty base, a filter cn=..., dn=...
> 
> Could you quickly explain to me what you do for the group membership, because I'm not fluent in Go to be able to follow your thought.
> 
> Thanks

distinguishedName comes from this new config option (which you can override the default setting of): https://github.com/authelia/authelia/pull/2178/files#diff-9e2351f62725f17ebbfe9b0db5b6510f234ae1f25fbc641a80f2c767b1695397R215

Which should be `dn` if you use `implementation: freeipa` as per here: https://github.com/authelia/authelia/pull/2178/files#diff-c1278aed7954b018e82841f8e1831f454e44a03977785d380f85725d5a5995cfR105

Alexsaphir:
One of the issue is that distinguishedName is not replaced by it's value.

If I understand : 
_At login :_
1- _Check if username/mail and password are valid_
2- _Query memberOf attributes_
3- _ For each entry fetch the 'short name'(cn, for FreeIPA) who will be used in the config file and store the result.

_When accessing a block with group dependency :_
A-_[Optional] Update the info by redoing step 2,3_
B-_Check if the user is in the group_

It's at step 3 that you use distinguishedName, but I can't reproduce what you do with Active Directory.However, for OpenLDAP/FreeIPA, by settings base to the value of the memberOf attributes and using a filter to ensure it's group we can query the cn attribute.
(One issue is that we do not check if the group is taken in the correct part of the tree, so maybe just extracting the cn from the string obtained by memberOf and doing some substring comparaison would be more easy).

james-d-elliott:
@Alexsaphir can you show me your ldap config minus any sensitive information? I can confirm via a test the replacement occurs if configured correctly.

Alexsaphir:
```
  ldap:
    # implementation: custom
    implementation: freeipa
    groups_attribute: memberOf
    distinguished_name_attribute: memberOf



    url: ldap://freeipa
    start_tls: false
    base_dn: dc=alexsaphir,dc=com

    username_attribute: uid
    mail_attribute: mail
    display_name_attribute: displayname

    additional_users_dn: cn=users,cn=accounts
    users_filter: (&({username_attribute}={input})(objectClass=person))

    additional_groups_dn: cn=groups,cn=accounts
    group_name_attribute: cn
    # groups_filter: (&(member=uid={input},cn=users,cn=accounts,dc=alexsaphir,dc=com)(objectclass=groupofnames))

    user: uid=admin,cn=users,cn=accounts,dc=alexsaphir,dc=com

```

james-d-elliott:
Not entirely sure how that's working, I forgot to allow distinguished_name_attribute to the valid keys list.. That being said the default for freeipa is `dn`, if you turn trace logging on when the latest commit builds you will be able to see the filter we're calculating in the logs.

james-d-elliott:
I can confirm, the default value for the distinguished name replacement is `dn` in freeipa, Though your results are really strange.. as mentioned I did make a mistake fixed by 7a3f310 but that shouldn't prevent the freeipa default working.. also your authelia should fail to start without that fix when you have `distinguished_name_attribute` in your config.

Alexsaphir:
It's working now. But previously authelia didn't failed to start

james-d-elliott:
Nice, are you using docker or something else?

Alexsaphir:
Yes using Docker. Having some weird results, need to check more

james-d-elliott:
Nice, are you using docker or something else?

> 
> 
> Yes using Docker. Having some weird results, need to check more
> 
> EDIT : It seem not all groups are fetched. Manual ldap query report more entry than authelia

Are you using the same search base? Based on your configuration Authelia's search base will be `cn=groups,cn=accounts,dc=alexsaphir,dc=com`, and it will not check nested groups. If you provide a LDAP query that does nested searches with freeipa I can experiment with a method to fix that.

Alexsaphir:
Freeipa doesnt support nested search when requesting `member` of a group. But using `memberOf` of a user we get all groups. 

Moreover, the use of a dn in a filter don't seem to be supported. Why not just append `group_name_attribute` with the group in the conf file, `additional_groups_dn` and `base_dn`.

And check with base = `additional_users_dn` + `base_dn`
filter=`(&({username_attribute}={input})(memberOf={dn_of_the_group}))`
attrs=`cn` 

With that we just need to check that the response is not empty.

And I think there is an issue because when set to dn, `distinguished_name_attribute` is still replaced by `distinguishedName`. But I don't really think the need for this attribute.

james-d-elliott:
> we don't really need to perform query on the group tree like the default implementation

I'm not sure all users will be happy with the distinguished name or a portion of the distinguished nameas the value returned by a group, and as far as I can tell there is no way to get additional information about a DN without doing the query, am I mistaken?

Alexsaphir:
For example in the conf file, let say it's says `"group: admins, superuser"`.

With the existing settings in the ldap conf you can write the dn for each group.
Then perform an query to the LDAP server where base is the user base, and a filter like that:

``` (&(uid={username})(|(memberOf={dn_group1})(memberOf={dn_group2})))```

Ask only for one attribute like uid and check if the respons is empty or not.

Alexsaphir:
I read the code a litte more and understand that the list of string Profile.Groups should match the groups name use in the conf file. 

Then, a step need to be perform when adding to the group list to only get the cn of the groups.
It could be something like that
[I do not think it's possible with LDAP to only get a rdn from a  dn]

1- Length of the dn greater than p.groupsBaseDN
2- Begin by the group_name_attribute and end by p.groupsBaseDN
3- Extract the substring and append it to the list

It allow to enforce the p.groupsBaseDN.

Alexsaphir:
Wait I I think I've made a mistake. When I said recursive I mean as a group is a member of another group not like a group is a sub entry of one. (I don't think memberOf works for the second case, FreeIPA use the first case)

james-d-elliott:
> Then, a step need to be perform when adding to the group list to only get the cn of the groups.

Or the attribute the admin configures, it must be a search.

> With the existing settings in the ldap conf you can write the dn for each group.
> Then perform an query to the LDAP server where base is the user base,

The base for group searches is configurable by the admin, it's `base_dn` plus `additional_groups_dn`, 

> And I think there is an issue because when set to dn, `distinguished_name_attribute` is still replaced by `distinguishedName`. But I don't really think the need for this attribute.

I mean potentially, we need to do a search based on the distinguished name attribute though, currently this is only for freeipa, but as soon as we move to other LDAP systems that may not be the value. 

> Wait I I think I've made a mistake. When I said recursive I mean as a group is a member of another group not like a group is a sub entry of one. (I don't think memberOf works for the second case, FreeIPA use the first case)

Yeah for recursive searching, we can either loop through searches by checking the memberOf attribute of each group, or we can trust that the memberOf has the recursive groups. Based on your earlier messages I am guessing the later is not true, though I'd be interested in what LDAP query you use to produce a different list of groups.

Alexsaphir:
Sorry for the delay. I've tried to provide a implementation of what I was trying to explain.
It's badly done, but can help explain what I was trying to say.

https://github.com/Alexsaphir/authelia/commit/768a6f58f93e217f33103ae63b441912e1f830c0

`profile.Groups` store directly the group that authelia use, i remove the `GroupNameAttribute` and `groupsBaseDN` to obtain it with some check

Like you can see,I only need one query to the user to obtain the user data and the group memberships.

Alexsaphir:
@james-d-elliott  I've improved my implementation and obtain valid results using my existing FreeIPA installation.

https://github.com/Alexsaphir/authelia/commit/6a771bf9487e64f4a324baf2881bbdbb33798d25

james-d-elliott:
@Alexsaphir I don't think this approach is a good one. As an alternative I'd suggest we instead either just use the dn itself, or we use the groups filter to periodically fetch all groups and store them in a map[string]string, with the distinguishedName as the key, and the name attribute as the value.

I can dedicate some time to this during the week/weekend. If you were able to hop on Matrix/Discord that'd help. I've never used FreeIPA but will be setting it up to implement this appropriately.

PR #2179, opened 2021-07-13
fix(authentication): ldap connection left open

james-d-elliott:
The recent ldap changes in cb71df5 left a connection to the LDAP server open at startup. This resolves this which prevents an ugly log message and unnecessary open sockets.


nightah (approved):
LGTM.

PR #2181, opened 2021-07-14
perf(authentication): improve active directory default users filter

james-d-elliott:
This adds a performance change to the default Active Directory users filter. Basically as per TechNet the (sAMAccountType=805306368) filter is the same as (&(objectCategory=person)(objectClass=user)) except the performance is better.


nightah (approved):
LGTM.

I assume this works both for SambaAD and MSAD?

james-d-elliott:
> 
> 
> LGTM.
> 
> I assume this works both for SambaAD and MSAD?

I think so, let me check with 9e74a5b @nightah

PR #2184, opened 2021-07-14
fix(handlers): send status 303 auth requests that are not get/head

### LINKED ISSUE #2173: Automatic Logout Page - Inactivity/Session Expired
### Description

Display a webpage with a message " You have been automatically logged out" and an option to sign in again. Triggered by session expiry, cookie delete, etc.

### Use Case

I find myself using the webpage of an service I am hosting without knowing my session has expired until I click on a link and the page does not respond. I have to then manually refresh to login again. 
A service such as Grafana allows editing and moving of dashboard elements even though the session has unknowingly expired resulting in futile work.

This feature would prevent accidental interaction with an 'expired' webpage.

james-d-elliott:
When a request occurs, if the browser is not performing a HTTP GET/HEAD request, the 302 status code is not valid. This commit resolves this. MDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302. This PR changes the behaviour, so that if the request method is blank, `GET`, `HEAD`; the response code is `302`, otherwise it is `303`.

### DIFF
--- a/internal/handlers/handler_verify.go
+++ b/internal/handlers/handler_verify.go
@@ -228,8 +228,15 @@ func handleUnauthorized(ctx *middlewares.AutheliaCtx, targetURL fmt.Stringer, is
 		}
 
 		ctx.Logger.Infof("Access to %s (method %s) is not authorized to user %s, redirecting to %s", targetURL.String(), friendlyMethod, friendlyUsername, redirectionURL)
-		ctx.Redirect(redirectionURL, 302)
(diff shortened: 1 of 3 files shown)


clems4ever (changes requested):
LGTM but an additional test in the suites would probably be a good idea. There is already one testing the redirect done by /verify if I remember well.

clems4ever (approved):
LGTM

PR #2199, opened 2021-07-23
fix(examples): change config in example projects to solve startup issues

### LINKED ISSUE #2198: Minor configuration issues in lite and local deployments
I encountered a few errors while setting up the lite and local deployments on an Amazon Linux 2 instance. I think these changes may help with onboarding:

### Lite deployment
- Use sudo for docker in `setup.sh`
- Check if docker-compose succeeded before showing success message in `setup.sh`
- Use `log_level` until docker image supports new `log.level` key in `authelia/configuration.yml` (solves "config key not expected: log.level" error)

### Local deployment
- Fix casing in password hint in `authelia/users_database.yml`
- Mount `./traefik` rather than `./traefik/acme.json` to solve access and permission issues for that file
- Make authelia hostname consistent with local example project in `docker-compose.yml`
- Use `log_level` until docker image supports new `log.level` key in `authelia/configuration.yml` (solves "config key not expected: log.level" error)

hoflogic:
Fixes #2198.


examples/compose/lite/authelia/configuration.yml
```diff
@@ -5,8 +5,7 @@
 
 host: 0.0.0.0
 port: 9091
-log:
-  level: debug
+log_level: debug
```
james-d-elliott, line 8:
```suggestion
log:
  level: debug
```

Please remove this change. The newest version of Authelia will be deprecating the old configuration method. We should instead make sure the documentation for the lite bundle has people checkout the latest tagged version of the repository. See an example here: https://www.authelia.com/docs/getting-started.html#steps

examples/compose/local/authelia/configuration.yml
```diff
@@ -5,8 +5,7 @@
 
 host: 0.0.0.0
 port: 9091
-log:
-  level: debug
+log_level: debug
```
james-d-elliott, line 8:
```suggestion
log:
  level: debug
```

Same as above.

james-d-elliott (changes requested):
A few notes. Also interested in input from @nightah regarding

hoflogic:
Regarding the request to use sudo for docker, the documentation for docker states that sudo is required by default and adding users to the docker group appears (in my opinion) to be slightly discouraged or approached with caution. For that reason I felt that using sudo would reduce onboarding issues for a wider group of people.

See: https://docs.docker.com/engine/install/linux-postinstall/

examples/compose/local/setup.sh
```diff
@@ -110,7 +110,12 @@ else
   password
 fi
 
-docker-compose up -d
+sudo docker-compose up -d
+
+if [[ $? != 0 ]]; then
+  echo "docker-compose encountered an error"
```
nightah, line 116: Is this actually necessary if `docker-compose` bails out with an error it will have whatever the output of said meaningful error is presented to the user.

nightah (dismissed):
I think the use of `sudo` is fine, given that we're asking for permissions initially in the script it's not a huge deal to also use sudo to run up the containers.

Just one small nit about the error message, do we really need it there?

examples/compose/local/setup.sh
hoflogic: I agree that could be removed.

nightah:
@hoflogic I'm having trouble merging this and I think it's due to a GH bug.

Can you open another PR (don't close this one yet) with the same changes but from a new branch of your fork instead of `master`?

hoflogic:
Sure, would you like me to squash the commits in the new branch while I'm at it or keep the history as-is?

nightah:
Keeping history is fine we'll squash on merge to `master`.
You could clean clone our master and then cherrypick your commits to avoid all the merge commits.

nightah (approved):
LGTM!

nightah:
@hoflogic it's probably worthwhile considering signing your commits in future; https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification/about-commit-signature-verification.

We require signed commits for the `master` branch which is what was preventing me from merging earlier.

PR #2200, opened 2021-07-23
fix(notifier): remove SMTPS warning

### LINKED ISSUE #2194: (Understandable) Misinformation about using SMTP port 465 for notifications in the logs
You may know me from the current OIDC documentation updates. I'm also member of the core maintainers team at [docker-mailserver/docker-mailserver](https://github.com/docker-mailserver/docker-mailserver).

## The Issue at Hand

I was configuring Authelia's notifier settings and stumbled upon this error message

``` INI
level=warning msg="Notifier SMTP client configured to connect to a SMTPS server. It's highly recommended you use a non SMTPS port and STARTTLS instead of SMTPS, as the protocol is long deprecated."
``` 

While this did not bother me in the beginning, I figured I should report this.

## Rationale

First of all, 465 _was_ SMTPS, but isn't anymore. It is now _submission**s**_.

There is **a lot** of misinformation about port 465 on the internet. I'd **highly suggest to read [this section from our mail server documentation](https://docker-mailserver.github.io/docker-mailserver/edge/config/security/understanding-the-ports/)**. It states that port 465 is "a submission port since 2018".

To get the full view, one should read

1. [RFC 8314 Section 7.3](https://datatracker.ietf.org/doc/html/rfc8314#section-7.3) 
2. [RFC 8314 Section 3.3](https://datatracker.ietf.org/doc/html/rfc8314#section-3.3)
3. [RFC 8314 Section 4.1](https://datatracker.ietf.org/doc/html/rfc8314#section-4.1)

TL;DR 465 is recommended over 587 as it communicates the use of TLS more clearly. If configured correctly, there is no security advantage of 465 over 587, but with 465, TLS is _mandatory_, and this is the advantage. The message in the logs is therefore not accurate (outdated ?).

georglauterbach:
Closes #2194


georglauterbach:
I have no experience with CI/tests in Go. Maybe you can hint to where I can test the line properly (since Codecov was complaining).

nightah:
The errors are unrelated to your changes, no need to worry about the codecov checks.

james-d-elliott (approved):
LGTM

PR #2213, opened 2021-07-30
docs: update and unify contact options

james-d-elliott:
This updates and unifies the contact options so it is easier to maintain. All contact options now link back to one of two locations, and both of these locations are a copy and paste for the most part.


nightah (approved):
LGTM.

PR #2220, opened 2021-08-02
ci(buildkite): skip integration test stage for examples

james-d-elliott:
This skips integration tests for any commits/PR's that only make changes to the examples/ folder as it's not a change that requires this check.


nightah (approved):
LGTM.

PR #2221, opened 2021-08-02
fix(examples): change config in example projects to solve startup issues

hoflogic:
See #2199 and #2198. Changes cherry-picked to `fix-example-configuration` branch.


nightah:
Thanks @hoflogic looks like it wasn't a bug, I'll merge the other one in and explain in the other PR.

PR #2224, opened 2021-08-02
test(web): drop enzyme in favor of react-testing-library

nightah:
Enzyme is falling behind in maintenance, it is currently maintained by one primary developer and still [does not support React 17](https://github.com/enzymejs/enzyme/pull/2430) despite it being released in October 2020.

[react-testing-library](https://testing-library.com/docs) (RTL) is [recommended by Facebook](https://reactjs.org/docs/test-utils.html#overview) and encourages writing tests that avoid testing implementation details.


james-d-elliott (approved):
LGTM

PR #2239, opened 2021-08-04
perf(authentication): improve ldap dynamic replacement performance

james-d-elliott:
This change means we only check the filters for the existence of placeholders that cannot be replaced at startup. We then utilized cached results of that lookup for subsequent replacements.


nightah (approved):
LGTM.

PR #2240, opened 2021-08-05
feat(authentication): allow customizable ldap connection timeout

james-d-elliott:
This implements both a connection timeout for LDAP connections, and makes it configurable by administrators. The default is 5s. The reason for this commit is currently if a connection to an LDAP server cannot be established it does not timeout in a reasonable timeframe.


nightah (approved):
LGTM.

PR #2242, opened 2021-08-05
ci(buildkite): make test concurrency gate conditional

nightah:
This ensures that the test concurrency gate step does not run on tagged releases are the integration steps are also skipped.


james-d-elliott (approved):
LGTM

PR #2243, opened 2021-08-05
fix(cmd): remove microbadger references

nightah:
MicroBadger is dead, so we need to remove all references of it.


james-d-elliott (approved):
LGTM

PR #2246, opened 2021-08-05
ci(buildkite): allow retry on successful docker deploy steps

nightah:
Occasionally during a manifest deployment tags can be removed and the step may fail. To ensure the manifest step can be completed successfully it would require re-pushing the tags that had been removed.

Turning on the `permit_on_passed` option allows us to control this all through the Buildkite interface as opposed to manual intervention.


clems4ever (approved):
LGTM

PR #2248, opened 2021-08-05
fix(notifier): regression in startup check

james-d-elliott:
The changes in the koanf commit a7e867a699a77f0b9d564defd81094fd410e4404 introduced a regression in the startup procedure that prevented the option to disable the startup check from having an effect. This ensures the startup check is not performed if it is disabled as it was intended.


nightah (approved):
LGTM.

PR #2250, opened 2021-08-05
fix: healthcheck always returns healthy

james-d-elliott:
The healthcheck will always return 0 (healthy) since we don't source the env file before checking the ENV var.


nightah (approved):
LGTM.

PR #2252, opened 2021-08-05
fix(configuration): ldap timeout option not recognized as valid

james-d-elliott:
This is so an unexpected error doesn't occur when someone uses the new ldap timeout key which we missed adding to the list of valid keys in b2a49e1780041943862590984070a2268a87d823.


nightah (approved):
LGTM.

PR #2267, opened 2021-08-07
ci(buildkite): allow manual retry on successful steps

nightah:
This permits manual retry on specific steps which can cause problematic issues for example when a node runs out of disk space.

By allowing this we should be able to recover problematic builds instead of forcing a complete rebuild which may be undesirable on the `master` or other production branches.


james-d-elliott (approved):
LGTM

PR #2268, opened 2021-08-07
fix(configuration): make notifier logging consistent and more accurate

james-d-elliott:
This ensures the notifier logs are more accurate to give people a clear picture of if they either have no notifier specified or multiple.


nightah (approved):
LGTM.

PR #2257, opened 2021-08-06
fix(handlers): ensure server path is appended to the oidc issuer

james-d-elliott:
This ensures the server path is appended to the oidc well-known issuer information.

PR #2272, opened 2021-08-09
fix: oidc issuer path and strip path middleware

### LINKED ISSUE #2260: OIDC: Path issues
### Description

As discussed in chat, there are some issues with OIDC when using the path middleware 

### Expected Behaviour

_N/A_

### Reproduction Steps

_N/A_

### Additional Information

_**Note 1 (james-d-elliott):** Currently the only known fix is to configure Authelia and NGINX to use subdomain configuration. This will work for the user in question, but other users will experience issues in the future. It would be good if we fix this to get as much information from Bradley as possible in order to replicate it. I believe currently he's using a fairly stock swag configuration, with the path set to `authelia`. We should make sure in his case that the sections we change are commented out and replaced so reverting to his old configuration if he desires is a very simple adjustment._

nightah:
This ensures the server.path requests append the base_url to the oidc well-known issuer information and adjusts server.path configuration to only strip the configured path instead of the first level entirely, regardless of its content.

Co-authored-by: James Elliott <james-d-elliott@users.noreply.github.com>

Fixes #2260, Closes #2257, Closes #2258.

### DIFF
--- a/internal/handlers/const.go
+++ b/internal/handlers/const.go
@@ -60,6 +60,8 @@ const (
 
 // OIDC constants.
 const (
+	pathOpenIDConnectWellKnown = "/.well-known/openid-configuration"
+
 	pathOpenIDConnectJWKs          = "/api/oidc/jwks"
 	pathOpenIDConnectAuthorization = "/api/oidc/authorize"
 	pathOpenIDConnectToken         = "/api/oidc/token" //nolint:gosec // This is not a hard coded credential, it's a path.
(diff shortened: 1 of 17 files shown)


james-d-elliott (approved):
LGTM

james-d-elliott (changes requested):
I think we still have to change `handleOIDCWorkflowResponse` to use ctx.GetExternalRootURL?

james-d-elliott (approved):
LGTM

BCNelson:
This Worked for me and fixed #2260

PR #2298, opened 2021-08-18
docs: add nextcloud community tested oidc apps

nreymundo:
Heya! Just did some trial and error until I managed to get Nextcloud to auth through OpenID Connect with Authelia and figured I'd post the callback URL I used to get it working. 

I put the latest stable nextcloud version (currently `22.1.0`) as the minimum version because that's what I have at hand to test on but since the OIDC login stuff actually depends on an app from their app store and seems to be available going back several versions it might very well work just the same with older installations. 

Let me know if you need any more info or if there are any tests/best-practices/whatever I should follow with the PR. 

Thanks for the awesome work on this :D.


james-d-elliott (approved):
LGTM

james-d-elliott:
@all-contributors please add @nreymundo for docs

PR #2309, opened 2021-08-22
refactor(web): eslint glob pattern

nightah:
The current eslint glob pattern is not Windows friendly, this change works for both Linux and Windows.


james-d-elliott (approved):
LGTM

PR #2345, opened 2021-09-06
fix(suites): prevent dev workflow overriding .healthcheck.env

nightah:
Due to how the dev workflow allows Authelia developers to run the binary hot-reloaded we should disable the Docker health check for said workflow to prevent accidental commits upstream.


james-d-elliott (approved):
LGTM

PR #2348, opened 2021-09-07
make it easier to bootstrap (defer to filesystem notification as opposed to smtp)

johndpope:


nightah:
If you have a look at our [Getting Started docs](https://www.authelia.com/docs/getting-started.html) the local examples utilise filesystem for the notifier.

Anything beyond local is actually intended as a Production type deployment where multiple users would interact with said platform, this needs an actual SMTP service backing Authelia to work appropriately.

PR #2352, opened 2021-09-08
ci: add husky with pre-commit and commit-msg hooks

nightah:
This change includes two new hooks as part of our GitHub workflow with husky:

* `pre-commit`: Performs linting with golangci-lint and eslint/prettier
* `commit-msg`: Ensures that the commit messages conform to our guidelines and will error and provide context to a user when they do not.

The `prepare` command which has been included is executed each time a `yarn install` is executed.


james-d-elliott (approved):
LGTM

Issue #2369, opened 2021-09-12
Authelia on Debian11: level=warning msg="Error reading hosts memory limit: open /sys/fs/cgroup/memory/memory.limit_in_bytes: no such file or directory"


Tropaion:
Hello,
I'm setting up authelia on Debian11 with Nginx Proxy Manager and everything is fine except for this warning:
level=warning msg="Error reading hosts memory limit: open /sys/fs/cgroup/memory/memory.limit_in_bytes: no such file or directory"

I suppose this has something to do with Debian11?
Is there a way to fix this?

nightah:
Thanks for the report @Tropaion, the warning isn't anything to be worried about. We perform a startup check to ensure that the memory allocated for password hashing isn't grossly misconfigured.

The error is specifically to do with many OSes now opting to CGroupsv2 as the default instead of CGroups v1.
The method we use to check the maximum memory on the host looks like it's only available in CGroups v1 and doesn't have a comparable alternative in CGroups v2. The `memory.max` alternative in CGroups v2 only returns `max` as the default and a value if memory limits have been set on the respective container.

@james-d-elliott @clems4ever we originally put that warning in there when we had a much higher memory default (1GB), given that we have dropped the default down significantly I think it's safe to remove the warning. Thoughts?

clems4ever:
I'm fine with removing it.

PR #2380, opened 2021-09-16
fix(server): remove obselete memory check warning

nightah:
Given the fact that many Linux OSes are defaulting to CGroups v2 and also Authelia changing the default memory config for argon2id this warning is now obselete.

Fixes: #2369.


james-d-elliott (approved):
LGTM

PR #2381, opened 2021-09-16
feat: builds with gox and buildx

nightah:
This change builds all of Authelia respective binaries in parallel within a single step and distributes as necessary to subsequent steps, we now also build and distribute for the following OS/Architecture:

* `freebsd/amd64`

Our CI/CD pipeline now also utilises docker buildx as a default for builds and pushes.


james-d-elliott (approved):
LGTM

PR #2403, opened 2021-09-21
Docs trusted proxies

P0KeT:


james-d-elliott:
This will be handled in the already open PR, there's no need to open a new one and merge it. I still need to check haproxy and nginx.

PR #2410, opened 2021-09-22
feat: hardened authelia binaries

nightah:
This change ensures that all Authelia binaries which are compiled and distributed are hardened with the following standards:

* RELRO
* Stack canary
* NX
* PIE/ASLR
* Stripped RPATH AND RUNPATH
* Stripped Symbols
* Fortify

The musl variants currently [do not support Fortify](https://wiki.musl-libc.org/future-ideas.html#Fortify).


son3ofSatan:
hi lady

james-d-elliott (approved):
LGTM

PR #2422, opened 2021-09-28
feat(web): replace babel-loader and terser with esbuild

nightah:
This PR allows us to utilise [esbuild](https://esbuild.github.io/) for compilation and bundling as opposed to babel and terser respectively.

This allows faster builds both during production bundling and during development.


james-d-elliott (approved):
LGTM

PR #2427, opened 2021-09-30
refactor: replace sqlite library with the original cgo variant

nightah:
This change reverts our SQLite library back to the original for compatibility and performance reasons, given that we now always package with CGO.


james-d-elliott (approved):
LGTM

PR #2459, opened 2021-10-06
docs: activedirectory groups filter example

jacobschwan:
Modify groups filter example for activedirectory to pull in all groups specified user belongs to, including due to group nesting, and not just groups the user is directly assigned to.

### DIFF
--- a/docs/configuration/authentication/ldap.md
+++ b/docs/configuration/authentication/ldap.md
@@ -226,7 +226,7 @@ makes sure that value is not 0 which means the password requires changing at the
 |Implementation |Users Filter  |Groups Filter|
 |:-------------:|:------------:|:-----------:|
 |custom         |n/a           |n/a       |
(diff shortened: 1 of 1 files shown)


james-d-elliott:
If you'd like, you can alter the groups_filter section to something similar to this since I've since then found documentation for Samba AD indicating it doesn't support it.

```md
### groups_filter
Similar to [users_filter](#users_filter) but it applies to group searches. 

#### Transitive group membership
Authelia doesn't natively support searching transitive group membership. Some LDAP implementations however
may support a filter that does this. In order to do this with Microsoft Active Directory (Samba Active Directory at the
time of this writing doesn't appear to support it), you may use the `memberOf:1.2.840.113556.1.4.1941` filter, similar
to the example below:

`(&(member:1.2.840.113556.1.4.1941:={dn})(objectClass=group)(objectCategory=group))`
```

PR #2534, opened 2021-10-25
refactor(suites): replace selenium with go-rod

nightah:
This change replaces [tebeka/selenium](https://github.com/tebeka/selenium) with [go-rod](https://github.com/go-rod/rod).

We no longer have a Chromedriver/external driver dependency to utilise Selenium as we instead utilise the Chrome Devtools Protocol to communicate with the browser.

Rod [documents](https://go-rod.github.io/#/why-rod) benefits of choosing the library as opposed to the available alternatives.


james-d-elliott (approved):
LGTM

PR #2535, opened 2021-10-25
fix(web): misleading 2fa registration errors

### LINKED ISSUE #2449: Misleading used/expired 2FA registration token error
## Bug Report

User clicks on the "Register" button in the 2FA registration mail repeatedly. Looks like token gets used and subsequent attempts are erroneous. User is shown "Failed to generate the code to register your device"

Instead relevant error message should be shown a'la "The provided link has been used or is expired, please resend the mail"

nightah:
This change provides the means to notify a user when they have erroneously opened the device registration link in a device or browser which did not initiate the process. It also clears up an ambiguous message which may mislead users when their verification token is either expired or used.

Fixes #2449.


james-d-elliott (approved):
LGTM

PR #2597, opened 2021-11-11
feat: customizable static assets

### LINKED ISSUE #1630: Customizable logo & favicon
Would love to see an option to customize the logo and favicon!

image

nightah:
This change provides the means to override specific assets from the embedded Go FS with files situated on disk.

We only allow overriding the following files currently:
* favicon.ico
* logo.png

Here's an example replacing the logo:

Screenshot_2021-11-12-10-02-32

Closes #1630.


internal/server/server.go
```diff
@@ -48,10 +47,10 @@ func registerRoutes(configuration schema.Configuration, providers middlewares.Pr
 	r.GET("/api/"+apiFile, serveSwaggerAPIHandler)
 
 	for _, f := range rootFiles {
-		r.GET("/"+f, embeddedFS)
+		r.GET("/"+f, middlewares.AssetOverrideMiddleware(configuration.Server.AssetPath, embeddedFS))
```
james-d-elliott, line 50:
```suggestion
                if configuration.Server.AssetPath != "" {
		    r.GET("/"+f, middlewares.AssetOverrideMiddleware(configuration.Server.AssetPath, embeddedFS))
		} else {
		    r.GET("/"+f, embeddedFS)
		}
```

internal/server/server.go
```diff
@@ -48,10 +47,10 @@ func registerRoutes(configuration schema.Configuration, providers middlewares.Pr
...
 	}
 
-	r.GET("/static/{filepath:*}", embeddedFS)
+	r.GET("/static/{filepath:*}", middlewares.AssetOverrideMiddleware(configuration.Server.AssetPath, embeddedFS))
```
james-d-elliott, line 53:
```suggestion
	if configuration.Server.AssetPath != "" {
		    r.GET("/static/{filepath:*}", middlewares.AssetOverrideMiddleware(configuration.Server.AssetPath, embeddedFS))
		} else {
		    r.GET("/"+f, embeddedFS)
		}
```

james-d-elliott (approved):
LGTM. A couple of suggestions but they're probably not essential. Basically only utilizing the middleware if the assets path is configured.

internal/server/server.go
nightah: I have the conditional check inside the middleware otherwise it'll just ignore the logic.
That way we can unconditionally apply it to all embedded filesystems and avoid several conditionals in the other area.

PR #2616, opened 2021-11-22
fix(notifier): force use of sender email in smtp from cmd

### LINKED ISSUE #2571: Allow to set FROM email header in smtp notifier
## Feature Request

allow setting the `FROM` smtp header in smtp notifier

### Description

Currently only the `sender` configuration key can be set, which sets the `MAIL FROM` SMTP header as well as the `From` header in the email body.

The documentation states possible limitations on the `sender` key, namely 

> It should be noted that some SMTP servers require the username provided to have access to send from the specific address listed here.

These restrictions are only relevant for the `MAIL FROM` SMTP header.

On the other hand, setting the `From` header individually would allow the emails send by authelia to show up much nicer in email programs since it would be possible to use the `Authelia <admin@example.com>` format. `Authelia` can be set to any string desired and shows up as the email's sender in mail programs.

For difference between `MAIL FROM` and `From`:
https://en.wikipedia.org/wiki/Simple_Mail_Transfer_Protocol#SMTP_transport_example

### Use Case

Improve email formatting of smtp notifier. Especially important when you don't own a custom mail server where you can set up a custom email for every service.

james-d-elliott:
This change addresses an issue with the usage of the full sender configuration option in the `MAIL FROM` SMTP command. If a user includes a name in the sender this shouldn't be sent in the `MAIL FROM` command, instead we should extract it and use just the email portion.

Fixes #2571


internal/notification/const.go
```diff
@@ -1,4 +1,13 @@
...
+import (
+	"regexp"
+)
+
 const fileNotifierMode = 0600
 const rfc5322DateTimeLayout = "Mon, 2 Jan 2006 15:04:05 -0700"
+
+var (
+	// RegexpValidEmail matches if the email conforms to a known standard.
+	RegexpValidEmail = regexp.MustCompile(`^[0-9a-zA-Z+-_~]+@([a-zA-Z0-9]([a-zA-Z0-9-]+[a-zA-Z0-9])?\.){1,127}[a-z-A-Z]+$`)
```
nightah, line 12: This regex will fail to catch valid 2 character domains like `x@me.com`.

I think it would probably make sense to replace the validation logic where this regex is used with: https://pkg.go.dev/net/mail#ParseAddress.

nightah (changes requested):
Just a small issue/nit, otherwise this LGTM.

config.template.yml
```diff
@@ -587,8 +587,10 @@
     ## Can also be set using a secret: https://www.authelia.com/docs/configuration/secrets.html
     password: password
 
-    ## The address to send the email FROM.
-    sender: admin@example.com
+    ## The sender is used to configure who the email is from. This is used for the MAIL FROM command and the FROM header.
```
autheliabot, line 590: **[yamllint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>[error] line too long (121 > 120 characters) (line-length)

internal/configuration/config.template.yml
```diff
@@ -587,8 +587,10 @@
     ## Can also be set using a secret: https://www.authelia.com/docs/configuration/secrets.html
     password: password
 
-    ## The address to send the email FROM.
-    sender: admin@example.com
+    ## The sender is used to configure who the email is from. This is used for the MAIL FROM command and the FROM header.
```
autheliabot, line 590: **[yamllint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>[error] line too long (121 > 120 characters) (line-length)

internal/notification/smtp_notifier.go
```diff
@@ -20,40 +20,41 @@
 	configuration *schema.SMTPNotifierConfiguration
 	client        *smtp.Client
 	tlsConfig     *tls.Config
+	log           *logrus.Logger
```
autheliabot, line 23: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>undeclared name: `logrus` (typecheck)

nightah (approved):
LGTM.

PR #2622, opened 2021-11-23
feat(regulator): enhance authentication logs

### LINKED ISSUE #116: authentication_traces: second factor authentication (and IPs) not logged
Hi,

If I'm not mistaken, the second factor authentication is not logged.
Maybe it would be a good idea to replace “isAuthenticationSuccessful” by two distincts fields – one for each factor.
What's more, logging IP addresses would also be interesting, both for log analysis/forensics in a SIEM. and to trap bruteforce attempts.
My point is that bruteforce prevention could also be based on the originating IP address, rather than the username: as of now, if user `foobar` is subject to a bruteforce attack, it may be unable to *legitimately* log in – denial of service risk.

**P.-S.:** If anyone is interested in on how to prevent bruteforce attacks against Authelia's login page, one solution is to use Fail2Ban, with the following filter.
```fail2ban-filter
[Definition]

failregex = ^<HOST> -.*\] "POST /api/firstfactor.* HTTP/\d+\.\d+" [54]\d\d .*$
            ^<HOST> -.*\] "POST /api/totp.* HTTP/\d+\.\d+" 5\d\d .*$

ignoreregex =
```

This filter can be applied on an Nginx access log (default log format) dedicated to Authelia (to prevent false positives).

### LINKED ISSUE #1293: Extend the information stored in authentication_logs
Hello,
I would like to propose to extend the information that is stored in the authentication_logs sql table. Currently only the user, time and login success/failure is stored.

But I think it would be really awesome to also store for which website the user tried to authenticate and its IP. This way the database can be used to trace logins. Maybe another database table could be used to actually track when a user open page xy. 

Maybe a setting could be added to delete records older then xy days/weeks/... 

Thanks.

james-d-elliott:
This adds additional logging to the authentication logs such as type, remote IP, request method, redirect URL, and if the attempt was done during a ban. This also means we log attempts that occur when the attempt was blocked by the regulator for record keeping purposes, as well as record 2FA attempts which can be used to inform admins and later to regulate based on other factors.

Fixes #116, Fixes #1293.


internal/models/type_ipaddress.go
```diff
@@ -6,6 +6,11 @@
 	"net"
 )
 
+func NewIPAddressFromString(ip string) (ipAddress IPAddress) {
```
autheliabot, line 9: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>exported: exported function NewIPAddressFromString should have comment or be unexported (revive)

internal/storage/sql_provider.go
```diff
@@ -355,7 +355,9 @@ func (p *SQLProvider) LoadU2FDevice(ctx context.Context, username string) (devic
 
 // AppendAuthenticationLog append a mark to the authentication log.
 func (p *SQLProvider) AppendAuthenticationLog(ctx context.Context, attempt models.AuthenticationAttempt) (err error) {
-	if _, err = p.db.ExecContext(ctx, p.sqlInsertAuthenticationAttempt, attempt.Time, attempt.Successful, attempt.Username); err != nil {
+	if _, err = p.db.ExecContext(ctx, p.sqlInsertAuthenticationAttempt,
+		attempt.Time, attempt.Successful, attempt.Banned, attempt.Username,
+		attempt.Type, attempt.RemoteIP, attempt.RequestURI, attempt.RequestMethod); err != nil {
 		return fmt.Errorf("error inserting authentiation attempt: %w", err)
```
nightah, line 361:
```suggestion
		return fmt.Errorf("error inserting authentication attempt: %w", err)
```

nightah (approved):
LGTM only found a tiny typo.

PR #2655, opened 2021-12-01
refactor(handlers): utilize referer for auth logging rm/rd

james-d-elliott:
This utilizes the referrer query parameters instead of current request query parameters for logging the requested URI and method. Additionally adds some minor performance tweeks


nightah (approved):
LGTM.

PR #2658, opened 2021-12-02
fix(storage): duo/u2f upsert failure on postgresql

james-d-elliott:
This replaces the standard duo_devices upsert with a PostgreSQL specific one and ensures the u2f_devices upsert uses the new unique key for the ON CONFLICT check.


nightah (approved):
LGTM.

PR #2662, opened 2021-12-02
fix(storage): rebind all queries

james-d-elliott:
Fixes an issue in the new storage system. We forgot to rebind the queries to the PostgreSQL compatible format.


nightah (approved):
LGTM.

PR #2664, opened 2021-12-02
feat(storage): encrypt u2f key

james-d-elliott:
Adds encryption to the U2F public keys. While the public keys cannot be used to authenticate, only to validate someone is authenticated, if a rogue operator changed these in the database they may be able to bypass 2FA. This prevents that.


nightah (approved):
LGTM.

PR #2669, opened 2021-12-03
fix(models): potential panic generating jwt

james-d-elliott:
This ensures that at the time the JWT is generated for identity verification requests that a panic can't occur and instead an error will be returned.


nightah (approved):
LGTM.

PR #2676, opened 2021-12-04
fix(storage): don't check exp against time using sql

### LINKED ISSUE #2672: Cannot scan TOTP QR code in 4.33
### Description

I upgraded to 4.33 and set the storage:encryption_key, then set two factors auth on some of my subdomains. When accessing these domains, I get prompted to enter my username/password, then get send to the register device page. I click the Register Device link and receive an e-mail. Clicking on register in the e-mail bring me back to the (same) browser where a blurred QR code is displayed.

Log shows: "authelia token is not in db, it might have already been used".

When rolling back to 4.32, I can register the device.

### Expected Behaviour

I expect to be able to register for TOTP as I was able to prior to 4.33.

### Reproduction Steps

1. Start from scratch with no existing db.sqlite3 and no authelia docker image.
2. Start authelia/authelia:latest with two_factors set for some subdomains.
3. Access subdomain and follow through the TOTP registration process.
4. You will get a blurred QR code with error message "Authelia failed to register device ..."
5. Log will show "authelia token is not in db, it might have already been used".
6. Stop authelia, remove db, docker system prune -a --volumes.
7. Remove storage:encryption key and start authelia/authelia:4.32
8. Access subdomain and follow through the TOTP registration process.
9. Registration works.
10. Once registered in 4.32, you can upgrade to 4.33 and you will be able to access the two factors protected sites using the TOTP code.

### Additional Information

_N/A_

james-d-elliott:
This is already checked by JWT validation. There is no need and it's leading to timezone issues.

Fixes #2672

### DIFF
--- a/internal/models/types.go
+++ b/internal/models/types.go
@@ -51,6 +51,8 @@ func (ip *IP) Scan(src interface{}) (err error) {
 	switch v := src.(type) {
 	case string:
 		value = v
+	case []byte:
+		value = string(v)
 	default:
 		return fmt.Errorf("invalid type %T for IP %v", src, src)
 	}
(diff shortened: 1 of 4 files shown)


nightah (approved):
LGTM.

PR #2767, opened 2022-01-05
fix(middlewares): smart delay on reset password

james-d-elliott:
This adds a smart delay on reset password attempts to prevent username enumeration. Additionally utilizes crypto rand instead of math rand. It also moves the timing delay functionality into its own handler func.


james-d-elliott:
@clems4ever if you're still about this one was one I wanted to merge in before the release, if you're still about I'll looking at unit tests for the OIDC changes.

nightah (approved):
LGTM, gonna take this for a test run today to see if I have any additional feedback to provide.

PR #2776, opened 2022-01-07
fix(storage): use boolean instead of integer

### LINKED ISSUE #2775: PostgresSQL error
### Description

(100% cpu usage on 1 thread on Authelia once, that's why i started to look for problems)

Got an error from PostgreSQL log when i login on Authelia. Repeats every time i login. Not a SQL expert. But feels like something is not coded right in Authelia for PostgreSQL.
But Authelia is still working fine. Nothing in Authelias (on debugging) log.

### My setup

Docker Container on both PostgreSQL and Authelia. 
Same user-defined bridge.
Linux slackware/unraid.

### Things i tried to mitigate the problem

Updated Nginx/swag config file for authelia. Think that was the high CPU usage. But i'm not certain. Not seen it again (4 hour)

Updated from postgreSQL 13 -> 14. Nothing changed.

Purged the whole Authelia Database and started a brand new. Nothing changed.

### Additional Information

From postgres log:

2022-01-07 05:53:50.975 CET [264] ERROR: operator does not exist: boolean = integer at character 134

2022-01-07 05:53:50.975 CET [264] HINT: No operator matches the given name and argument types. You might need to add explicit type casts.
2022-01-07 05:53:50.975 CET [264] STATEMENT:
SELECT time, successful, username
FROM authentication_logs
WHERE time > $1 AND username = $2 AND auth_type = '1FA' AND banned = 0
ORDER BY time DESC
LIMIT $3
OFFSET $4;

james-d-elliott:
This fixes an issue where an integer is used instead of a boolean.

Fixes #2775.


clems4ever (approved):
LGTM

PR #2790, opened 2022-01-12
fix(configuration): mail address decode hook func

james-d-elliott:
This fixes an issue with the mail address decode hook func which previously would attempt to decode any struct type to a mail address.


clems4ever (changes requested):
Can we add a unit test?

james-d-elliott:
I actually think there is one, let me double check. It's part of configurati

> Can we add a unit test?

@clems4ever done!

PR #2799, opened 2022-01-15
refactor: handle scope descriptions client side

james-d-elliott:
This allows i18n for the OIDC descriptions.


web/src/i18n/locales/en.json
```diff
@@ -48,6 +48,13 @@
...
         "You're being signed out and redirected": "You're being signed out and redirected",
-        "Your supplied password does not meet the password policy requirements": "Your supplied password does not meet the password policy requirements."
+        "Your supplied password does not meet the password policy requirements": "Your supplied password does not meet the password policy requirements.",
+        "Use OpenID to verify your identity": "Use OpenID to verify your identity",
+        "Access your display name": "Access your display name",
+        "Access your group membership": "Access your group membership",
+        "Access your email addresses": "Access your email addresses",
+        "Accept": "Accept",
+        "Deny": "Deny",
+        "The above application is requesting the following permissions": "The above application is requesting the following permissions"
```
james-d-elliott, line 58: @mind-ar feel like translating those before I merge this?

james-d-elliott: Keep in mind `OpenID` is a terminology that shouldn't be translated.

mind-ar: Hi @james-d-elliott , just created https://github.com/authelia/authelia/pull/2825 pointing to this branch 
I don't know if this is the correct way to do a contribution over a branch other than master
let me know if i need to use another method

james-d-elliott: Yep that should be fine.

nightah (approved):
LGTM, though don't we need to update the OpenAPI spec for ConsentGetResponseBody?

james-d-elliott:
Actually, it's not documented since OIDC beta, will make sure to add it to a list to doc.

PR #2801, opened 2022-01-15
fix(oidc): add preferred username claim

### LINKED ISSUE #2798: OIDC claim for preferred username
Looks like "standard" one is `name`, at least that's what Harbor suggests using.

Please add the username claim that would allow forwarding username from AD

james-d-elliott:
This adds the missing preferred username claim to the ID Token for OIDC.

Fixes #2798


clems4ever (approved):
LGTM. Adding some unit tests could be a good idea, however.

PR #2813, opened 2022-01-20
fixed cloudflare NTP wrong port

thehedgefrog:
Fixed the wrong reference to `time.cloudflare.com:123` - the actual correct port is `1234`.

Unintended: trimmed some trailing white space.


thehedgefrog:
After testing it appears the error was caused by a misconfiguration on my part and not an application bug.  Closing.

james-d-elliott:
Thanks anyway, we appreciate the choice to contribute when people see issues and solutions. :)

PR #2829, opened 2022-01-22
fix(handlers): include preferred_username claim in meta

james-d-elliott:
This includes the preferred_username claim in the meta. Also uses the consts for all the applicable claims and scopes.


clems4ever (approved):
LGTM

PR #2835, opened 2022-01-25
fix(ntp): disable failure setting is ignored

james-d-elliott:
Disable failure was ignored previously, this ensures it isn't.


nightah (approved):
LGTM.

PR #2845, opened 2022-02-02
[WIP] feature SignUp

abs0lom:
## Motivation
For make authelia usable as a auth provider, I will implement Signup feature.

This functionality depends on the authentication backend. If the authentication backend is LDAP, the implementation will change the way the user object is stored.

It will therefore be necessary to retrieve from the configuration how the user is stored to ensure that it is created in the correct way.

## Security
When account creation is enabled, it becomes possible to know which email or username exists in the backend by spamming the creation form.

One solution is to ban the user for a certain amount of time after a certain number of unsuccessful attempts and to double the banning time for each additional failed attempt.

But this does not solve everything. Any suggestions to ensure the safety of the feature are appreciated.

## Tasks
- [X] create signup view into web
- [ ] create route for signup
- [ ] create user signup feature into authentication backend file
- [ ] create user signup feature into authentication backend LDAP (depending to implementation)
- [ ] add config key to enable/disable signup 
- [ ] make documentation against signup


DunklerPhoenix:
I have a feature request for this idea.
It would be nice to have an invite only mode with link, code or/and email.

james-d-elliott:
I think in general this idea is good, but you may want to discuss the overall design plan in an issue in advance otherwise it may end up being an implementation we ultimately don't take up. For starters I'm a bit confused as to the condition required for a user to sign up..

If the idea is that they customize their account themselves such as their display name and password (can't see any other attributes we'd want to allow users to modify) this probably fits better as a user control panel item, and they use the existing functionality to reset their password if they don't know it.

jaen:
@james-d-elliott I would like to see this feature merged, especially if it supports the invite-only mode. My knowledge of go is mostly limited to "it's just C with a garbage collector and CPS, what's the big deal", I've done a fix or two to traefik and wrote simple multi-homing connection quality helper for my Unifi setup. Do you have any suggestions on what I could do to move this forward?

davama:
This is very cool feature!

not to throw a wrench here but just a thought:

since there are different auth backends and their setup could very, why not just have authelia offload the creation (when invited/user not exist/ other security checks) to some webhook or script.

this way the onus is left to the admin to do whatever they want

my .02

thank you for this PR

Cajuteq:
Hi, pretty interested in allowing user signup as it drastically increase the number of project where we could use authelia as the auth. Is this feature still on tracks ? Is there a discussion related to the overall signup/invitation design ? I can allocate some dev time for it to be implemented

nightah:
Apologies for missing this @Cajuteq there hasn't been any discussion about this outside of this specific thread.

Per @james-d-elliott's earlier commentary:
> I think in general this idea is good, but you may want to discuss the overall design plan in an issue in advance otherwise it may end up being an implementation we ultimately don't take up.

We're open to the feature however the implementation in its current state is incomplete as it only includes the frontend portion. Really for this feature to be accepted it needs to handle either of our auth backends gracefully.

We're happy for someone to propose a design for evaluation and acceptance to ensure the contribution would be accepted.

Cajuteq:
no problem, I'll try to do something in the following week

PR #2848, opened 2022-02-05
fix(server): use of inconsistent methods for determining origin

james-d-elliott:
This unifies the methods to obtain the X-Forwarded-* header values and provides logical fallbacks. In addition, so we can ensure this functionality extends to the templated files we've converted the ServeTemplatedFile method into a function that operates as a middlewares.RequestHandler.

### DIFF
--- a/internal/server/template.go
+++ b/internal/server/template.go
@@ -7,16 +7,15 @@ import (
 	"path/filepath"
 	"text/template"
 
-	"github.com/valyala/fasthttp"
-
 	"github.com/authelia/authelia/v4/internal/logging"
+	"github.com/authelia/authelia/v4/internal/middlewares"
 	"github.com/authelia/authelia/v4/internal/utils"
 )
 
 // ServeTemplatedFile serves a templated version of a specified file,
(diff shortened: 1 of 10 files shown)


internal/server/template.go
```diff
@@ -51,16 +50,16 @@ func ServeTemplatedFile(publicDir, file, assetPath, duoSelfEnrollment, rememberM
...
-				scheme = "http"
-			default:
+				break
+			case "http", "https":
 				scheme = proto
 			}
 		}
 
-		baseURL := scheme + "://" + string(ctx.Request.Host()) + base + "/"
+		baseURL := scheme + "://" + string(ctx.XForwardedHost()) + ctx.Configuration.Server.Path + "/"
```
nightah, line 62:
```suggestion
		baseURL := scheme + "://" + string(ctx.XForwardedHost()) + base + "/"
```

This needs to change otherwise it'll always include the path.

nightah (changes requested):
Small issue, otherwise LGTM.

PR #2856, opened 2022-02-08
feat(server): csp template

### LINKED ISSUE #2853: Allowing modification of Content-Security-Policy header through configuration.yaml
## Feature Request

Authelia includes a Content Security Policy by default which is fairly strict (great!), using nonce values. However I was thinking if this could be something that could be appended to through the configuration.yaml?

I could easily override the entire header through nginx, but that would then mean manually maintaining nonce values or downgrading the stricter policy set currently which could change on each update and become hard to maintain.

Using Grafana as an example, they provide a config key like this:

```
content_security_policy_template = """script-src 'self' 'unsafe-eval' 'unsafe-inline' 'strict-dynamic' $NONCE;object-src 'none';font-src 'self';style-src 'self' 'unsafe-inline' blob:;img-src * data:;base-uri 'self';connect-src 'self' grafana.com ws://$ROOT_PATH wss://$ROOT_PATH;manifest-src 'self';media-src 'none';form-action 'self';"""
```

$NONCE and $ROOT_PATH are dynamic values which are parsed before outputting the header at render time. You can add anything else you want to the policy but maintain the nonce values required.

Perhaps a similar principle could be done for Authelia, given `nonce-%s` is a variable in the main Go server template?

### Use Case

Being able to add or append additional properties such as report-uri for tracking and monitoring Content Security Policy violations, without overriding the header entirely.

james-d-elliott:
Implements the ability for advanced users to override the CSP.

Closes #2853.


clems4ever (approved):
LGTM

Issue #2636, opened 2021-11-26
How to integrate Authelia with django


jrmzlara:
Has anyone implemented (combined) Authelia as SSO and Django (both django and DRF)? Or maybe point out some articles/references?

bielfrontera:
I have some Django applications using Authelia for authentication. I just followed [authelia documentation regarding protecting a location using nginx](https://www.authelia.com/docs/deployment/supported-proxies/nginx.html) and added a custom PersistentRemoteUserMiddleware class:

```python
from django.contrib.auth.middleware import PersistentRemoteUserMiddleware


class PersistentHttpRemoteUserMiddleware(PersistentRemoteUserMiddleware):
    header = 'HTTP_REMOTE_USER' 
``` 
(see Django documentation about (REMOTE_USER)[https://docs.djangoproject.com/en/3.2/howto/auth-remote-user/#using-remote-user-on-login-pages-only]).

I had also to configure the variable LOGOUT_REDIRECT_URL in settings in order to complete the logout with authelia (value https://auth.<domain>.<tdl>/logout )

clems4ever:
@bielfrontera , thanks for sharing your experience. It would be awesome if you could contribute to the community section of the documentation: https://www.authelia.com/docs/community/.

What do you think?

clems4ever:
The documentation from @bielfrontera will soon be available in the doc. Thank you very much for your contribution.

PR #2858, opened 2022-02-08
docs: add community example of integrating authelia with Django

bielfrontera:
Extend the information given at #2636 about how to integrate authelia with Django.


clems4ever:
Thank you for this contribution @bielfrontera .

PR #2879, opened 2022-02-18
Create LDAP_Recommendation.md

shreyasajj:
Add a Recommendation for lldap

### DIFF
--- a/docs/community/LDAP_Recommendation.md
+++ b/docs/community/LDAP_Recommendation.md
@@ -1,0 +1,209 @@
+---
+layout: default
+title: LDAP Recommendation
+parent: Community
+nav_order: 7
+has_children: true
+has_toc: false
+---
+# LDAP Recommendation
+## Not Maintained By Authelia
+[https://github.com/nitnelave/lldap](https://github.com/nitnelave/lldap)
+Nitnelave create a simple and elegant way to create light weight ldap
+
+# Prerequisite
+* Linux Machine
+* Docker
+* Docker-compose
+* Authelia installed on the same machine
+
+# Setup
+1. Create docker file in the same project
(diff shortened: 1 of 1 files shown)


clems4ever:
Hello, thanks for you contribution @shreyasajj .

You covered something that is kind of missing around [here](https://www.authelia.com/docs/deployment/) in our official documentation so far, i.e., deployment of Authelia along with an LDAP server, which is cool and I'm grateful for that. However, I would not advertise your documentation as a 'recommendation' because:

1. lldap, unlike an AD or an OpenLDAP server and as far as I'm concerned, is not made for production because it's only a non-complete implementation of LDAP and it implements opinionated defaults as stated in the README.
2. 'recommendation' could lead people to think that the core team advises using this project which we don't.
3. Describing the deployment for each LDAP server out there won't be manageable in the long run when it comes to keeping the documentation up-to-date.

For those reasons I will, unfortunately, decline this PR but I opened issue #2886 to document/implement a deployment with a standard OpenLDAP server that we will officially support. Feel free to contribute to that instead if you want.

I'm sorry and I hope you understand my point.

shreyasajj:
I totally understand, I did ask in the authelia chat before making a pr request, but I totally understand where you are coming from. I thought It would help people to have an easy setup and give some support to the owner of the github repo for ldap.

james-d-elliott:
Yep, that's my bad, sorry to waste your time. I thought it was a guide on configuring OpenLDAP, and you were wanting to point to it in your docs. I should have spent more time looking at the link, deepest apologies!

SAOPP:
Btw guys, I got working combination ldap with authelia in my setup, if it can help some one, I can share it ova here.

PR #2887, opened 2022-02-20
build: disable pre-commit hook by env variable

clems4ever:
One can set the NO_HOOK env variable to avoid running the pre-commit
hook. It can be useful when performing large refactorings.


james-d-elliott (approved):
LGTM

Issue #2921, opened 2022-02-28
Expose options to enable PKCE for OAuth 2 Authorization Code flow


FineWolf:
## Feature Request
Add options in the configuration schema to enable PKCE for authorization code flow.

### Description
When authenticating a client application (SPA-type frontend application), the recommended approach to properly authenticate is to use the Authorization Flow with PKCE. [Proof Key for Code Exchange is an extension of the authentication code flow](https://oauth.net/2/pkce/) which aims to prevent CSRF and authorization code injection attacks by requiring the client application to prove that it is the owner of a token request.

The work to enable this in Authelia would be minimal since Fosite (the underlying OIDC/OAuth2 library) already has support for PKCE.

The required changes would be as follows:

- Add 3 parameters in `OpenIDConnectConfiguration` (`internal\configuration\schema\identity_providers.go`) to expose PKCE options, specifically:
  - `EnforcePKCE` / `enforce_pkce` (bool - if set to true, requires clients to perform authorization code flows with PKCE - default false)
  - `EnforcePKCEForPublicClients` / `enforce_pkce_for_public_clients` (bool - if set to true, requires only public clients to perform authorization code flows with PKCE - default false - see note below)
  - `EnablePKCEPlainChallengeMethod`  / `enable_pkce_plain_challenge_method` (bool - if set to true, allows `challenge_method = plain` for PKCE challenges - default false [should use `S256` instead])
- Populate relevant default values (see note below)
- In `internal\oidc\provider.go`, enable `OAuth2PKCEFactory` and populate the relevant config from the configuration object.
- Edit the documentation to reflect the changes and recommended values.

---

If you would prefer that default values are set to the most secure and sensible option, I would recommend setting `EnforcePKCEForPublicClients` to true. However, this would be a breaking changes for public clients using the authorization code flow.

---

I can work on the PR required for this feature.

james-d-elliott:
SGTM. You're welcome to do it, otherwise I'm probably able to take a look in a few days.

james-d-elliott:
Also while I see the logic behind this being a breaking change, as OIDC is very clearly marked as beta and subject to breaking changes, I'm inclined to say we just go ahead with the default of enabled. I'd be interested in what @clems4ever and @nightah have to say.

Configuration would look similar to this if we disabled by default:

```yaml
identity_providers:
  oidc:
    pkce:
      enforce: false
      enforce_public_clients: false
      enable_plain_challenge: false
```

[...]

PR #2924, opened 2022-02-28
feat(oidc): add pkce support

FineWolf:
Implementation for #2921 .

Two new options are added in the configuration schema:

- `enforce_pkce`
   String. Can either be `no`, `public_only`, or `yes`. Default is `public_only` (not a lot of clients supports PKCE when the code exchange is done in the backend).
- `enable_pkce_plain_challenge`
   Boolean. Defaults to `false`.

Documentation has been updated to reflect the new options.
`provider.go` has been modified to enable PKCE based on the new options.
Tests have been adjusted for the new options.


james-d-elliott:
I believe we'll need to also add the metadata option code_challenge_methods_supported, which should include S256, and plain if the plain option is enabled. Unless I'm reading the spec wrong not including this would signal to clients that PKCE is not supported.


Excerpt from https://datatracker.ietf.org/doc/html/rfc8414#section-2:
```
code_challenge_methods_supported
      OPTIONAL.  JSON array containing a list of Proof Key for Code
      Exchange (PKCE) [[RFC7636](https://datatracker.ietf.org/doc/html/rfc7636)] code challenge methods supported by this
      authorization server.  Code challenge method values are used in
      the "code_challenge_method" parameter defined in [Section 4.3 of
      [RFC7636]](https://datatracker.ietf.org/doc/html/rfc7636#section-4.3).  The valid code challenge method values are those
      registered in the IANA "PKCE Code Challenge Methods" registry
      [[IANA.OAuth.Parameters](https://datatracker.ietf.org/doc/html/rfc8414#ref-IANA.OAuth.Parameters)].  If omitted, the authorization server
      does not support PKCE.
```

FineWolf:
@james-d-elliott You are absolutely right, small oversight on my part.

PR has been adjusted to include the appropriate metadata.

clems4ever (commented):
Thanks for this PR @FineWolf , it's all good for me in terms of code. And thanks for the test. Great job!

I only have one last question. I'm wondering if the values for this `enforce_pkce` option should not be called `all`, `public_only` and `none` instead of `yes`, `no` and `public_only`. Because when I read `yes` at the beginning I thought the type of the attribute was supposed to be a bool and so `true` instead of `yes` would make more sense but it's not (it's a string, an enum semantically) and therefore having a mix between `yes`, `no` and a third value did not make sense to me. I think all values of this option should reflect what this policy affects instead of sometimes stating what's it affects (public_only) and sometimes whether it affects anything (yes or no). Do you see what I mean?

FineWolf:
> I only have one last question. I'm wondering if the values for this `enforce_pkce` option should not be called `all`, `public_only` and `none` instead of `yes`, `no` and `public_only`.

Indeed. Actually, `never`, `public_clients_only`, `always` is a bit more descriptive. I've updated the PR to use those values.

docs/configuration/identity-providers/oidc.md
```diff
@@ -159,6 +159,7 @@ identity_providers:
     id_token_lifespan: 1h
     refresh_token_lifespan: 90m
     enable_client_debug_messages: false
+    enforce_pkce: public_clients_only
```
james-d-elliott, line 162:
```suggestion
    enforce_pkce: always
```

config.template.yml
```diff
@@ -684,6 +684,10 @@ notifier:
     ## security reasons.
     # minimum_parameter_entropy: 8
 
+    ## SECURITY NOTICE: It's not recommended changing this option, and highly discouraged to have it set to 'never'
+    ## for security reasons.
+    # enforce_pkce: public_clients_only
```
james-d-elliott, line 689:
```suggestion
    # enforce_pkce: always
```

docs/configuration/identity-providers/oidc.md
```diff
@@ -309,6 +310,39 @@ This controls the minimum length of the `nonce` and `state` parameters.
 make certain scenarios less secure. It is highly encouraged that if your OpenID Connect RP does not send these parameters
 or sends parameters with a lower length than the default that they implement a change rather than changing this value.
 
+### enforce_pkce
+<div markdown="1">
+type: string
+{: .label .label-config .label-purple } 
+default: public_only
```
james-d-elliott, line 317:
```suggestion
default: always
```

docs/configuration/identity-providers/oidc.md
```diff
@@ -309,6 +310,39 @@ This controls the minimum length of the `nonce` and `state` parameters.
...
+{: .label .label-config .label-blue }
+required: no
+{: .label .label-config .label-green }
+</div>
+
+[Proof Key for Code Exchange](https://datatracker.ietf.org/doc/html/rfc7636) enforcement policy: if specified, must be either `never`, `public_clients_only` or `always`.
+
+If set to `public_clients_only` (default), PKCE will be required for public clients using the Authorization Code flow.
+
+When set to `always`, PKCE will be required for all clients using the Authorization Code flow.
```
james-d-elliott, line 327:
```suggestion
When set to the default of `always`, PKCE will be required for all clients using the Authorization Code flow.

If set to `public_clients_only`, PKCE will be required for public clients using the Authorization Code flow.
```

internal/configuration/config.template.yml
```diff
@@ -684,6 +684,10 @@ notifier:
     ## security reasons.
     # minimum_parameter_entropy: 8
 
+    ## SECURITY NOTICE: It's not recommended changing this option, and highly discouraged to have it set to 'never'
+    ## for security reasons.
+    # enforce_pkce: public_clients_only
```
james-d-elliott, line 689:
```suggestion
    # enforce_pkce: always
```

internal/configuration/schema/identity_providers.go
```diff
@@ -49,6 +52,7 @@ var DefaultOpenIDConnectConfiguration = OpenIDConnectConfiguration{
 	AuthorizeCodeLifespan: time.Minute,
 	IDTokenLifespan:       time.Hour,
 	RefreshTokenLifespan:  time.Minute * 90,
+	EnforcePKCE:           "public_clients_only",
```
james-d-elliott, line 55:
```suggestion
	EnforcePKCE:           "always",
```

internal/configuration/validator/const.go
```diff
@@ -121,6 +121,9 @@ const (
 		"more clients configured"
 	errFmtOIDCNoPrivateKey = "identity_providers: oidc: option 'issuer_private_key' is required"
 
+	errFmtOIDCEnforcePKCEInvalidValue = "identity_providers: oidc: option 'enforce_pkce' must be 'never', " +
+		"'public_clients_only' or 'always', but it is configured as '%s'"
```
james-d-elliott, line 125: Nice job sticking to the implied style!

james-d-elliott (changes requested):
LGTM! Few changes based on discussions with the team. We feel since it's configurable and OIDC is specifically a beta feature, we can make a potentially breaking change and have them change it as a preference to security over convenience.

In addition I was wondering what you thought about changing enable_pkce_plain_challenge to something like `pkce_code_challenge_methods` as a slice with the relevant options? I"m not hard set on it and it would require some additional validation logic, but felt like maybe it's a decent alternative in the event additional challenge modes become available over time.

FineWolf:
I can definitely change `pkce_code_challenge_methods` to a slice with `S256` and `plain` as valid options. I won't be able to make those changes before Thursday however.

---

As for changing `enforce_pkce`'s default value to `always`, I would personally recommend against it.

Support for PKCE isn't widespread yet, and it was really meant to solve the issue where frontend/SPA/mobile applications couldn't use the `client_secret` as a reliable way of authenticating the application making the token request. Without PKCE, frontend/SPA/mobile applications were forced to use the implicit flow or the authorization code flow which were vulnerable to interception attacks.

In a situation where a client app is not a public app and has a `client_secret` to rely on to authenticate the source of the token request, PKCE at most an added countermeasure.

Nothing however is stopping private applications from requesting `PKCE` when `enforce_pkce` is set to `public_clients_only`; if a `code_challenge` is sent along with the authorization request, it will be taken into account and will enforce the token request to have the appropriate `code_verifier` along with it. However, it will allow authorization requests without them. The application however, because it is a private app, still needs to provide its `client_secret` to authenticate the source of the token request.

`public_clients_only` is a secure and sane default in my opinion. It still allows private clients to enroll in PKCE if they so desire, without outright rejecting the authorization request if no `code_challenge` is provided. Public apps using the auth code flow however are susceptible to authorization code interception attacks however, and have the most to gain by forcing applications to provide a PKCE challenge.

internal/configuration/config.template.yml
FineWolf: See https://github.com/authelia/authelia/pull/2924#issuecomment-1056113203

docs/configuration/identity-providers/oidc.md
FineWolf: See https://github.com/authelia/authelia/pull/2924#issuecomment-1056113203

docs/configuration/identity-providers/oidc.md
FineWolf: See https://github.com/authelia/authelia/pull/2924#issuecomment-1056113203

FineWolf: Also, comment for myself, that should have been `default: public_clients_only`.

config.template.yml
FineWolf: See https://github.com/authelia/authelia/pull/2924#issuecomment-1056113203

docs/configuration/identity-providers/oidc.md
FineWolf: See https://github.com/authelia/authelia/pull/2924#issuecomment-1056113203

docs/configuration/identity-providers/oidc.md
james-d-elliott: Oh Yep! I'll make this change quickly if it hasn't been done. Good spot!

nightah:
@allcontributors please add @FineWolf for code, docs and tests.

PR #2930, opened 2022-03-01
fix(oidc): make preferred_username a profile scope claim

james-d-elliott:
This corrects an issue with the preferred_username which should be part of the profile scope as per https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims.

Introduced in ddbb21a via #2829


nightah (approved):
LGTM.

PR #2985, opened 2022-03-09
fix(configuration): cannot disable remember me

### LINKED ISSUE #2978: Remember me checkbox if duration equal 0
Hi there,

It seemed to me that if option `remember_me_duration` was set to 0, the "Remember me" checkbox was not visible on the login screen. Did I dream of this ?

Currently, it is visible
image

image

As always, thank you for your hard work.

james-d-elliott:
This allows users to disable remember me again.

Fixes #2978

### DIFF
--- a/config.template.yml
+++ b/config.template.yml
@@ -461,7 +461,7 @@ session:
   inactivity: 5m
 
   ## The time before the cookie expires and the session is destroyed if remember me IS selected.
-  ## Value of 0 disables remember me.
+  ## Value of -1 disables remember me.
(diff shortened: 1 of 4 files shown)


ToqQrrl:
It would seem that the WebUI is still showing the "Remember Me" checkbox and I can not disable "Remember Me" although the remember_me_duration value is set to -1 in the "session" section of the configuration.yml file.

I'm on version 4.34.4.

I have tried it on more that one computer in case it was a cache issue but it is not.

PR #2999, opened 2022-03-13
fix(web): show appropriate default and available methods

### LINKED ISSUE #2998: One-Time password device registration is showed even when it's disabled
### Description

I've deployed Authelia `4.34.4` via Helm `authelia-0.8.15`. I've disable TOTP and WebAuthn is enabled. The backend user database is OpenLDAP.

I went to setup a new user and was welcomed with the usual login/password form. The login details were accepted as expected and I was able to proceed to the next screen.

On the next screen, and despite TOTP being disable, the default device registration option showed is One-Time Password. Upon clicking on the `Methods` link, only `Security-Key | Webauthn` is available as expected.

image

### Expected Behaviour

If TOTP is disabled then device registration method should default to Webauthn. Maybe the `Methods` link should also be hidden since there's only 1 method available.

### Reproduction Steps

1. Deploy Authelia with TOTP disable and WebAuthn enabled and a valid user database
2. Access the Authelia interface and authenticate with an unregistered user
3. The device registration method showed is `One-Time Password`

james-d-elliott:
This ensures that; the preferred method endpoint only accepts enabled methods, the preferred method set when a user does not have a preferred method is a method that is available, that if a user has a preferred method that is not available it is changed to an enabled method with preference put on methods the user has configured, that the frontend does not show the method selection option when only one method is available.

Fixes #2998


internal/handlers/handler_user_info.go
```diff
@@ -19,6 +18,13 @@ func UserInfoGet(ctx *middlewares.AutheliaCtx) {
 		return
 	}
 
+	if changed := userInfo.SetDefaultMethod(ctx.AvailableSecondFactorMethods()); changed {
```
clems4ever, line 21: the method name could be clearer. Usually the "set" prefix is for setters, i.e., dumb methods setting an attribute of the object but in this case, the method seems to be more complex according to the signature returning a bool so I suggest renaming it to ChangeDefaultMethod instead.

clems4ever: However, taking a step back, I think updating objects in a GET HTTP endpoint is not a good idea in general in terms of the user contract. It violates the semantic of GET.
IMO, to workaround this issue, this corner case should be treated in the client (browser) instead. The client should retrieve the available factors and compare to the retrieved user method. If, still in the client, the user method is not available then the client should send a POST to set the new method derived from one available.

internal/handlers/handler_user_info_test.go
```diff
@@ -41,7 +42,17 @@ type expectedResponse struct {
...
+	description string
+
+	db      model.UserInfo
+	api     *model.UserInfo
+	loadErr error
+	saveErr error
+	config  *schema.Configuration
+}
+
+func TestUserInfoEndpoint_SetCorrectMethod(t *testing.T) {
```
clems4ever, line 55: A get endpoint should not update any state. This violates the GET semantic, it's not a good idea.

internal/storage/sql_provider.go
```diff
@@ -219,7 +218,7 @@ func (p *SQLProvider) LoadUserInfo(ctx context.Context, username string) (info m
 	case err == nil:
 		return info, nil
 	case errors.Is(err, sql.ErrNoRows):
-		if _, err = p.db.ExecContext(ctx, p.sqlUpsertPreferred2FAMethod, username, authentication.PossibleMethods[0]); err != nil {
+		if _, err = p.db.ExecContext(ctx, p.sqlUpsertPreferred2FAMethod, username, ""); err != nil {
```
clems4ever, line 221: Upserting in a method called LoadUserInfo seems broken to me. This relates to previous comments.

clems4ever (changes requested):

internal/storage/sql_provider.go
james-d-elliott: Yeah  I understand,the reason this was done was to prevent the methods used to check which 2FA technologies the user had registered each had their own method. Adding a single LoadUserInfo method allows us to reduce the number of methods by 3, and execute 1 query instead of 4 in the instance of the preferred method existing and 3 instead of 4 in the instance of the preferred method not existing (first time).

internal/model/user_info_test.go
```diff
@@ -0,0 +1,130 @@
...
+	assert.Equal(t, SecondFactorMethodDuo, info.Method)
+
+	info = UserInfo{
+		Method:      "",
+		HasDuo:      false,
+		HasTOTP:     false,
+		HasWebauthn: false,
+	}
+
+	changed = info.SetDefaultPreferred2FAMethod([]string{SecondFactorMethodTOTP, SecondFactorMethodWebauthn, SecondFactorMethodDuo})
```
autheliabot, line 94: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>ineffectual assignment to changed (ineffassign)

web/src/services/UserInfo.ts
```diff
@@ -1,7 +1,7 @@
 import { SecondFactorMethod } from "@models/Methods";
 import { UserInfo } from "@models/UserInfo";
 import { UserInfo2FAMethodPath, UserInfoPath } from "@services/Api";
-import { Get, PostWithOptionalResponse } from "@services/Client";
+import { Get, Post, PostWithOptionalResponse } from "@services/Client";
```
autheliabot, line 4: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'Get' is defined but never used.

clems4ever:
Tests are still broken

james-d-elliott:
It's a false positive most likely, retrying it.

internal/handlers/handler_user_info_test.go
```diff
@@ -123,6 +137,207 @@ func TestMethodSetToU2F(t *testing.T) {
...
+				TOTP: schema.TOTPConfiguration{
+					Disable: true,
+				},
+				DuoAPI: &schema.DuoAPIConfiguration{},
+			},
+			loadErr: nil,
+			saveErr: nil,
+		},
+		{
+			description: "",
```
clems4ever, line 242: is it expected that the description is empty here?

james-d-elliott, line 242:
```suggestion
			description: "should default new users to totp if all enabled",
```

james-d-elliott: Nope, thanks, nice one!

internal/model/user_info_test.go
```diff
@@ -0,0 +1,131 @@
+package model
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUserInfo_SetDefaultMethod_ShouldConfigureConfigDefault(t *testing.T) {
```
clems4ever, line 9: OPT: use sub-tests

web/src/services/UserInfo.ts
```diff
@@ -39,8 +39,8 @@ export function toString(method: SecondFactorMethod): Method2FA {
     }
 }
 
-export async function getUserInfo(): Promise<UserInfo> {
-    const res = await Get<UserInfoPayload>(UserInfoPath);
+export async function postUserInfo(): Promise<UserInfo> {
+    const res = await Post<UserInfoPayload>(UserInfoPath);
```
clems4ever, line 43: If we just use the POST endpoint, why not totally remove the GET one from the backend?

clems4ever (approved):
LGTM

internal/model/user_info_test.go
```diff
@@ -0,0 +1,220 @@
...
+	testName := func(i int, have UserInfo, availableMethods []string) string {
+		method := have.Method
+
+		if method == "" {
+			method = "none"
```
autheliabot, line 16: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>string `none` has 2 occurrences, make it a constant (goconst)

PR #3008, opened 2022-03-14
docs: adjust acl policy example to be possible

### LINKED ISSUE #3006: Conflicting explanation in docs/configuration/access-control.html
The page https://www.authelia.com/docs/configuration/access-control.html lists on the very top the following example (excerpt):

```
  rules:
  - domain: public.example.com
    policy: bypass
    networks:
    - internal
    - 1.1.1.1
    subject:
    - ["user:adam"]
```

Which doesn't make much sense: How does Authelia identify user `adam` if authentication is bypassed?

Further down the documentation states:

> bypass
> This policy skips all authentication and allows anyone to use the resource. This policy is not available with a rule that includes a [subject](https://www.authelia.com/docs/configuration/access-control.html#subject) restriction because the minimum authentication level required to obtain information about the subject is [one_factor](https://www.authelia.com/docs/configuration/access-control.html#one_factor).

==> **The example should be in line with the explanation.**

Alestrix:
Fix #3006

### DIFF
--- a/docs/configuration/access-control.md
+++ b/docs/configuration/access-control.md
@@ -23,7 +23,7 @@ access_control:
 
   rules:
   - domain: public.example.com
-    policy: bypass
+    policy: one_factor
     networks:
     - internal
     - 1.1.1.1


james-d-elliott (approved):
LGTM

james-d-elliott:
@all-contributors please add @Alestrix for docs

PR #3065, opened 2022-03-27
feat: implement mutual tls in the web server

### LINKED ISSUE #3041: Support for mTLS
Currently, the Authelia server config supports specifying a TLS certificate and key. Options to allow or require a client certificate, and to specify a CA certificate to verify the client certificate should be added. This enhances security by better identifying connecting clients or proxy servers.

clems4ever:
mutual tls prevent other clients than the proxies to communicate with Authelia. This reduces the attack
surface.

Fix #3041

### DIFF
--- a/docs/security/measures.md
+++ b/docs/security/measures.md
@@ -189,6 +189,17 @@ Authelia protects your users against open redirect attacks by always checking if
 to a subdomain of the domain protected by Authelia. This prevents phishing campaigns tricking users into visiting
 infected websites leveraging legit links.
 
+## Mutual TLS
+
+For the best security protection, configuration with TLS is highly recommended. TLS is used to secure the connection between
+the proxies and Authelia meaning that an attacker on the network cannot perform a man-in-the-middle attack on those
+connections. However, an attacker on the network can still impersonate proxies. In order to mitigate this risk, mutual
+TLS can be used.
+Mutual TLS brings mutual authentication between Authelia and the proxies. Any other party attempting to contact Authelia
+will not even be able to create a TCP connection.
+
+To configure mutual TLS, please refer to [this document](../configuration/server.md#client_certificates)
+
 ## Additional security
 
 ### Reset Password
--- a/docs/security/measures.md
+++ b/docs/security/measures.md
@@ -192,11 +192,13 @@ infected websites leveraging legit links.
 ## Mutual TLS
 
 For the best security protection, configuration with TLS is highly recommended. TLS is used to secure the connection between
-the proxies and Authelia meaning that an attacker on the network cannot perform a man-in-the-middle attack on those
-connections. However, an attacker on the network can still impersonate proxies. In order to mitigate this risk, mutual
-TLS can be used.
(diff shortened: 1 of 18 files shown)


internal/server/server.go
```diff
@@ -1,14 +1,19 @@
 package server
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"embed"
 	"io/fs"
+	"io/ioutil"
```
james-d-elliott, line 8: The stdlib package `io/ioutil` is discouraged for new code from 1.16 onwards., there should be equivalent methods in the `io` or `os` packages.

internal/configuration/validator/server_test.go
```diff
@@ -1,6 +1,8 @@
 package validator
 
 import (
+	"io/ioutil"
```
james-d-elliott, line 4: The stdlib package io/ioutil is discouraged for new code from 1.16 onwards., there should be equivalent methods in the io or os packages.

internal/commands/certificates.go
```diff
@@ -1,22 +1,16 @@
...
-	"crypto/ecdsa"
-	"crypto/ed25519"
 	"crypto/elliptic"
-	"crypto/rand"
-	"crypto/rsa"
-	"crypto/x509"
-	"crypto/x509/pkix"
-	"encoding/pem"
 	"fmt"
+	"io/ioutil"
```
james-d-elliott, line 6: The stdlib package io/ioutil is discouraged for new code from 1.16 onwards., there should be equivalent methods in the io or os packages.

docs/security/measures.md
```diff
@@ -189,6 +189,17 @@ Authelia protects your users against open redirect attacks by always checking if
 to a subdomain of the domain protected by Authelia. This prevents phishing campaigns tricking users into visiting
 infected websites leveraging legit links.
 
+## Mutual TLS
+
+For the best security protection, configuration with TLS is highly recommended. TLS is used to secure the connection between
```
james-d-elliott, line 194: Maybe we should add a note about under what situations this is recommended? I'm not entirely sure it adds any explicit protection in a single host situation for example. Thoughts?

james-d-elliott (changes requested):
Few changes otherwise LGTM. We should probably still also discuss factorizing the crypto generation functionality.

docs/security/measures.md
clems4ever: mutual TLS is recommended in most cases imo. It's not a matter of number of instances but rather a matter of which component in the private network can reach Authelia (other services on the same machine, other servers, routers, firewalls). Even the proxy trust could be spoofed. The only case where I think mTLS does not add anything is in the case of service mesh or network overlays configured with an ACL (defining which services can talk to each other) specifically for the proxy-to-Authelia communication. In all other cases, any component from within the internal network could try to contact Authelia and attempt an attack.

Even though we can expect an attack to be hard or even unfeasible against Authelia if we have done our job well (proper working authentication and authorization mainly), no one could ever guarantee there won't be any breach... mTLS allows to rather be safe than sorry.

james-d-elliott (approved):
LGTM

PR #3081, opened 2022-03-30
feat(configuration): configurable default second factor method

### LINKED ISSUE #2838: Allow setting of "default" 2fa method for new users
### Description

I would like to be able to set the "default" 2fa method for new users for the registration flow. Being able to set it to duo push for example

We have left the totp settings option blank in configuration.yml (we dont want any auth method other than duo)
This works as expected.
However.
When an ldap user logs in to authelia for the first time, they get a prompt to register their device. As soon as they hit register it will return an error `There was a problem initiating the registration process`. 
this seems to be because as soon as you login you are redirected to the totp registration process, the user doesn't get a choice initially (they can hit methods and then choose push, which works, but isn't a great user experience)

### Use Case

In my usecase we have an existing ldap & duo setup. All users exist in both. 
We wish users to only use duo as an option. we want the onboard flow to go from login to authelia  > follow duo push setup flow.
right now they have to login to authelia > press methods (assuming they've read the documentation email they've been given) > press push > follow duo push setup flow

### Additional info

I brought this up in discord. @james-d-elliott pointed out that https://github.com/authelia/authelia/blob/v4.33.2/internal/storage/sql_provider.go#L212 would probably need to be adjusted to the below as a temporary workaround (assuming i can get a local build working well)
```
         if _, err = p.db.ExecContext(ctx, p.sqlUpsertPreferred2FAMethod, username, authentication.Push); err != nil {
```

james-d-elliott:
This allows configuring the default second factor method.

Closes #2838


internal/handlers/handler_register_duo_device.go
```diff
@@ -45,7 +45,7 @@ func SecondFactorDuoDevicesGet(duoAPI duo.API) middlewares.RequestHandler {
 		}
 
 		if result == allow {
-			ctx.Logger.Debugf("Device selection not possible for user %s, because Duo authentication was bypassed - Defaults to Auto Push", userSession.Username)
+			ctx.Logger.Debugf("Device selection not possible for user %s, because Duo authentication was bypassed - Default to Auto Push", userSession.Username)
```
james-d-elliott, line 48:
```suggestion
			ctx.Logger.Debugf("Device selection not possible for user %s, because Duo authentication was bypassed - Defaults to Auto Push", userSession.Username)
```

PR #3005, opened 2022-03-14
feat(oidc): provide cors config including options handlers

james-d-elliott:
This adjusts the CORS headers appropriately for OpenID Connect. This includes responding to OPTIONS requests appropriately for CORS preflight requests and discovery. It allows users to configure which endpoints utilize the CORS handler with a default of disabled, and allows them to either configure Origins implicitly by request, implicitly by including the expected Origins from clients, or from explicitly adding Origins to a list.

ToDo:
- [x] Make users explicitly enable it on non-metadata endpoints.

PR #3087, opened 2022-04-01
fix(server): respond with 404/405 appropriately

james-d-elliott:
This adjusts the not found handler to not respond with a 404 on not found endpoints that are part of the /api or /.well-known folders, and respond with a 405 when the method isn't implemented.

Extracted from #3005


nightah (approved):
LGTM.

PR #3106, opened 2022-04-03
fix(web): update client rendering method

nightah:
This PR utilises the React 18 Client Rendering API along with createRoot as opposed to the React 17 ReactDOM.render method.


james-d-elliott (approved):
LGTM

PR #2965, opened 2022-03-05
feat(oidc): persistent storage

james-d-elliott:
This moves the OpenID Connect storage from memory into the SQL storage, making it persistent and allowing it to be used with clustered deployments like the rest of Authelia.

Closes #2677

PR #3116, opened 2022-04-05
feat(oidc): pairwise subject identifiers

james-d-elliott:
Allows configuring clients with a sector identifier to allow pairwise subject types.


james-d-elliott:
Merge after #2965.

PR #3217, opened 2022-04-19
fix(suites): add missing traefik routes

nightah:
This change includes missing routes for both the Traefik and Traefik2 suites, issues would have manifested running dev mode tests for these suites when attempting to load translations.


james-d-elliott (approved):
LGTM

PR #3218, opened 2022-04-19
refactor(suites): validate totp inputs

nightah:
This change validates the inputs for the TOTP code entry.
This was previously discarded and left unvalidated during the move to rod from within the integration tests.

The change has been made on the back of feedback in https://github.com/go-rod/rod/issues/497.


james-d-elliott (approved):
LGTM

PR #3307, opened 2022-05-06
docs: caddy integration

### LINKED ISSUE #1241: Authelia + Caddy v2 Configuration?
Hello!
I noticed that Caddy v2 doesn't have a proxy integration on Authelia docs: https://www.authelia.com/docs/deployment/supported-proxies/

Are you planning on adding one?

james-d-elliott:
This adds docs on integration with Caddy.

Closes #1241


mholt:
Awesome!

Do you think we can actually wait on merging this until we have the shorthand `forward_auth` syntax ready? I think it will be good if people use that right out of the gate, and they'll be less intimidated too. Kind of like how we recommend people use `php_fastcgi` instead of manually configuring the [expanded form](https://caddyserver.com/docs/caddyfile/directives/php_fastcgi#expanded-form) manually.

james-d-elliott:
> Awesome!
> 
> Do you think we can actually wait on merging this until we have the shorthand `forward_auth` syntax ready? I think it will be good if people use that right out of the gate, and they'll be less intimidated too. Kind of like how we recommend people use `php_fastcgi` instead of manually configuring the [expanded form](https://caddyserver.com/docs/caddyfile/directives/php_fastcgi#expanded-form) manually.

Hey @mholt that was the plan. The short (recommended) form is just a placeholder for now. While we include the long form we're going to encourage the short form (and discourage the long form, probably more aggressively than we already do) as configuration issues leading to security problems are easy to achieve. In addition resolving things that are missing/problems from the short form should not be hard now that we have an open channel.

I would like to get a more complete example with mTLS and trusted proxies, as security is a key part of the Authelia ecosystem. But I will spend some time this weekend looking into that and ask in the chat if needed.

docs/deployment/supported-proxies/caddy.md
```diff
@@ -0,0 +1,127 @@
...
+
+[Caddy] is a reverse proxy supported by **Authelia**. 
+
+_**Important:** Caddy officially supports the forward auth flow in version 2.5.1 and greater. You must be using this 
+version in order to use either Caddyfile. If you'd like to use it before 2.5.1 is released you need to build 
+[this branch](https://github.com/caddyserver/caddy/pull/4739)._ 
+
+Authelia offers integration support for the official forward auth integration method Caddy provides, we
+can't reasonably be expected to offer support for all of the different plugins that exist. As we have direct contact
+with the Caddy developers and their integration being written so well solving issues should be relatively straightforward.
```
nightah, line 17:
```suggestion
can't reasonably be expected to offer support for all of the different plugins that exist.
```

Not sure if we need this additional sentence, if you think we do probably something like this:

> As we have direct contact with the Caddy developers and their integration is written so well, solving issues should be relatively straightforward.

nightah (approved):
More of a WDYT as opposed to a hard requirement for change.

PR #3330, opened 2022-05-10
feat(logging): allow time replacement in log file name

### LINKED ISSUE #3210: Rotate file log
### Description

Can some configuration be done to rotate the authelia log file (perhaps with something like `logrotate`) periodically, or when it gets too large?
At the moment a single file is used, and it can grow quite a bit over time. For example, this is over the past 2-3 months:
image

I'm using the official docker container (`authelia/authelia`), but I'm guessing it's the same situation for the standalone package.

### Use Case

_N/A_

james-d-elliott:
This allows replacing %d with a date time format in the log file_name option.

Closes #3210


nightah (approved):
LGTM

PR #3342, opened 2022-05-10
fix(web): loading from root instead of base path

### LINKED ISSUE #3341: Locales requested from server root when running from subpath (www.example.com/authelia_portal/)
## Bug Report
Hi, first thanks for all the amazing work and this project!

### Description

I'm running authelia (portal) in from a subpath, like `www.example.com/authelia_portal/` using nginx. I have set the

``
server:
  path: "authelia_portal"
``

variable in the config.

Everthing ostensibly runs fine, but I see log entries from the access.log of the service running on the root (`www.example.com/`) location of the form

`- [IP] [09/May/2022:17:43:36 +0200] "GET /locales/en/portal.json HTTP/2.0" 404 2720 "https://example.com/authelia_portal" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:100.0) Gecko/20100101 Firefox/100.0" 0.380`

whenever someone logs in via authelia. Also strings are not localized, remaining English instead of German in this case.
Introducing an extra location block for /locales/ and proxying to `http://localhost:[autheliaPort]/locales/` resolves the issue.

### Expected Behaviour

Locales are requested from the subfolder domain.

### Reproduction Steps

Run authelia from a subfolder domain (like `www.example.com/authelia_portal/`), alter config accordingly and check access.log of the root (`www.example.com/`) for 404 errors.

### Additional Information

_N/A_

james-d-elliott:
This fixes an issue where we were loading the locales from the root rather than the base path.

Fixes #3341


nightah (approved):
LGTM.

PR #3437, opened 2022-05-26
fix(commands): hash-password usage instructions

james-d-elliott:
This fixes the hash-password usage instructions and ensures it uses mostly a configuration source based config.


nightah (approved):
LGTM.

PR #3466, opened 2022-06-03
test(web): fix tests to conform to react18

nightah:
This change drops the use of ReactDOM which is not supported in React 18. We also fix any test warnings related to i18next.


james-d-elliott (approved):
LGTM

PR #3479, opened 2022-06-04
ci: add integration containers for duo and haproxy

nightah:
This change utilises specific integration containers for the DuoPush and HAProxy suites.
In the case of DuoPush suite specifically in dev mode the container will be built on suite startup.


james-d-elliott (approved):
LGTM

PR #3507, opened 2022-06-11
fix(server): handled errors not logged correctly

### LINKED ISSUE #3506: HTTP 431 for unknown reason
### Description

Authelia v4.35.6 & Nginx Proxy Manager v2.9.18 (Docker) on Unraid v6.10.2

A couple hosts return HTTP 500 to the browser. Upon inspecting the logs, Authelia is returning HTTP 431 causing the server-side error. It only happens for certain hosts from certain browsers (more details below).

NPM log snippet (first line error.log, second line access.log)
```
2022/06/10 17:35:51 [error] 13601#13601: *567457 auth request unexpected status: 431 while sending to client, client: 192.168.1.5, server: server2.service1.mydomain.com, request: "GET /login HTTP/2.0", host: "server2.service1.mydomain.com"
[10/Jun/2022:17:35:51 -0600] - - 500 - GET https server2.service1.mydomain.com "/login" [Client 192.168.1.5] [Length 576] [Gzip -] [Sent-to 192.168.1.2] "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/537.36" "-"
```

I can confirm that removing Authelia from the Nginx config on this proxy host makes it function as it should.

#### The weird part

This only happens for certain browsers on certain machines for certain domains. For example:
- server2.service1.mydomain.com does not work from my **home** desktop, but it does work from my phone (on same and different networks) and from my work desktop.
- server1.service1.mydomain.com does not work from my **work** desktop, but it does work from my phone and from my home desktop.

In either case, Chrome in Guest or Incognito mode **does** work.

### Expected Behaviour

_N/A_

### Reproduction Steps

Unsure, but I'm willing to help troubleshoot.

### Additional Information

- So far, both hosts that have this problem are new additions to NPM & Authelia configuration.
- Both use the wildcard *.service1.mydomain.com in the configuration.yml
- server1.service1... and server2.service1... were added when I got a second server, server1 has always been there and was previously configured as just service1.mydomain.com (which is now a redirection host to server1.service1...)
- Other, working additions to NPM & Authelia config were made at/around the same time, including ones using the wildcard *.service2...

james-d-elliott:
This fixes an issue where errors handled by the ErrorHandler were not correctly logged. It also ensures the errors are logged with fields to make them easy to diagnose.

Fixes #3506

### DIFF
--- a/internal/server/handlers.go
+++ b/internal/server/handlers.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"fmt"
 	"net"
 	"os"
 	"strconv"
@@ -9,6 +10,7 @@ import (
 
 	duoapi "github.com/duosecurity/duo_api_golang"
 	"github.com/fasthttp/router"
+	"github.com/sirupsen/logrus"
 	"github.com/valyala/fasthttp"
 	"github.com/valyala/fasthttp/expvarhandler"
 	"github.com/valyala/fasthttp/pprofhandler"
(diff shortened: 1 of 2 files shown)


nightah (approved):
LGTM.

PR #3531, opened 2022-06-16
feat(storage): add unix socket support for mysql provider

LongerHV:
I would like to run Authelia as a SystemD service and connect to a local mysql database via an unix socket
(in a similar way this can be done with redis).

### DIFF
--- a/internal/configuration/validator/storage_test.go
+++ b/internal/configuration/validator/storage_test.go
@@ -55,14 +55,20 @@ func (suite *StorageSuite) TestShouldValidateLocalPathIsProvided() {
 	suite.Require().Len(suite.validator.Errors(), 0)
 }
 
+// MySQL over TCP.
 func (suite *StorageSuite) TestShouldValidateMySQLHostUsernamePasswordAndDatabaseAreProvided() {
-	suite.config.MySQL = &schema.MySQLStorageConfiguration{}
+	suite.config.MySQL = &schema.MySQLStorageConfiguration{
+		SQLStorageConfiguration: schema.SQLStorageConfiguration{
+			Port: 3306,
+		},
+	}
 	ValidateStorage(suite.config, suite.validator)
 
-	suite.Require().Len(suite.validator.Errors(), 3)
+	suite.Require().Len(suite.validator.Errors(), 4)
 	suite.Assert().EqualError(suite.validator.Errors()[0], "storage: mysql: option 'host' is required")
-	suite.Assert().EqualError(suite.validator.Errors()[1], "storage: mysql: option 'username' and 'password' are required")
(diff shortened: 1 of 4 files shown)


internal/configuration/validator/storage_test.go
```diff
@@ -55,13 +55,44 @@
 	suite.Require().Len(suite.validator.Errors(), 0)
 }
 
+// MySQL over TCP
```
autheliabot, line 58: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Comment should end in a period (godot)

internal/configuration/validator/storage_test.go
```diff
@@ -55,13 +55,44 @@
...
+			Database: "database",
+		},
+	}
+	ValidateStorage(suite.config, suite.validator)
+
+	suite.Require().Len(suite.validator.Warnings(), 0)
+	suite.Require().Len(suite.validator.Errors(), 0)
+}
+
+// MySQL over unix socket
```
autheliabot, line 88: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Comment should end in a period (godot)

james-d-elliott:
Just taking a quick look, I'd recommend  if you want to do this to not utilize the port at all for the check. Instead check the host for a prefix of `unix:`, this way the change is not breaking.

LongerHV:
@james-d-elliott thank you for the feedback. I initially wanted to check for simply `/` prefix in th host field, but I saw that redis session provider has support for unix sockets and it checks if the port is set to 0. If the prefix based solution is preffered, I'll do just that.

james-d-elliott:
> @james-d-elliott thank you for the feedback. I initially wanted to check for simply `/` prefix in th host field, but I saw that redis session provider has support for unix sockets and it checks if the port is set to 0. If the prefix based solution is preffered, I'll do just that.

Good spot! I may have to take a closer look and reconsider how we do this on a broader scale. Honestly I wish I could just change them all to an address pattern. Maybe a better approach is to add the address field and if it's not specified it's built assuming TCP, if it isn't it's expected to be the correct TCP or UNIX format.

james-d-elliott:
Sorry about getting back to you so late. I'm implementing this in #4231 for both SQL providers.


Thanks for your contribution it's appreciated regardless, @all-contributors please add @LongerHV for features

james-d-elliott:
@all-contributors please add @LongerHV for features

james-d-elliott:
https://github.com/all-contributors please add @LongerHV for ideas

james-d-elliott:
@all-contributors please add @LongerHV for ideas and code

PR #3586, opened 2022-06-24
docs: use details shortcode

james-d-elliott:


nightah (changes requested):
Looks like you missed the closing tags for the Traefik page and I think it's worth removing the Secret header from integration/kubernetes/introduction given that it's covered explicitly.

nightah (approved):
LGTM.

PR #3628, opened 2022-06-29
refactor(middlewares): factorize responses

james-d-elliott:


nightah (commented):
I think you'll also need to adjust the OpenAPI spec for some of the responses.

james-d-elliott:
> I think you'll also need to adjust the OpenAPI spec for some of the responses.

I had a quick look and we only document the application/json content from what it looks like. I'll have a quick look to see where these methods are used and add some docs maybe as part of this.

PR #3653, opened 2022-07-02
fix(configuration): address parsing failure

james-d-elliott:
This fixes an issue with parsing address types from strings.


nightah (commented):
I had these sitting in review, though might need to just address in a different PR?

PR #3662, opened 2022-07-05
refactor: factorize verify handler

james-d-elliott:


internal/handlers/handler_verify.go
```diff
@@ -155,40 +147,30 @@ func verifySessionCookie(ctx *middlewares.AutheliaCtx, targetURL *url.URL, userS
...
 
-		if inactiveLongEnough {
-			// Destroy the session a new one will be regenerated on next request.
-			err := ctx.Providers.SessionProvider.DestroySession(ctx.RequestCtx)
-			if err != nil {
-				return "", "", nil, nil, authentication.NotAuthenticated, fmt.Errorf("unable to destroy user session after long inactivity: %s", err)
-			}
-
-			ctx.Logger.Warnf("User %s has been inactive for too long", userSession.Username)
+		ctx.Logger.Warnf("Session destroyed for user '%s' after the session being inactive longer than the maximum timeout and the session not being marked as remembered", userSession.Username)
```
nightah, line 159:
```suggestion
		ctx.Logger.Warnf("Session destroyed for user '%s' after the session being inactive longer than the maximum timeout", userSession.Username)
```

Do we need to say anything about the session being remembered? Because we're assuming the remembered duration is long, it could actually still be sufficiently short enough that inactivity can still kick in.

nightah (changes requested):
Mostly LGTM, the requested change is more of a WDYT?

PR #3769, opened 2022-07-30
Grafana Integration Doc Changes

will-scargill:
When integrating Grafana with Authelia I encountered a problem where Grafana wasn't passing Authelia the correct redirect uri resulting in the following error.

`The 'redirect_uri' parameter does not match any of the OAuth 2.0 Client's pre-registered redirect urls`

Making the following change to my grafana.ini config fixed the issue for me.


james-d-elliott (approved):
LGTM

PR #3799, opened 2022-08-07
fix(utils): domain suffix improperly checked

james-d-elliott:
This fixes an issue where a domain of example.com would be considered to be a suffix of the URL https://app.notexample.com.


nightah (approved):
LGTM.

PR #3855, opened 2022-08-17
feat(notification): add disable_start_tls option

mind-ar:
this PR adds `disable_start_tls:true` to notification section


internal/configuration/schema/notifier.go
```diff
@@ -23,6 +23,7 @@ type SMTPNotifierConfiguration struct {
 	StartupCheckAddress mail.Address  `koanf:"startup_check_address"`
 	DisableRequireTLS   bool          `koanf:"disable_require_tls"`
 	DisableHTMLEmails   bool          `koanf:"disable_html_emails"`
+	DisableStartTLS     bool          `koanf:"disable_start_tls"`
```
james-d-elliott, line 26:
```suggestion
	DisableStartTLS     bool          `koanf:"disable_starttls"`
```

Really unsure which is better, I'm just leaving this here for discussion. We also need to add documentation to this, similar to my log comment.

internal/notification/smtp_notifier.go
```diff
@@ -156,6 +156,12 @@ func (n *SMTPNotifier) dial() (err error) {
 
 // Do startTLS if available (some servers only provide the auth extension after, and encryption is preferred).
 func (n *SMTPNotifier) startTLS() error {
+	// Skips STARTTLS if is disabled in configuration.
+	if n.config.DisableStartTLS {
+		n.log.Warn("STARTTLS configuration is set to disable STARTTLS requirement (only useful for unauthenticated emails over plain text)")
```
james-d-elliott, line 161:
```suggestion
		n.log.Warn("Notifier SMTP connection has opportunistic STARTTLS explicitly disabled which means all emails will be sent insecurely over plain text and this setting is only necessary for non-compliant SMTP servers which advertise they support STARTTLS when they do not")
```

		n.log.Debugf("Notifier SMTP connection is already encrypted, skipping STARTTLS")

james-d-elliott: Also maybe we should be logging this in the config validation or in the SMTP newup?

james-d-elliott (changes requested):
LGTM, only a few changes. We plan to add this to 4.37.

internal/configuration/schema/notifier.go
mind-ar: applied suggested change

docs/content/en/configuration/notifications/smtp.md
```diff
@@ -132,6 +133,13 @@ to leave this as is, but you can customize it if you have issues or you desire t
 For security reasons the default settings for Authelia require the SMTP connection is encrypted by TLS. See [security]
 for more information. This option disables this measure (not recommended).
 
+### disable_starttls
+
+{{< confkey type="boolean" default="false" required="no" >}}
+
+For security reasons the default settings for Authelia require the SMTP connection do use STARTTLS when is available. See [security]
```
mind-ar, line 140: @james-d-elliott ,  i need suggestion for this text

docs/content/en/configuration/notifications/smtp.md
```diff
@@ -132,6 +133,13 @@ to leave this as is, but you can customize it if you have issues or you desire t
...
+for more information. This option disables this measure (not recommended).
```
james-d-elliott, line 141:
```suggestion
Some SMTP servers ignore SMTP specifications and claim to support STARTTLS when they in fact do not.
For security reasons Authelia refuses to send messages to these servers.
This option disables this measure and is enabled  *__AT YOUR OWN RISK__*. It's *__strongly recommended__* 
that instead of enabling this option you either fix the issue with the SMTP server's configuration or 
have the administrators of the server fix it. If the issue can't be fixed by configuration we recommend 
lodging an issue with the authors of the SMTP server.
See [security] for more information.
```

docs/content/en/configuration/notifications/smtp.md
james-d-elliott: Left a review! LGTM btw.

james-d-elliott (approved):

internal/suites/Standalone/configuration.yml
```diff
@@ -95,6 +95,8 @@ notifier:
     port: 1025
     sender: admin@example.com
     disable_require_tls: true
+    # disables STARTTLS (not recommended)
+    disable_starttls: false
```
james-d-elliott, line 99:
```suggestion
```

james-d-elliott (approved):

PR #3858, opened 2022-08-18
ci: add eslint rule for sorting imports

nightah:
The [ignoreDeclarationSort](https://eslint.org/docs/latest/rules/sort-imports#ignoredeclarationsort) and [memberSyntaxSortOrder](https://eslint.org/docs/latest/rules/sort-imports#membersyntaxsortorder) options have been turned on and omitted respective to not conflict with the existing import/order rules.

### DIFF
--- a/web/.eslintrc.js
+++ b/web/.eslintrc.js
@@ -30,5 +30,14 @@ module.exports = {
                 },
             },
         ],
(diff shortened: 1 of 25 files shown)


james-d-elliott (approved):
LGTM

PR #3949, opened 2022-09-03
docs: Clarify that Chrome did not deprecate WebAuthn

polarathene:
Prevents communicating a misleading deprecation of WebAuthn, when it's actually referring to the (_now legacy_) U2F API that Chrome is dropping (_and has already removed_).

---

**Side-note:**

Your CI reacted to a minor update to docs and looks like it's running the whole suite of jobs that has nothing to do with docs.

### DIFF
--- a/docs/content/en/roadmap/active/webauthn.md
+++ b/docs/content/en/roadmap/active/webauthn.md
@@ -14,7 +14,7 @@ aliases:
   - /r/webauthn
 ---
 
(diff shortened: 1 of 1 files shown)


james-d-elliott:
That's definitely not the whole suite. The CI pipeline is dynamic. Here's a full example: https://buildkite.com/authelia/authelia/builds/13925

james-d-elliott (approved):

PR #3950, opened 2022-09-03
docs: Clarify port 465 is the preferred port

polarathene:
As per the 2018 RFC 8314, this port is no longer intended for SMTPS, but repurposed as a _submissions service_ port. Where available, it is advised as the port that should be preferred over any port which uses STARTTLS.

Note that as ports intended only for mail submission (ports 465 & 587), authentication is required to send mail through. Otherwise, use the standard SMTP port 25.

---

There was a [previous discussion](https://github.com/authelia/authelia/issues/2194) in 2021 regarding this. It looks like the config docs for the SMTP port was updated to acknowledge that, but these security docs have not been.


james-d-elliott (changes requested):
Mostly LGTM, please adjust in accordance with the [style guidelines](https://www.authelia.com/contributing/development/guidelines-style/#line-length).

polarathene:
> please adjust in accordance with the [style guidelines](https://www.authelia.com/contributing/development/guidelines-style/#line-length).

I can do this, although I don't understand why you want to enforce that.

I worked on the document with a narrow width and personally found it frustrating that text was being split instead of leveraging word-wrap that any sane text editor should offer?

---

From the same UX perspective the 120 line width is intended for, it demands users with views more narrow than 120 chars drop word-wrap and scroll horizontally, or deal with equally horrid experience like this (_which is not pleasant to edit those lines with_):

Screenshot_20220903_191323

With word-wrap off (100 char width view):

Screenshot_20220903_192144

With line width breaks removed and word-wrap on (100 char width view):

Screenshot_20220903_192353

Even at a view width of 120 columns/chars, for some content I'd still be forced to turn word-wrap off and scroll beyond that, but that's an opt-in choice, which I can toggle between. Word-wrap can be toggled, whereas enforced line widths not so much:

Screenshot_20220903_191741

Since this is rendered and typically viewed via live docs or a Markdown preview pane, **most viewers of the raw markdown** would be doing so with the intention of editing it. To encourage contributions, having a **pleasant editing experience** helps. I know that the fixed line width frustrated me while revising the content.

---

**I'm happy to respect your style guidelines** and make the requested change.

I am just curious whenever I encounter this as a mandatory requirement, since **I know of no-one that actually needs a fixed line width** that couldn't leverage word-wrap?

Even Githubs diff view for reviewers uses word-wrapping (_I don't see an option to disable it?_):

Screenshot_20220903_193043

---

The document already contains lines exceeding 120 chars (table excluded). Do you want those addressed as well, even though they're unrelated to this PR? Or would you like to tackle those separately?

This one is 126 in width:

Screenshot_20220903_193725

polarathene:
That should respect the 120 char line width requirement for this contribution. I've left the 4 lines that exceed this width but are unrelated to changes in this PR as-is, feel free to address them in this PR or separately :+1:

james-d-elliott:
So there are a multitude of reasons. There are sill a lot of IDE's/editors which don't automatically word wrap. This is because in coding generally it makes it harder to understand code due to structure being an integral part of the coding ux. It's nearly universally discouraged for code itself (not applicable to markdown, but relevant for later). 

In addition some more advanced IDE's offer merge resolution in a 3 pane window such as GoLand and all IntelliJ IDE's. This mode is less usable with word wrapping, but you can manage to fit 120 characters side by side by side.

It's also good for people with editors that don't support word wrapping, or don't turn it on for non-code related areas and leave it off for all, or for ones which are unable to do it per extension at all and it's only a global option.

Lastly it's easier for consistency that all areas that are intended to be readable are formatted similarly. This means guidelines are almost always the same regardless of the file you're editing.

james-d-elliott:
@all-contributors please add @polarathene for docs

polarathene:
> There are sill a lot of IDE's/editors which don't automatically word wrap.

Do you have any that come to mind that is relevant to the majority of contributors / maintainers?

- For prose, such as markdown content it seems reasonable to prefer word-wrap being enabled.
- While it may not be automatically enabled or on by default for content, usually it's supported by a menu click or keyboard shortcut, which is preferable to horizontal scrolling.
- With 120 char width linebreaks, users with smaller and larger widths are penalized with negative UX. Changes required for them to adjust is much more effort, or enforces them to maintain a minimum width of 120.

---

> This is because in coding generally it makes it harder to understand code due to structure being an integral part of the coding ux.
> It's nearly universally discouraged for code itself (not applicable to markdown, but relevant for later).

Agree :+1: 

But that content rarely benefits from such lengths due to it's nature. It's very different from written content containing paragraphs of text, other than perhaps comments.

---

> In addition some more advanced IDE's offer merge resolution in a 3 pane window such as GoLand and all IntelliJ IDE's.
> This mode is less usable with word wrapping, but you can manage to fit 120 characters side by side by side.

Depends on the changes.

AFAIK it really shouldn't be an issue with any decent merge or diff tool? I'd like to know where the advantages are for fixed line-width that doesn't outweigh the UX disadvantages compared to word-wrap?

**word-wrap (with before/after adjusted to demonstrate such diffs):**

Screenshot_20220906_142723

**120 char fixed width, no word-wrap:**

Screenshot_20220906_143019

Not seeing a striking difference there? (_with Meld, it has decent scrolling capabilities to align panes vertically regarding changes that would affect vertical offset if that was a concern_)

---

Not much to see if the fixed width leaves word-wrap off (_horizontal scrolling or unused white-space_), so next screenshots will enable word-wrap there to highlight the awkward UX this decision causes to word-wrap users (_not that it's much better UX for anyone not getting a good 120 char fit per pane on their display_):

**no line-breaks (word-wrap enabled, width 100 char):**

Screenshot_20220906_144002

**120 char line-breaks (word-wrap enabled, width 100 char):**

Screenshot_20220906_144135

**no line-breaks (word-wrap enabled, width 140 char):**

Screenshot_20220906_144307

**120 char line-breaks (word-wrap enabled, width 140 char):**

Screenshot_20220906_144350

---

Even at 120 char width for a pane at that font-size above, if a user needs a larger font for accessibility reasons, this completely messes up the experience for them as well, they basically have to get a much larger screen space, or use the OS support to zoom and pan around the display (_unlikely since the user likely prefers a larger system scale / font as a whole_):

**word-wrap no fixed-width line-breaks supports this fine:**

Screenshot_20220906_144622

**120 char fixed width formatting (word-wrap enabled):**

Screenshot_20220906_145054

**120 char fixed width formatting (word-wrap disabled, horizontal scroll required):**

Screenshot_20220906_145114

---

> It's also good for people with editors that don't support word wrapping, or don't turn it on for non-code related areas and leave it off for all, or for ones which are unable to do it per extension at all and it's only a global option.

I can only assume this is a personal preference of yours if you've chosen to enforce it. That's fine if it caters to the most active contributors / maintainers needs, their comfort should probably be prioritized.

However, if this is something just for the sake of debate / assumption - I hope that the examples provided above contrast how poorly that experience is under different situations, while word-wrap without fixed-width formatting is unaffected. All I can do is raise awareness about it :sweat_smile: 

> Lastly it's easier for consistency that all areas that are intended to be readable are formatted similarly.
> This means guidelines are almost always the same regardless of the file you're editing.

If you find that is more worthwhile than the disadvantages it imposes on certain content, that's fine :+1: 

---

Cheers for the merge and detailed reasoning! :grin:

PR #3956, opened 2022-09-04
fix(handlers): fix forbidden acl for anonymous user

### LINKED ISSUE #3084: Authelia asks for login even it is a policy deny rule
### Description

I am using the v4.34.6 which is latest as of writing, and here are my rules:

```
access_control:
  default_policy: deny

  networks:
    - name: internal
      networks:
        - 192.168.1.0/24

  rules:
    - domain: auth.domain.example
      policy: bypass

## this is the rule I am asking for
    - domain: s.domain.example
      resources: "^[/]?$"
      policy: deny

    ## Admin access to s.domain.example/admin*
    - domain: s.domain.example
      resources:
        - "^/admin.*"
      policy: two_factor
      subject:
        - "group:admin"

    - domain: s.domain.example
      resources: "^/.+"
      policy: bypass
```

### Expected Behaviour

When I have not login:
- When I access `s.domain.example` or `s.domain.example/`, I should immediately land on fobidden page, instead of being redirected to login page. After logging in here, then it shows the fobidden page.

When I have logged in:
- it should immediately land on fobidden page.

### Reproduction Steps

1. Use the above config
2. When not logged in, access `s.domain.example` or `s.domain.example/`
3. The login page will be shown, instead of fobidden page

1. Use the above config
2. When logged in, access `s.domain.example` or `s.domain.example/`
3. The fobidden page shown

mind-ar:
When an anonymous user tries to access a forbidden resource
with no subject, we should response with 403.
Fixes #3084

### DIFF
--- a/internal/authorization/authorizer.go
+++ b/internal/authorization/authorizer.go
@@ -54,7 +54,7 @@ func (p Authorizer) IsSecondFactorEnabled() bool {
 }
 
 // GetRequiredLevel retrieve the required level of authorization to access the object.
-func (p Authorizer) GetRequiredLevel(subject Subject, object Object) Level {
+func (p Authorizer) GetRequiredLevel(subject Subject, object Object) (bool, Level) {
 	logger := logging.Logger()
 
 	logger.Debugf("Check authorization of subject %s and object %s (method %s).",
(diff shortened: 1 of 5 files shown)


james-d-elliott (approved):
LGTM

PR #4037, opened 2022-09-20
fix(authorization): regex subj doesn't redirect anon user

### LINKED ISSUE #4038: Authz: Domain Regex / Resource criteria with users/groups doesn't redirect
### Version

v4.36.8, v4.36.7

### Deployment Method

Docker

### Reverse Proxy

Traefik

### Reverse Proxy Version

N/A

### Description

When using rules with subjects that are not part of the `subject` criteria such as `domain_regex` or `resources` with the regex pattern using named sub expression matcher for `User` or `Groups`, anonymous uses are not automatically redirected. Instead they are handed a 403 Forbidden response.

See: https://github.com/authelia/authelia/discussions/4036

### Reproduction

1. Define a rule with either a `domain_regex` criteria or `resources` criteria, and ensure it's the only rule for a domain.
2. Visit the site as an anonymous user.

### Expectations

_No response_

### Logs

_No response_

### Configuration

_No response_

### Documentation

_No response_

james-d-elliott:
This fixes an issue with the authorization policies where if the Domain Regex or Resources criteria would incorrectly return 403 Forbidden statuses instead of 302 Found statuses. It should also slightly improve performance for rules with `subject` criteria. 

Fixes #4038


internal/authorization/util.go
```diff
@@ -70,24 +70,42 @@ func schemaSubjectToACLSubject(subjectRule string) (subject SubjectMatcher) {
...
 	}
 
-	for _, domainRegexRule := range domainRegexRules {
-		domains = append(domains, NewAccessControlDomainRegex(domainRegexRule))
-	}
+	/*
+		for _, domainRegexRule := range domainRegexRules {
+			domains = append(domains, NewAccessControlDomainRegex(domainRegexRule))
+		}
+	*/
```
nightah, line 82:
```suggestion
```

I suspect this was accidentally left in.

nightah (changes requested):
Only a tiny nit, otherwise LGTM.

nightah (approved):
LGTM.

PR #4151, opened 2022-10-08
docs: update Seafile OIDC integration

philwilei:
I recently used this guide to integrate my Seafile with my Authelia. I ran into a couple issues along the way and updated the docs accordingly:
* Remove the `groups` scope from the Seafile config since Seafile doesn't use this information (as far as I can tell) and the given Authelia config doesn't even grant this scope.
* Fix the `OAUTH_ATTRIBUTE_MAP` in the Seafile config. When using the old values Seafile was reporting "Required user info not found" and login would fail.
* Update the tested versions.
* Add a note regarding an issue with the provided configuration and hints on how to mitigate. Also add a step for installing a required pip package.
* Reference the sources that helped me fix the issues I had.


docs/content/en/integration/openid-connect/seafile/index.md
```diff
@@ -89,6 +97,10 @@ which will operate with the above example:
 
 * [Seafile OAuth Authentication Documentation](https://manual.seafile.com/deploy/oauth/)
 
+[^1]:https://forum.seafile.com/t/oauth-question-error-with-sso/5481/2
+[^2]:https://forum.seafile.com/t/oauth-question-error-with-sso/5481/3
+[^3]:https://manual.seafile.com/deploy/oauth/#oauth
+
```
james-d-elliott, line 103:
```suggestion
```

docs/content/en/integration/openid-connect/seafile/index.md
```diff
@@ -52,18 +59,19 @@ OAUTH_AUTHORIZATION_URL = 'https://auth.example.com/api/oidc/authorization'
...
+    "openid",
+    "profile",
+    "email",
 ]
 OAUTH_ATTRIBUTE_MAP = {
-    "id": (True, "preferred_username"),
+    "email": (True, "email"),
     "name": (False, "name"),
-    "email": (False, "email"),
+    "id": (False, "not used"),
```
james-d-elliott, line 69:
```suggestion
```

docs/content/en/integration/openid-connect/seafile/index.md
```diff
@@ -33,13 +33,20 @@ This example makes the following assumptions:
...
+the `OAUTH_ATTRIBUTE_MAP` or modifying `seahub/seahub/oauth/views.py` in your [Seafile] installation[^2].*
+
 ## Configuration
 
 ### Application
 
 To configure [Seafile] to utilize Authelia as an [OpenID Connect] Provider:
 
-1. Edit your [Seafile] `seahub_settings.py` configuration file and add configure the following:
+1. Install the `requests_oauthlib` pip package which is required but not installed with [Seafile] by default[^3].
```
james-d-elliott, line 47:
```suggestion
1. Install `requests_oauthlib` which is required but not installed with [Seafile] by default. See the [Seafile] documentation in the [see also](#see-also) section for more information.
```

docs/content/en/integration/openid-connect/seafile/index.md
```diff
@@ -33,13 +33,20 @@ This example makes the following assumptions:
 * __Client ID:__ `seafile`
 * __Client Secret:__ `seafile_client_secret`
 
+*__Important Note:__ [Seafile] only uses email to identify unique user accounts[^1]. When using the
+[File First Factor](../../../configuration/first-factor/file.md) there may be multiple Authelia users with the same
+email, resulting in them being assigned the same [Seafile] account. This issue could be mitigated by tuning
+the `OAUTH_ATTRIBUTE_MAP` or modifying `seahub/seahub/oauth/views.py` in your [Seafile] installation[^2].*
+
```
james-d-elliott, line 40:
```suggestion
```

docs/content/en/integration/openid-connect/seafile/index.md
```diff
@@ -52,18 +59,19 @@ OAUTH_AUTHORIZATION_URL = 'https://auth.example.com/api/oidc/authorization'
...
-    "id": (True, "preferred_username"),
+    "email": (True, "email"),
     "name": (False, "name"),
-    "email": (False, "email"),
+    "id": (False, "not used"),
 }
 ```
 
+Remember to restart [Seafile] so your configuration changes take effect.
+
```
james-d-elliott, line 74:
```suggestion
```

james-d-elliott (changes requested):
Thanks for choosing to contribute! I really appreciate it. I understand the motivation but I think most of this should be part of the Seafile documentation not our documentation.

Can you check if the removal of the `id` line in the OAUTH_ATTRIBUTE_MAP works and let me know? In addition I've included several changes which remove sections that should be part of the Seafile docs.

philwilei:
Thanks for the prompt feedback on my PR.  I experimented with removing the `id` mapping, however Seafile sets a default value which causes authentication to break when I don't overwrite it. There may be a better solution, but I am no OAuth expert and the Seafile docs [suggest](https://manual.seafile.com/deploy/oauth/#__code_6) doing it this way so I would consider it acceptable.

james-d-elliott:
> Thanks for the prompt feedback on my PR. I experimented with removing the `id` mapping, however Seafile sets a default value which causes authentication to break when I don't overwrite it. There may be a better solution, but I am no OAuth expert and the Seafile docs [suggest](https://manual.seafile.com/deploy/oauth/#__code_6) doing it this way so I would consider it acceptable.

Yep perfect, that's the feedback I was after. Pitty they're not more explicit about it.

james-d-elliott (approved):
LGTM!

PR #4243, opened 2022-10-23
fix(utils): fix pre-commit scripts compatibility with posix shells

smkent:
The existing pre-commit and required-apps scripts invoke `/bin/sh` but utilize non-POSIX features such as `[[` conditionals. These scripts fail on systems such as Ubuntu Linux where `/bin/sh` ([dash, in this case](https://wiki.ubuntu.com/DashAsBinSh)) only supports POSIX features. This change replaces the usage of non-POSIX features with POSIX-compatible shell commands.


james-d-elliott (approved):
LGTM

PR #4373, opened 2022-11-14
docs: clarify cloudflare docs

james-d-elliott:


ja0nz:
As an alternative to this approach I suggest to include a reference to Cloudfares [authenticated origin pull](https://developers.cloudflare.com/ssl/origin-configuration/authenticated-origin-pull/). It seems to tackle the same issue namely ensuring the call to authelia comes from trusted sources, Cloudfare in this case. But feels more idiomatic to me.

PR #4467, opened 2022-12-04
Updated the documentation

jonaslewin:
Updated the documentation with the missing user keyword


james-d-elliott (approved):
LGTM. Better option may be to link to the CLI docs directly instead though. Let me know if you'd like to just merge as is however.

jonaslewin:
I'd suggest merging as is at this point.

PR #4483, opened 2022-12-06
feat(authentication): ldap time replacements

### LINKED ISSUE #1284: Support for password expiration
I am using the LDAP integration to connect Authelia to FreeIPA. I noticed that even after a user's password expires they're still able to log in to services protected by authelia. In addition using authelia to reset a users password does not reset it's expiry information.

I would like for users with expired passwords to not be able to log in and then properly reset their password through authelia which would "unexpire" their password. 

I'm not sure if there's a workaround/feature I'm missing here or if authelia doesn't currently support this. 

Cheers!

### LINKED ISSUE #1964: ldap: date time restriction
Authelia is awesome, 
I have a suggestion :
Is it possible to add a date time restriction based on a ldap field?, there is no support for date time Restriction auth on openldap so this might be a good idea.
You could use Unix time stamp or a date time schema defined in rules like d/MM/yyyy h:m, the data could be retrieved on ldap groups or users with a preference for the most restrictive date time.
You could eventually implement a start and end date validation as ldap fields...

thanks for reading

james-d-elliott:
This adds and utilizes several time replacements for both specialized LDAP implementations.

Closes #1964, Closes #1284


internal/regulation/regulator_test.go
```diff
@@ -5,6 +5,7 @@
 	"testing"
 	"time"
 
+	"github.com/authelia/authelia/v4/internal/utils"
```
autheliabot, line 8: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>File is not `goimports`-ed with -local github.com/authelia/authelia (goimports)

PR #4525, opened 2022-12-09
fix: handle null devices list get response properly

smkent:
When a user opens the webauthn devices settings page without any devices configured (e.g. on initial enrollment), the UI loading bar stays indefinitely. This is due to the user devices endpoint returning `null` instead of an empty list, which the UI does not handle gracefully. This change adds a `GetWithOptionalResponse` function in the UI to handle this case, similar to the existing `PostWithOptionalResponse` method in the same file.

### DIFF
--- a/web/src/services/Client.ts
+++ b/web/src/services/Client.ts
@@ -19,16 +19,19 @@ export async function Post<T>(path: string, body?: any) {
     return res;
 }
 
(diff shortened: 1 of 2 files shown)


smkent:
Closing in favor of https://github.com/authelia/authelia/pull/4537

PR #4640, opened 2022-12-25
test: use `T.Setenv` to set env vars in tests

Juneezee:
A testing cleanup.

This PR replaces `os.Setenv` with `t.Setenv`. Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

This saves us at least 2 lines (error check, and cleanup) on every instance.

Reference: https://pkg.go.dev/testing#T.Setenv

```go
func TestFoo(t *testing.T) {
	// before
	os.Setenv(key, "new value")
	defer os.Unsetenv(key)
	
	// after
	t.Setenv(key, "new value")
}
```

### DIFF
--- a/internal/commands/util_test.go
+++ b/internal/commands/util_test.go
@@ -61,7 +61,7 @@ func TestLoadXEnvCLIStringSliceValue(t *testing.T) {
 			}
 
 			if tc.envValue != "" {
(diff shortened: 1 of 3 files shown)


james-d-elliott (approved):
LGTM

PR #4647, opened 2022-12-26
fix(notification): smtp auth not configured

james-d-elliott:
This fixes an issue introduced by the pending migration to the new SMTP library in 0bb657e11cfa6b9f16cd28340291b121fa1579bc where the auth mechanism is never defined. This only affects commits that are yet to be versioned.

### DIFF
--- a/internal/notification/smtp_auth.go
+++ b/internal/notification/smtp_auth.go
@@ -1,0 +1,81 @@
+package notification
+
+import (
+	"fmt"
+	"net/smtp"
+	"strings"
+
+	"github.com/authelia/authelia/v4/internal/configuration/schema"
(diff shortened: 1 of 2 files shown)


internal/notification/smtp_auth.go
```diff
@@ -0,0 +1,81 @@
+package notification
+
+import (
+	"fmt"
+	"net/smtp"
+	"strings"
+
+	"github.com/authelia/authelia/v4/internal/configuration/schema"
```
autheliabot, line 8: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>File is not `goimports`-ed with -local github.com/authelia/authelia (goimports)

PR #4665, opened 2022-12-30
Fix Directory structure in SWAG Documentation

cdestefano:
Updated SWAG Reverse Proxy documentation to map to the same directory using a docker bind mount of swag (/config) to allow the commands to be used without manual directory moving / updating after creating each file.


james-d-elliott (approved):

cdestefano:
Thanks for making my first update go smoothly!

Issue #4669, opened 2022-12-31
CSP blocks javascript on the oidc autorization page


levkoburburas:
### Version

v4.37.5

### Deployment Method

Other

### Reverse Proxy

NGINX

### Description

This code:
https://github.com/authelia/authelia/blob/633680e8e5e8188b0f5667b5c77eb692f0e58a0d/internal/server/handlers.go#L139
applies Content-Security-Policy headers with "default-src 'none'" for all oidc pages generated by `github.com/ory/fosite`

which in turn blocks the execution of the inline javascript which redirects users back to the client:
https://github.com/ory/fosite/blob/e570564951ec312d5dc1c1c21180ca3e759c5075/authorize_helper.go#L24

![image](https://user-images.githubusercontent.com/62853952/210153942-575751c0-e226-42cf-a286-817e9be0c360.png)

The workaround is to remove "Content-Security-Policy" is assigned by server and set the "correct value" in the `nginx proxy` configuration. 

```
proxy_hide_header Content-Security-Policy;
add_header Content-Security-Policy "default-src 'self' 'unsafe-inline';" always;
```

### Reproduction

Use **OpenID Connect**  with a client that supports only  `form_post`  response_mode.

### Logs (Authelia)

```shell
identity_providers:
  oidc:
    clients:
      -
        response_modes:
          - form_post

```
```


### Logs (Proxy / Application)

_No response_

### Documentation

_No response_

### Pre-Submission Checklist

- [X] I agree to follow the [Code of Conduct](http://www.authelia.com/code-of-conduct)
- [X] This is a bug report and not a support request
- [X] I have read the security policy and this bug report is not a security issue or security related issue
- [X] I have either included the complete configuration file or I am sure it's unrelated to the configuration
- [X] I have provided all of the required information in full with the only alteration being reasonable sanitization in accordance with the [Troubleshooting Sanitization](https://www.authelia.com/r/sanitize) reference guide
- [X] I have checked for related proxy or application logs and included them if available
- [X] I have checked for related issues and checked the documentation

levkoburburas:
Pull request #4670 which solves this issue.

james-d-elliott:
That's unfortunate, unsafe inline is not really an answer. I appreciate the PR however it's likely this can be entirely handled by the `FormPostHTMLTemplateProvider` without the need for any additional machinery except for setting the header itself (probably with a sha256).

levkoburburas:
[...]

PR #4670, opened 2022-12-31
feat(server): custom oidc authorization page for `form_post` response method

levkohimins:
This pull request solves two issues:

1. Applies correct [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) to the oidc authorization page, see related issue [#4669](https://github.com/authelia/authelia/issues/4669).
2. Customize oidc authorization page (text with animation).  

Redirecting to Bitwarden Password Manager - Authelia 2022-12-31 at 8 54 38 PM


james-d-elliott (changes requested):
Thanks for the PR. However I'd rather this is solved by adjusting the template returned from `fosite.FormPostHTMLTemplateProvider` using sha256 hashes instead (in addition to making a special CSP middleware for the authorization endpoint which adds the sha256's when the response mode is `form_post`).

levkoburburas:
> Thanks for the PR. However I'd rather this is solved by adjusting the template returned from `fosite.FormPostHTMLTemplateProvider` using sha256 hashes instead (in addition to making a special CSP middleware for the authorization endpoint which adds the sha256's when the response mode is `form_post`).


Thank you for the review!

This could be a great solution, if only this approach also supported passing variables. It is important for me to display the clientID/clientDescription, in fact, where the redirect goes.

I will try to make code changes in `fosite`, if they accept them, I will redo this PR, otherwise this approach is preferable for me.

james-d-elliott:
I don't think it's necessary to implement the client description in the short term. However I'm relatively certain they'd be open to implementing an interface addition to the configurator with a signature similar to below:

```go
type AuthorizeFormPostResponseProvider interface {
    GetWriteAuthorizeFormPostResponse(ctx context.Context) func(redirectURL string, parameters url.Values, template *template.Template, rw io.Writer)
}
```

levkoburburas:
I liked your first suggestion, because it seems to me that it gives more opportunities

```go
type FormPostResponseProvider interface {
    GetWriteFormPostResponse(ctx context.Context) func(rw io.Writer, template *template.Template, redirectURL string, parameters url.Values)
}
```

I created the PR [#731](https://github.com/ory/fosite/pull/731) with discussed changes.

> If you'd like I can just push the fix and you can work on the nice to have, I'll leave it up to you however.

Thank you, but it's up to you. I work with forked repo, it doesn't bother me.

levkoburburas:
By the way, I'm not very good at frontend, perhaps you should that page it and improve something at your discretion.

james-d-elliott:
We've moved to our own fork of fosite, so we can probably fix this up in a cleaner way. Was thinking of doing this for 4.39 but it seems too far diverged. Let me know if you're still interested in a custom form post page.

PR #4672, opened 2023-01-01
feat(server) nt password authentication in ldap backend

levkoburburas:
If you are using FreeAPI as your ldap backend and two-factor authentication is enabled **in** FreeAPI, the users will not be able to authenticate using login and password on the Authelia login page when `authentication_backend` is set to `bind`. 

The reason is that two-factor authentication is also used for binding in [389 Directory Server](https://directory.fedoraproject.org/) that FreeAPI uses. Depending on the type of two-factor authentication activated in FreeIPA, if it's OTP, then the user must input Password+OTP on the Authelia login page.

The solution is to use the NT Password from user DN instead of trying to check the password by ldap binding.

The PR is adding new two config setting in ldap backend.

```yml
authentication_backend:
  ldap:
    authentication_method: 
    nt_password_attribute: 
```

*authentication_method* - takes `bind` or `nt_password`. By default `bind`.
*nt_password_attribute* - has a default value of `ipaNTHash` when *implementation* is set to `freeipa`.

If you choose to accept these changes, perhaps you should improve the documentation I added for them.


james-d-elliott:
I have to really think about this, I appreciate the PR I just have to figure out how we can best support something like this long term.

PR #4696, opened 2023-01-04
refactor(web): remove query-string package

nightah:
This change drops the redundant query-string package and utilises native react hooks from `react-router-dom`.


james-d-elliott (approved):
LGTM

PR #4714, opened 2023-01-06
Move a comment to the appropriate configuration item

chillinPanda:
Hello,

in the configuration.yml for the lite example a configuration option was moved but the comment for it was overlooked.

I just moved it to the appropriate line.


james-d-elliott:
@all-contributors please add @chillinPanda for docs

PR #4762, opened 2023-01-13
test(suites): load environment into suites

james-d-elliott:


internal/suites/suite_network_acl_test.go
```diff
@@ -10,11 +10,15 @@ import (
...
 type NetworkACLSuite struct {
-	suite.Suite
+	*BaseSuite
 }
 
 func NewNetworkACLSuite() *NetworkACLSuite {
-	return &NetworkACLSuite{}
+	return &NetworkACLSuite{
+		BaseSuite: &BaseSuite{
+			Name: traefikSuiteName,
```
nightah, line 19:
```suggestion
			Name: networkACLSuiteName,
```

internal/suites/utils.go
```diff
@@ -84,6 +85,57 @@ func (rs *RodSession) collectCoverage(page *rod.Page) {
...
+func (s *BaseSuite) SetupSuite() {
+	s.LoadEnvironment()
+}
+
+func (s *BaseSuite) LoadEnvironment() {
+	if s.Name == "" {
+		return
+	}
+
+	fmt.Printf("Checking Suite %s for .env file\n", s.Name)
```
nightah, line 97: Can we change all the logging here to be debug?

nightah (changes requested):
Couple of small changes.

Also I noticed the following:
* BypassAll, DuoPush, HighAvailability, Standalone suites all seem to run the load of env vars two or more times, probably because of the underlying scenario suites, we should prevent this.

* The following test seemed to error, not sure if that's related to this change or not:
```
=== RUN   TestOIDCSuite/TestOIDCScenario/TestShouldDenyConsent/error_uri
    verify_is_oidc.go:28: Test Skipped as the element is not expected.
```

james-d-elliott:
> * The following test seemed to error, not sure if that's related to this change or not:

Yeah this was because that suite checks the responses in a recursive way, the skip is for when the error is not expected.

> * BypassAll, DuoPush, HighAvailability, Standalone suites all seem to run the load of env vars two or more times, probably because of the underlying scenario suites, we should prevent this.

Do we want to maybe convert all of them into a single suite?

internal/suites/utils.go
```diff
@@ -84,6 +85,87 @@
 	}
 }
 
+func (s *BaseSuite) SetupSuite() {
+	s.SetupLogging()
+	s.SetupEnvironment()
+}
+
+func (s *BaseSuite) SetupLogging() {
+	if os.Getenv("SUITE_SETUP_LOGGING") == "true" {
```
autheliabot, line 94: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>string `true` has 3 occurrences, but such constant `t` already exists (goconst)

nightah (approved):
LGTM.

PR #4771, opened 2023-01-15
fix(utils): fix suite setup timeout

mind-ar:
After PR #4751 the suites times out if takes more of 60 seconds.
This PR honors the timeout configured at waitUntilAutheliaBackendIsReady (90s)


james-d-elliott (approved):
LGTM

Issue #3736, opened 2022-07-23
compartmentalize by running systemd unit as authelia user and group


feinedsquirrel:
### Description

I brought up in the matrix chat a long time ago that I could help transition the systemd unit to use the `user=` and `group=` keys. (I'm finally able to work on it.)
As I understand, best practice with systemd units is that we set those to "authelia", as by default the service would otherwise run as user and group "root". It isn't necessarily a vulnerability, but more of another layer in sandboxing (though don't quote me, I'm not an expert).

### Use Case

For anyone running Authelia as a systemd service, the process(?) would run as user and group "authelia" instead of "root", increasing sandboxing.

### steps to complete

I've got a working example on my machine, though it may not be perfect (as I said before, I'm not an expert).
What I see needing completed:

- [ ] add `user=authelia` and `group=authelia` to the systemd config (example can be seen in issue #568)
- [ ] add two files: `authelia.tmpfiles.conf` and `authelia.sysusers.conf`, placing them at `/usr/lib/sysusers.d/authelia.conf` and `/usr/lib/tmpfiles.d/authelia.conf`, respectively (at least for Arch Linux--I've got a functioning sample PKGBUILD)
- [ ] add to the redis documentation that if using this systemd unit, the `authelia` user needs to be added to the `redis` group (I suspect this may have been one of the problems the original poster in #568 was having, _if_ they were using redis)
- [ ] any other changes to be compatible with the debian, docker, and any other builds (I'm unfamiliar with these)

If this looks correct, I can create a WIP pull request from my fork, and testing and other alterations as needed can begin.
Let me know. Thanks!

PR #4982, opened 2023-02-23
run systemd unit as authelia user

feinedsquirrel:
supersedes pull request #3770 
because I have since learned how to better manage forks, and make them so I can rebase on upstream, rather than creating new commits that are simply merges of upstream changes.

I'm also removing [WIP], as this has been working for me for the past year, and from my experience admin-ing ubuntu, these changes shouldn't cause any issues there. It is ready for testing across all deploy types.

vv------------ original request ------------vv

Addressing issue #3736 

I've tested these changes on my own machine, OS=Arch Linux. The Debian and Docker builds need to be tested.

The PKGBUILDs for `authelia`, `authelia-bin`, and `authelia-git` will need two lines added to them, after the `install ... _pkgname.service` line. The leading source directory needs to be altered slightly from this example to match for the `authelia-bin` and `authelia-git` packages:

```
<authelia/PKGBUILD>
...
install -Dm644 "${srcdir}/${pkgname}-${pkgver}/${pkgname}.service" "${pkgdir}/usr/lib/systemd/system/${pkgname}.service"
+ install -Dm644 "${srcdir}/${pkgname}-${pkgver}/${pkgname}.sysusers.conf" "${pkgdir}/usr/lib/sysusers.d/${pkgname}.conf"
+ install -Dm644 "${srcdir}/${pkgname}-${pkgver}/${pkgname}.tmpfiles" "${pkgdir}/usr/lib/tmpfiles.d/${pkgname}.conf"
...
```

Similar changes will need to be made for the Debian builds.

The `authelia.sysusers.conf` file creates the `authelia` user (and implicitly the group), and the `authelia.tmpfiles` ensures the configuration directory is owned by this user, so the daemonized process can read the config.

The `sysusers.conf` file also adds the `authelia` user to the `redis` group.
- case a: the end user doesn't have redis installed
  - The `redis` group is implicitly created, and the authelia user is added to it, shouldn't cause any harm
  - If the end user installs redis after, that installation will explicitly create the `redis` group, and "sysusers.conf" will recognize it, and simply ensure the `authelia` user is still a member of that group
- case b: end user has redis installed, but wants to use `memory` config option
  - They can still use `memory` config option, shouldn't cause any harm also being part of the redis group
- case c: the user has redis and wants to use the `redis` config option
  - The authelia daemon user is already set as a member of the `redis` group, and all is ready for the end user to use this option.

We shouldn't need to alter any documentation for redis, since the `sysusers.conf` file handles it. However, if we want to inform the user that the authelia daemon user is made part of that group, I can add a sentence in the `redis.md` doc.

## Summary by CodeRabbit

- **New Features**
  - Enhanced security by configuring Authelia to run under a dedicated, less-privileged user and group with updated permission settings.
  - Updated service configurations with revised naming conventions and file management rules.
  - Improved packaging to include a comprehensive set of configuration and service files.
  - Added new configuration entries for managing system users and temporary files.

- **Chores**
  - Introduced an automated post-installation process to ensure proper system user setup and secure temporary file handling.


authelia.sysusers.conf
```diff
@@ -0,0 +1,3 @@
+#Type Name       ID      GECOS             Home directory Shell
+u     authelia   -       "authelia user"
```
james-d-elliott, line 2:
```suggestion
u     authelia   -       "Authelia Daemon User"
```

james-d-elliott (changes requested):
Probably a more descriptive gecos for this user. Also how does this affect users who upgrade? What about users with/without redis?

feinedsquirrel:
Whether the user does or does not have redis installed I believe is answered in the cases (a), (b), and (c) in the top comment. Let me know if they are unclear.
As for upgrade path, I'm thinking I need to alter the `tmpfiles.d` entry to include (or potentially switch the `e` line to) a `Z` line, that will recursively change the mode and ownership.
I also came up with another question. Are we assuming that the installers will always create `/etc/authelia/`? As a safety mechanism, I'm thinking I will expand the tmpfiles to include a `d` line to create the directory, retain the `e` line and add a `Z` line for updates.
I'll change it tonight when I'm back home, and once I push the update we'll need to test the different scenarios. I'm pretty sure the update path will be solved by adding the `Z` line.

authelia.sysusers.conf
```diff
@@ -0,0 +1,2 @@
+#Type Name       ID      GECOS                  Home directory Shell
+u     authelia   -       "Authelia Daemon User"
```
james-d-elliott, line 2:
```suggestion
u     authelia   -       "Authelia Daemon User"
m     authelia   redis
```

james-d-elliott (approved):

.buildkite/steps/debhelper.sh
```diff
@@ -17,7 +17,22 @@ sed -i -e '/^pkgname=/c pkgname=authelia' -e "/pkgver=/c $VERSION" -e '10,14d' -
 -e 's/sha256sums_aarch64.*/sha256sums_arm64=("SKIP")/' \
 -e 's/sha256sums_armv7h.*/sha256sums_armhf=("SKIP")/' \
 -e 's/x86_64/amd64/g' -e 's/aarch64/arm64/g' -e 's/armv7h/armhf/g' \
--e 's/CARCH/MAKEDEB_DPKG_ARCHITECTURE/g' PKGBUILD
+-e 's/CARCH/MAKEDEB_DPKG_ARCHITECTURE/g' \
+-e "s/provides=('authelia')/postinst='authelia.postinst'\nprovides=('authelia')/" PKGBUILD
```
james-d-elliott, line 21:
```suggestion
-e 's/CARCH/MAKEDEB_DPKG_ARCHITECTURE/g' PKGBUILD && \
sed -i '10i postinst="authelia.postinst"' PKGBUILD
```

james-d-elliott (changes requested):

.buildkite/steps/debhelper.sh
```diff
@@ -17,7 +17,22 @@ sed -i -e '/^pkgname=/c pkgname=authelia' -e "/pkgver=/c $VERSION" -e '10,14d' -
...
+# Trigger a reload of sysusers.d and tmpfiles.d to ensure the users and permissions are correct.
+systemd-sysusers
+systemd-tmpfiles --create
+
+AUTHELIAPLUSULTRA
+
+chmod +x authelia.postinst
 
 if [[ "${PACKAGE}" == "amd64" ]]; then
   docker run --rm -v $PWD:/build authelia/debpackager bash -c "cd /build && makedeb"
```
james-d-elliott, line 38:
```suggestion
  docker run --rm --platform linux/amd64 -v $PWD:/build authelia/debpackager bash -c "cd /build && makedeb"
```

james-d-elliott (dismissed):

PR #4996, opened 2023-02-26
refactor: add support to new docker compose cli on local setup.sh

Roriz:
## Description
There have been changes to the Docker Compose CLI interface. The new interface now uses Compose as a sub-command of Docker (docker compose -v) instead of a standalone command (docker-compose -v). See more in: https://docs.docker.com/compose/compose-v2/#differences-between-compose-v1-and-compose-v2

However, our local deployment setup script was only using the old version of docker-compose.

Additionally, we have added a warning for WSL users that they may need to manually make changes to the /etc/hosts file on Windows. Unfortunately, it seems that there is no way to automate this process within a script running inside of WSL. 🤔


james-d-elliott:
Thanks for contributing! 

We're aware of the compose v2 changes, and we use it with the current shell scripts. Several linux distros just alias `docker-compose` as `docker compose` to handle this.

Has this been tested in the following scenarios?

1. With `docker-compose` i.e. v1.
2. With `docker compose` i.e. v2.
3. With both `docker-compose` and `docker compose` i.e. both v1 and v2.
4. Without either.

Please also do one of the following:

- Add documentation for WSL2 in the contribution docs.
- Add note in the contribution docs regarding the current contribution docs work on WSL2.
- Remove the WSL specific information entirely.

examples/compose/local/setup.sh
```diff
@@ -27,7 +23,7 @@ if [[ ! -x "$(command -v docker)" ]]; then
   exit 1
 fi
 
-if [[ ! -x "$(command -v docker-compose)" ]]; then
+if [[ ! -x "$(command -v docker-compose)" && ! $(docker compose version) ]]; then
```
james-d-elliott, line 26: Factorize this with the other comment.

examples/compose/local/setup.sh
```diff
@@ -110,7 +111,7 @@ else
   password
 fi
 
-sudo docker-compose up -d
+sudo docker-compose up -d || sudo docker compose up -d
```
james-d-elliott, line 114: Instead of doing this, factorize it so that this is tested and the command is known (rather than using `||`).

For example:

```bash
COMPOSE=""

if [[ $(docker compose version) ]]; then
  COMPOSE="docker compose"
elif [[ -x "$(command -v docker-compose)"]]; then
  COMPOSE="docker-compose"
else
  echo "You must install Docker Compose on your machine";
  exit 1
fi
```

```bash
sudo ${COMPOSE} up -d
```

james-d-elliott (changes requested):
In addition to the above.

PR #5024, opened 2023-03-04
Create index.md

tetricky:
Integration notes for Misago forum software.


james-d-elliott (changes requested):
I'll commit these changes, just need you to check if they work when you get a chance. I removed the Origin config and made some minor adjustments. See https://640952f5205d160008157b9c--authelia-staging.netlify.app/integration/openid-connect/misago/

Also can you try (and provide an updated screenshot if it works, as well as a screenshot the same as the Basic settings screenshot with it enabled):

- Retrieving user data
- Access token location: `HTTP header (Bearer)`

tetricky:
> I'll commit these changes, just need you to check if they work when you get a chance. I removed the Origin config and made some minor adjustments. See https://640952f5205d160008157b9c--authelia-staging.netlify.app/integration/openid-connect/misago/
> 
> Also can you try (and provide an updated screenshot if it works, as well as a screenshot the same as the Basic settings screenshot with it enabled):
> 
>     * Retrieving user data
> 
>     * Access token location: `HTTP header (Bearer)`

I am, of course, happy with all the review changes. I can provide the screenshots requested, but I'm afraid I've not been able to work out how to provide these. I do not understand the code review process.

james-d-elliott:
What I'd like is if you can confirm this works (few changes to Authelia config, specifically fine tuning the client  and removing the Origin config since according to Misago's author CORS settings should not be needed): https://640952f5205d160008157b9c--authelia-staging.netlify.app/integration/openid-connect/misago/

Then I'd also like to confirm that  if changing the Retrieving user data section's Access token location from query string to HTTP header (Bearer) works. If that works I'd like a screenshot of that change.

tetricky:
> What I'd like is if you can confirm this works (few changes to Authelia config, specifically fine tuning the client and removing the Origin config since according to Misago's author CORS settings should not be needed): https://640952f5205d160008157b9c--authelia-staging.netlify.app/integration/openid-connect/misago/
> 
> Then I'd also like to confirm separately that if changing the Retrieving user data section's Access token location from query string to HTTP header (Bearer) works. If that works I'd like a screenshot of that change.

The revised authelia configuration works (with one caveat, which I will return to) in terms of how the authorization mechanism works to log into misago as a user.

The caveat is that the misago admin control panel is not part of the oauth 2 authentication scheme. In order to protect that from unauthenticated exposure, that needs authentication via a suitable reverse proxy. I use caddy, for this. For that mechanism to be available, then the grant types need to be:

```yml
        grant_types:
          - refresh_token
          - authorization_code
```

Yes it technically works from a strict authelia/misag log in perspective with just `authorization_code`, but this is not really entirely satisfactory.

>Then I'd also like to confirm separately that if changing the Retrieving user data section's Access token location from query string to HTTP header (Bearer) works. If that works I'd like a screenshot of that change.

That does not work. It only works using `Query String`. It is possible that the author has changed the code, or perhaps it was an initial misunderstanding/miscommunication.

james-d-elliott:
> That does not work. It only works using `Query String`. It is possible that the author has changed the code, or perhaps it was an initial misunderstanding/miscommunication.

Oh,, you'll need to adjust the method to POST as well I think. Authelia logs will reveal this however.

> The caveat is that the misago admin control panel is not part of the oauth 2 authentication scheme. In order to protect that from unauthenticated exposure, that needs authentication via a suitable reverse proxy. I use caddy, for this. For that mechanism to be available, then the grant types need to be:

The `refresh_token` grant type is ineffectual as the scopes requested do not ever issue a refresh token. Unless I'm missing something?

tetricky:
> The `refresh_token` grant type is ineffectual as the scopes requested do not ever issue a refresh token. Unless I'm missing something?

No. My mistake. I cleared all cookies and logged in clean, and you are correct. `refresh_token` is not required.

> > That does not work. It only works using `Query String`. It is possible that the author has changed the code, or perhaps it was an initial misunderstanding/miscommunication.
> 
> Oh, you'll need to adjust the method to POST as well I think. Authelia logs should reveal the issue however.

I changed the method to POST, but I can only get the flow to work for  `Query String`. I couldn't see anything in authelia logs, I'm pretty sure it's at the misago end.

james-d-elliott:
Okay neat. Only thing I'd like is a new screenshot to replace this with enabled selected, and the client secret set to `insecure_secret`: https://github.com/authelia/authelia/blob/e494a73dc036e122f6b848006c6516f7bdcc66ed/docs/content/en/integration/openid-connect/misago/misago-step-3-1.png

tetricky:
> Okay neat. Only thing I'd like then is a new screenshot to replace this with enabled selected, `Authelia` as the provider name rather than `authelia`, and the client secret set to `insecure_secret`: https://github.com/authelia/authelia/blob/e494a73dc036e122f6b848006c6516f7bdcc66ed/docs/content/en/integration/openid-connect/misago/misago-step-3-1.png

The revised screenshot is here:

https://misago-project.org/a/tqFWpt2PblhnYHpwNbuxpowTlqVi3N5BkBsEGXiKnYKYEjuhDwRnTwLs9FZDQ8c6/574/?shva=1

PR #5078, opened 2023-03-17
refactor(web): native vite env replacement

nightah:
#5073 introduced native env replacement within Vite therefore we can remove our html-transform hook.


james-d-elliott (approved):
LGTM

PR #5087, opened 2023-03-19
ci(buildkite): localised node artifacts

nightah:
This change ensures that vega Buildkite nodes will upload artifacts to a local MinIO instance and also retrieve artifacts from this location if the same constraint is met.


james-d-elliott (approved):
LGTM

PR #5094, opened 2023-03-20
docs: include pull request review checklist

james-d-elliott:


docs/content/en/contributing/guidelines/testing.md
```diff
@@ -0,0 +1,26 @@
...
+
+The following outlines the specific requirements we have for testing the Authelia code contributions.
+
+- While we aim for 100% coverage on changes and additions, we do not enforce this where it doesn't make practical sense:
+  - A test which just marks a line as tested is not necessarily an effectual test
+  - Sometimes there is limited ways in which tests can be performed and the limitation makes the test ineffectual
+- Tests should be named to reflect what they testing for and which part of the code they are testing
+- It's required for bug fixes that contributors create a test that fails prior to and passes
+  subsequent to fix being applied unless there is a specific reason to exclude this test, and this test is part of the
+  contribution
```
nightah, line 23:
```suggestion
- It's required for bug fixes that contributors create a test that fails prior to and passes
  subsequent to the fix being applied, this test must be included in the contribution,
  excluding this test will likely result in the fix being rejected unless explicitly agreed
  and advised otherwise by the core team
```

nightah (changes requested):
WYDT with regards to this suggestion, I'm open to change if you think it's necessary but we should convey the likely outcome and how the exception process may work.

PR #5170, opened 2023-04-04
[Docs] Added example prometheus job

boomam:
Added example Prometheus job to metrics/telemetry pages to aid with initial setup for new users.


james-d-elliott (approved):
LGTM

PR #5330, opened 2023-04-28
Additional screenshot, some detail improvements & wording updates

DennisGaida:
This is an addendum to https://github.com/authelia/authelia/pull/5287 and includes some improvements.

I don't understand how to insert line breaks just before the images, I now tried adding <kbd>Space</kbd><kbd>Space</kbd> at the end of the previous line, like you do in Markdown, but not sure this does the trick.


james-d-elliott:
While I think of it can you perform the [Developer Certificate of Origin Sign-Off](https://www.authelia.com/contributing/guidelines/commit-message/#developer-certificate-of-origin). We recently included this requirement in our contribution requirements.

It should also be the same as markdown, it should be possible to include an image similar to how it's done in markdown too but the theme we're using has a bug which I think will be fixed soon.

PR #5661, opened 2023-07-10
docs: update caddy subpath example, fix spelling issues

tzfx:
As referenced in discussion #5660 

Thought it would be helpful for users to have a heads-up that they need to update their `server.address` configuration in order to make the caddy subpath example work. Without it, the portal fails to render as it tries to pull javascript and css files from the wrong base path.

It didn't appear that any of the other proxy examples utilized subpaths, so I only did this for caddy. SWAG makes a passing mention in the caveat, however all the examples look to use subdomains.

I also noticed some minor spelling issues that were copypasta'd across the proxies markdown files.


docs/content/en/integration/proxies/caddy.md
```diff
@@ -158,6 +158,17 @@ nextcloud.example.com {
...
 
+*__Important:__ In order to use a subpath, you must also update your Authelia
+[server address configuration](../../configuration/miscellaneous/server.md#address) to listen on the new endpoint:*
+
+```yaml
+# configuration.yml
+server:
+  # address: "tcp://:9091"         # implicit default
+  address: "tcp://:9091/authelia"  # with subpath support.
+```
```
james-d-elliott, line 169:
```suggestion
```

docs/content/en/integration/proxies/caddy.md
```diff
@@ -158,6 +158,17 @@ nextcloud.example.com {
 
 #### Subpath
 
+*__Important:__ In order to use a subpath, you must also update your Authelia
+[server address configuration](../../configuration/miscellaneous/server.md#address) to listen on the new endpoint:*
```
james-d-elliott, line 162:
```suggestion
[server address configuration](../../configuration/miscellaneous/server.md#address) to listen on the new endpoint.*
```

james-d-elliott (changes requested):
The following changes will prevent us having to maintain additional areas of the docs unnecessarily. I'm also wondering why we don't have subpath config examples for other providers but do so for Caddy. Considering the idea of removing the subpath proxy configuration examples entirely.

docs/content/en/integration/proxies/caddy.md
tzfx: Updated latest commit with suggested changes.

tzfx:
>  Considering the idea of removing the subpath proxy configuration examples entirely.

My own selfish self really likes that a subpath config was included-- It worked perfectly (once I understood what was going on) for my use case in a way that didn't have me jumping through hoops with my dynamic dns provider on their free-tier. :)

Issue #3249, opened 2022-04-25
Ability to specify identity_providers.oidc.clients.[n].secret with environment variables.


alphabet5:
### Description

The ability to either substitute environment variables into the configuration, or overwrite with environment variables/files for client secrets.

```
Please Note: It is not possible to configure the access control rules section or OpenID Connect identity provider section using environment variables at this time.
```

### Use Case

I would like to be able to specify the client secret with a Kubernetes secret, to be shared between the client and authelia.

I'm not sure if something like `__0__` could be used for the index of the client list? 

It might be easier to allow {{ AUTHELIA_CUSTOM_ENV_NAME }} to be substituted into the config file?

james-d-elliott:
I've spent quite a lot of time trying to figure out a reasonable solution for this. Unfortunately most solutions have large pitfalls.

I would point out you can actually define the entire OIDC section as a secret in k8s, the method of doing so can be achieved using the following snippets:

```yaml
      volumes:
        - name: config-oidc
          secret:
            secretName: authelia-acl
            items:
              - key: configuration.oidc.yaml
                path: configuration.oidc.yaml
            defaultMode: 420
```

```yaml
      containers:
        - name: authelia
          image: docker.io/authelia/authelia:fix-oidc-pre-configured-consent
          command:
            - authelia
          args:
            - '--config=/configuration.yaml'
            - '--config=/configuration.oidc.yaml'
```

```yaml
          volumeMounts:
            - name: config-oidc
              readOnly: true
              mountPath: /configuration.oidc.yaml
              subPath: configuration.oidc.yaml
```

alphabet5:
Will the configuration of clients be removed when moving to SQL? 

Currently the use case I have for authelia is as a near-stateless ldap->oidc provider where adding a SQL dependency would not be ideal.

I wasn't able to dig up the source code, but this is how elastic is handling it with their beats configurations: https://www.elastic.co/guide/en/beats/filebeat/current/using-environ-vars.html

I don't think it's ideal, but I think it's probably the best mechanism I've seen.

Currently I'm using an init container with sed to replace environment variables within the config, and that's working well enough for now. 

```
command:
- 'sh'
- '-c'
[...]

PR #5716, opened 2023-07-28
add gettext to Dockerfile

davama:
simply need the `envsubst` to get the `AUTHELIA_IDENTITY_PROVIDERS_OIDC_CLIENTS_X_SECRET` environment variable to insert into the config.

Honestly didnt want to build my own image since it's just one package.

Hopefully this PR is ok

Thank you for the support!

**EDIT**

just to add, was looking to do something like [this](https://github.com/authelia/authelia/issues/3249) but looks like i need to render the config from env variables.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ ENV PATH="/app:${PATH}" \
     X_AUTHELIA_CONFIG="/config/configuration.yml"
 
 RUN \
-	apk --no-cache add ca-certificates su-exec tzdata
+	apk --no-cache add ca-certificates gettext su-exec tzdata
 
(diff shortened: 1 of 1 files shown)


james-d-elliott:
Check out the info on the filters here: https://authelia-next.netlify.app/configuration/methods/files/#file-filters

And the template functions available here (provided you use template instead of expand-env): https://authelia-next.netlify.app/reference/guides/templating/

davama:
Cool!
Didnt know that existed 
Will give it a shot

closing this Pr

Chuckame:
@james-d-elliott when would it be released ? We cannot see the documentation on the official website (templating not present in https://www.authelia.com/reference/guides)

james-d-elliott:
With 4.38.

PR #5853, opened 2023-08-16
Fixes an error caused by not providing an optional port value

paulcalabro:
This should fix an error message I was receiving when not providing an [optional port value](https://www.authelia.com/configuration/session/redis/#port):

```
---
session:
  domain: <SOME-DOMAIN>

  redis:
    host: <SOME-HOST>
```

```
authelia-application  | time="2023-08-15T21:43:24-07:00" level=error msg="Configuration: session: redis: option 'port' must be between 1 and 65535 but is configured as '0'"
authelia-application  | time="2023-08-15T21:43:24-07:00" level=fatal msg="Can't continue due to the errors loading the configuration"
```


internal/configuration/validator/session.go
```diff
@@ -218,7 +218,9 @@ func validateRedis(config *schema.SessionConfiguration, validator *schema.Struct
 
 	validateRedisCommon(config, validator)
 
-	if !path.IsAbs(config.Redis.Host) && (config.Redis.Port < 1 || config.Redis.Port > 65535) {
+	if config.Redis.Port == 0 {
+		config.Redis.Port = 26379
```
james-d-elliott, line 222: This should be the correct redis default port which is 6379, not 26379. Also this should utilize the schema.DefaultRedisConfiguration as the value should be set and used from there..

internal/configuration/validator/session_test.go
```diff
@@ -319,8 +319,9 @@ func TestShouldRaiseErrorWhenRedisHasHostnameButNoPort(t *testing.T) {
 	ValidateSession(&config, validator)
 
 	assert.False(t, validator.HasWarnings())
-	assert.Len(t, validator.Errors(), 1)
-	assert.EqualError(t, validator.Errors()[0], "session: redis: option 'port' must be between 1 and 65535 but it's configured as '0'")
```
james-d-elliott, line 323: This change likely reduces coverage. Please revert it and correct the test scenario / conditions so the test passes.

internal/configuration/validator/session_test.go
```diff
@@ -319,8 +319,9 @@ func TestShouldRaiseErrorWhenRedisHasHostnameButNoPort(t *testing.T) {
...
+	assert.False(t, validator.HasErrors())
+
+	assert.Equal(t, 26379, config.Redis.Port)
```
james-d-elliott, line 324: Please add a separate test for this (unless there is already one testing default values).

james-d-elliott (changes requested):

paulcalabro:
@james-d-elliott Does that look okay?

james-d-elliott (approved):
LGTM

PR #5876, opened 2023-08-19
refactor(validator): remove redundant nil check

Juneezee:
From the Go docs:

> "For a nil slice, the number of iterations is 0." https://go.dev/ref/spec#For_range

Therefore, an additional nil check for before the loop is unnecessary. Example: https://go.dev/play/p/2TugLncT_ob


james-d-elliott (approved):
LGTM

PR #5904, opened 2023-08-25
ci(web): replace vite-plugin-eslint with vite-plugin-checker

nightah:
This change replaces the existing ESLint plugin which also provides capability to perform additional checking, including TypeScript.


james-d-elliott (approved):
LGTM

PR #6199, opened 2023-10-27
Update supported-proxies.md

cmo-pomerium:
Adding Pomerium proxy to matrix list. The Pomerium team will come back to add a Pomerium guide to Authelia's documentation


james-d-elliott (changes requested):
We'll need to know which method of authorization handoff Promerium supports to merge this. I suspect the ExtAuthz handoff, but I couldn't find any docs on users being able to configure this with an external service. The only related docs that could be found were on configuring an external service to utilize promerium as an ExtAuthz service, not the other way around.

If this is not available it would be as compatible as IIS is. You could proxy authelia mostly fine, but would not be able to properly integrate it. We have not really added docs for this use case as it's basically documenting how to set specific headers which we already document here: https://www.authelia.com/integration/proxies/introduction/#required-headers

We'd however love to integrate with services like promerium in this way, or via OIDC.

PR #6261, opened 2023-11-11
docs: markdown fix on tailscale

aamirazad:
I know I only changed one character, but every bit counts, right? :sweat_smile:


james-d-elliott (approved):
LGTM and yes it does

PR #6269, opened 2023-11-13
Updated documentation for group filter in freeipa

NunoHiggs:
The current filter was unable to filter groups from freeipa. 

With `additional_groups_dn: cn=groups,cn=accounts` it can now filter correctly groups.
Tested on authelia v4.37.5 and Freeipa 4.9.10-9


james-d-elliott:
Is this the only base that a group can be created in with FreeIPA? Or can they be added to others? 

Maybe @davama @boomam or someone else would be able to provide input as well.

NunoHiggs:
The OU  for groups is not defined from my understanding like that in freeipa. It's a Organization Group but the definition is CN=blabla

NunoHiggs:
> > The OU for groups is not defined from my understanding like that in freeipa. It's a Organization Group but the definition is CN=blabla
> 
> This is they way we have it too.
> 
> But we do have an one ou structure for groups.
> 
> `ou=groups`
> 
> also, we use openldap, not freeipa.

Hi @davama 

My change is just for the documentation part of freeipa. Not any other ldap.

Thanks

boomam:
My only concern is whether the update to the docs reflects what FreeIPA has as their recommended filter format, compared to something that is (perhaps) situational to a specific deployment.

We know what is there worked as of the FreeIPA version when the doc was added/updated, and i wouldn't expect them to majorly change that kind of thing about their LDAP implementation between versions, as it would break a lot of installs.

I guess the question from me would be @NunoHiggs, are you trying to solve an issue you are having, or is this a change that is outlined by the latest FreeIPA docs?  
As if its the former i would argue that its probably not worth updating as it becomes situational, whereas if its the ladder, then lets update it.

NunoHiggs:
@boomam well tbh i have been working with IPA/IDM almost from since the start and i don't remember ever seeing it in the format OU=groups,dc=domain,dc=tld but always remember seeing it in cn=$groupname,cn=groups,cn=accounts,dc=domain,dc=tld

And i have built my entire auth ift around freeipa (3 headed cluster, with satellite auth nodes on several offsite locations), and it would hurt more to have to change everything  - auth, rbac, sudo, app/site ldap auth, CA, etc - to another provider since this is now working. 

I asked several colleagues that also use IPA/IDM extensively if they ever saw the group container described as OU=groups, or if ever it has been like that before.

boomam:
> I asked several colleagues that also use IPA/IDM extensively if they ever saw the group container described as OU=groups, or if ever it has been like that before, and they do not recall ever seeing it in OU=groups, even in compat


Very strange they've never heard of it, the 'OU=" format has been an LDAP/AD standard for decades....

NunoHiggs:
> ery strange they've never heard of it, the `OU=` format has been an LDAP/AD standard for decades....

Not on freeipa/redhat IDM. Again, this is just for documentation / how-to for freeipa, and not for all of ldap flavours

james-d-elliott:
Without something concrete we're probably just going to accept it. I'd rather have all of these things documented by means of an official implementation so will likely heavily rework this doc in the near future.

james-d-elliott (approved):

boomam:
> > ery strange they've never heard of it, the `OU=` format has been an LDAP/AD standard for decades....
> 
> Not on freeipa/redhat IDM. Again, this is just for documentation / how-to for freeipa, and not for all of ldap flavours. I was just trying to help with the apparent error on the page. If you see no value in that then let me know so i can discard this pull request.

Fair enough.  
As @james-d-elliott said, nothing wrong with extra content, its not like things are static and can never be updated. :-)

PR #6291, opened 2023-11-17
Update index.md for Synology DSM Configuration

scrapix:
Added the ldap section in authelias configuration.yml file.  It's important to have it configured as stated, otherwise ldap won't work with authelia

### DIFF
--- a/docs/content/en/integration/openid-connect/synology-dsm/index.md
+++ b/docs/content/en/integration/openid-connect/synology-dsm/index.md
@@ -87,6 +87,28 @@ identity_providers:
         - 'groups'
         - 'email'
       userinfo_signed_response_alg: 'none'
+
+  ldap:
+    implementation: custom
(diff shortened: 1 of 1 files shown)


james-d-elliott (changes requested):
Please remove the configuration itself and instead make a note about what specific element of the LDAP configuration is required when using Synology (likely that it's configured as the LDAP backend and that it has particular attributes).

I'd also be interested in creating a Synology implementation. Can you confirm there is no way to arrange synology users or groups in OU's? And can you create a test user and show all of their LDAP attributes under the following circumstances?

1. After creation.
2. After disabling the account.
3. After enabling the account but locking the account out by using a bad password.

james-d-elliott:
Closing this due to lack of activity, if you ever want to follow up we'd welcome it. Thanks for choosing to contribute.

PR #6369, opened 2023-12-04
fix email title width

Cajuteq:
Hi, on a large screen the Email Title is very small in width, I tracked this to be a simple change in a single file, so here is a small PR that should fix the title display on all width larger than 640.

Previous was :
image

I only tested the new css in a web view hence the font change : 
image


james-d-elliott (approved):
LGTM

PR #6391, opened 2023-12-07
Update traefik-ingress.md

boomam:
Cleaned up the examples and descriptions for creating middleware and ingresses, to make it more obvious what needs to be changed/what it is/etc.

Also moved the example from middleware to outside of the code block, as it was messing up the display of the original URL on the doc page.


docs/content/en/integration/kubernetes/traefik-ingress.md
```diff
@@ -75,6 +72,10 @@ spec:
 ...
 ```
 {{< /details >}}
+### Configuring URL in the proxy
+*We strongly suggest this is configured in the Session Cookies section of the Authelia configuration.*  
+Replace the address in the example middleware above with the following example syntax -  
+`address: 'http://authelia.default.svc.cluster.local/api/authz/forward-auth?authelia_url=https%3A%2F%2Fauth.example.com%2F'`
```
james-d-elliott, line 78:
```suggestion
{{< /details >}}
```

Sorry but we have zero support for not configuring the Authelia URL in the session cookies configuration. It is required for the new endpoints to operate. The query syntax is ONLY for overriding the value configured.

james-d-elliott (approved):
LGTM other than this small one which I'll commit.

PR #6404, opened 2023-12-11
build: explicitly use alpine 3.18

james-d-elliott:
Temporary fix for https://github.com/mattn/go-sqlite3/pull/1177 which will be reverted in the near future as that's merged.


nightah (approved):
LGTM.

PR #6480, opened 2023-12-30
systemd service: security hardening

stapelberg:
I used `systemd-analyze security authelia` and enabled a number of options that significantly reduce the attack surface. My instance of authelia still works with these options enabled, but comments welcome as to whether any of the settings might go a little too far…?

## Summary by CodeRabbit

- **Security Enhancements**
	- Enhanced security measures by restricting various capabilities, privileges, and access controls in service operations.


james-d-elliott:
Thanks for contributing. I think this should be okay. That being said I'm no systemd expert. I'll see if anyone else has any thoughts.

james-d-elliott:
@coderabbitai please review this PR.

nightah:
There's nothing harmful in the presented changes, I would suggest removing all the newlines and commentary for each option though.

Remember this file is only provided as a baseline because not all users would necessarily deploy Authelia via packaged mechanisms (some will just use the binary and the provided systemd unit).

authelia.service
```diff
@@ -7,5 +7,39 @@ Environment=AUTHELIA_SERVER_DISABLE_HEALTHCHECK=true
...
+ProtectKernelModules=yes
+
+# Make kernel variables (e.g. /proc/sys) read-only:
+ProtectKernelTunables=yes
+
+# Filter dangerous system calls. The following is listed as safe basic choice
+# in systemd.exec(5):
+SystemCallArchitectures=native
+SystemCallFilter=@system-service
+SystemCallErrorNumber=EPERM
```
james-d-elliott, line 42:
```suggestion
CapabilityBoundingSet=
NoNewPrivileges=yes
RestrictNamespaces=yes
ProtectHome=true
PrivateDevices=yes
PrivateUsers=yes
ProtectControlGroups=yes
ProtectKernelModules=yes
ProtectKernelTunables=yes
SystemCallArchitectures=native
SystemCallFilter=@system-service
SystemCallErrorNumber=EPERM
```

james-d-elliott (changes requested):

james-d-elliott (approved):

PR #6702, opened 2024-02-17
Grafana OIDC doc updates

liamgilbey:
As per the issue [here](https://github.com/grafana/grafana/issues/74154), as of Grafana v10, Grafana no longer supports email lookups by default.

To mitigate this when using Authelia, we need to include a few additional options in the Grafana configuration. This PR is to simply update the configuration to make these requirements clear.

## Summary by CodeRabbit

- **Documentation**
	- Updated the integration guide for OAuth in Grafana with new configuration options: `oauth_allow_insecure_email_lookup` and `skip_org_role_sync`.

### DIFF
--- a/docs/content/en/integration/openid-connect/grafana/index.md
+++ b/docs/content/en/integration/openid-connect/grafana/index.md
@@ -73,6 +73,8 @@ Add the following Generic OAuth configuration to the [Grafana] configuration:
 ```ini
 [server]
 root_url = https://grafana.example.com
+[auth]
+oauth_allow_insecure_email_lookup = true
 [auth.generic_oauth]
 enabled = true
 name = Authelia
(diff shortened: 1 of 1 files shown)


james-d-elliott:
I don't believe the original issue applies to us. We support and always provide the `sub` claim. If there is an actual issue we'd need to have a discussion about it first and this should not be a default configuration.

The change Grafana made here was to mitigate a CVE, so without more information this change is unlikely to be accepted. Theoretically they have a method now to link accounts using the sub claim?

nightah:
@liamgilbey I also believe you're trying to work around the error that suggests that the user already exists, if that's right I believe you can include the above keys in your Grafana config, and have the user login (which matches them again). Then you can just remove the key and it'll continue to work.

For users setting up Grafana as a new instance they won't run into the user already exists error therefore these keys shouldn't be required.

james-d-elliott:
Yep agreed. Would rather have a note about this in the guide rather than defaulting these potentially insecure options. New users should be fine.

liamgilbey:
Great points, thanks all

PR #6714, opened 2024-02-20
docs: Fix small typo example-two instead of example-three

DennisGaida:
Would make understanding things a bit more difficult. Small copy & paste error - the third example uses `example-three` obviously.


james-d-elliott:
Thanks, taken up locally in #6775

PR #6716, opened 2024-02-21
Feat: language picker

mind-ar:
this PR allows the end user to manually select the UI language
tasks:

- [x] Add Picker
- [x] Remove Duplicated Languages
- [x] Save user's choice in localStore
- [x] migrate locale list to api
- [x] generate locale name in correct language
- [ ] add tests


kacprzak80 (approved):

mind-ar:
imagen
imagen
imagen

mind-ar:
HI! this PR is almost ready for review, there's one problem: there are redundant locales that result in duplicated entries in the language picker, for example 
imagen
i can filter by locale (the two first letters) but that can be harmful for cases that is needed different variants of the same language

james-d-elliott:
I'll take a look at this once the conflicting changes are merged. I suspect we'll want something within the settings area too.

nightah:
> HI! this PR is almost ready for review, but there's one problem: there are redundant locales that result in duplicated entries in the language picker, for example imagen i can filter by locale (the two first letters) but that can be harmful for cases that is needed different variants of the same language

Why don't we group them with the top-level language code and include the language codes in the visual display, alongside the "clean name"?

cmd/authelia-gen/types.go
```diff
@@ -97,6 +97,7 @@ type Languages struct {
 	Defaults   DefaultsLanguages `json:"defaults"`
 	Namespaces []string          `json:"namespaces"`
 	Languages  []Language        `json:"languages"`
+	RealLng    []Language        `json:"realLng"`
```
james-d-elliott, line 100:
```suggestion
	RealLng    []Language        `json:"-"`
```

mind-ar:
> I'll take a look at this once the conflicting changes are merged. I suspect we'll want something within the settings area too. Looks mostly good however, would like to allow some additional things in general.

Added an option to the component to switch between two modes, so it can be used as part of a form
Select mode (`picker = false`)
imagen
Picker mode (`picker = true`)
imagen

nightah:
> > Why don't we group them with the top-level language code and include the language codes in the visual display, alongside the "clean name"?
> 
> @nightah something like this? imagen

I was thinking of something like a multi-level dropdown, so the top level would be the Language and base code and any variants that exist would sit under it, for example:
```
English (en)
-> en-AU
-> en-US
-> en-GB
```

If I were to use an example based on the actual generated list it would be something like this:
```
Portuguese (pt)
-> Portuguese (pt-BR)
```

mind-ar:
> I was thinking of something like a multi-level dropdown

i like the idea!
Just to clarify:
* the current dropdown  only shows the locales that have their own folder (ie: `pt`, `pt-BR`, `en`, `es-ES`, etc)
* with this scenario, if we want to show variants of one locale, it must be present as an folder (ie, we must have `en-AU`, `en-US`. `en-GB` folders).

i'm afraid if i add or rename some language folder (for example for unifying the two norwegian locales) i can break something in crowdin

@james-d-elliott do we still have crowdin? i did not found authelia project there.

james-d-elliott:
> @james-d-elliott do we still have crowdin? i did not found authelia project there.

Yeah we have a crowdin enterprise at https://translate.authelia.com

james-d-elliott:
First of all kudos for figuring out why the local detector was added and thanks for picking this up. I had some additional ideas which are not entirely necessary so if they don't seem easily doable I'd rather keep this PR clean and we can enhance later. Also if I didn't already mention we're going to make sure the major pending merge is merged before merging potentially problematic to resolve conflicts so we'll merge this after that if it ends up being ready after reevaluation. 

One element I"m really interested in and have been for a while is providing the languages available from the backend. This would theoretically allow people to add their own currently unsupported languages without us having to do an update.

This somewhat ties into this feature idea as we could also use `golang.org/x/text/language/display` to leverage `display.Self.Name(lang)` to translate the language into the most appropriate value automatically without needing translations and ensuring a user who can only speak français and has no idea what French is to easily swap.

mind-ar:
imagen

james-d-elliott:
Mostly exactly what I was thinking of. I was also thinking the list of available locales and fallbacks in the i18next config should be loaded from the backend instead of a static gen template. But thinking about it again maybe we'd have to be very careful about it especially when it comes to ETags. I'd love for a hot reload functionality eventually but we'd have to carefully design it.

mind-ar:
> I was also thinking the list of available locales and fallbacks in the i18next config should be loaded from the backend instead of a static gen template. But thinking about it again maybe we'd have to be very careful about it especially when it comes to ETags.

Just added a new handler for `/locales` that returns all the info needed. i had to move the logic from authelia_gen to `internal/utils` for reuse 
I didn't replaced the actual behavior of available locales list and fallbacks in i18next (still uses generated template), but i think we need to move towards to get this info from backend if we want to support custom locales
I need to investigate the Etag implementation

james-d-elliott:
@mind-ar can you rebase this and see if you want to make any additional adjustments.

mind-ar:
@james-d-elliott done!
I think that generating locale information on every request may have negative impact on performance. 
Should we load this into memory (or into a json file) on server startup?

james-d-elliott:
It should have a marginal impact. I think what we do here is handle it the same as the assets directory which I think is we just handle it at startup. The embedded assets can realistically be cached regardless and we just need to know what dirs exist in the asset locales directory after that.

james-d-elliott:
Also we can revisit this if we feel it needs to behave differently later, so whatever is fine by me for now.

james-d-elliott:
Also we can revisit this if we feel it needs to behave differently later, so whatever is fine by me for now. @mind-ar

mind-ar:
@james-d-elliott, there's an issue with Norwegian locales
  * there are two base names for this locale (`no` and `nb`) and they are identical (at least in our case)
  * this implies that there are two "Norwegian" in the language picker
  * one possible solution is to keep only one of these two locales, but i don't want to do this without your consent

For China and Taiwan, according with `language.Tag.Parent()` they does not share the base name, so they will be displayed as two different languages in the picker (the opposite case to Portuguese)

james-d-elliott:
Yea I ran into this issue too at some point, can't recall how I handled it. My recollection is there are actually 3 total. `no`, `nb`, and `nn-NO`. The first refers to either Norwegian locale, second is Bokmal, and third is Nynorsk. I'm surprised go doesn't have a field for this.. let me check in the debugger.

mind-ar:
> Yea I ran into this issue too at some point, can't recall how I handled it. My recollection is there are actually 3 total. `no`, `nb`, and `nn-NO`. The first refers to either Norwegian locale, second is Bokmal, and third is Nynorsk. I'm surprised go doesn't have a field for this.. let me check in the debugger.

after some research, i found that currently go/x/text/language uses CLDR v32.
imagen
Before v39, `nb` and `nn` were independent base locales and `no` was an alias of `nb`. on v39 they unified both `nb` and `nn` under `no`

As we have `no-NO` and `nb-NO`, and they are the same, my suggestion is to unify both locales into `nb` and, if a contributor wants to add Nynorsk, we add `nn`, until `go` updates to v39 of CLDR (i don't know if this breaks something in crowdin)

Here's some related information:
* https://cs.opensource.google/go/x/text/+/master:language/language.go;drc=71a9c9afc4cd710b9412f7f99f0d8e35b10e488a;l=140
* https://icu.unicode.org/design/norwegian-locales-changes-in-v39

james-d-elliott:
> The first refers to either Norwegian locale, second is Bokmal, and third is Nynorsk. I'm surprised go doesn't have a field for this.. let me check in the debugger.

Yeah this is what I meant here, I'm fine with whatever path we take here. We just have to ensure that the cases for browsers only advertising macrolanguages 
(`en`, `no`, `zh`, etc) are handled by one of the microlanguages, and that the relevant generate functions work with this too so we don't have to manually handle this each time languages are added.

mind-ar:
> We just have to ensure that the cases for browsers only advertising macrolanguages (`en`, `no`, `zh`, etc) are handled by one of the microlanguages, and that the relevant generate functions work with this too so we don't have to manually handle this each time languages are added (it will become a nightmare to manage pretty quickly).
Done

> Regarding caching, we can use the ETag / If-None-Match process to ensure the request is only processed when the values change. If we do this when/if we add logic to handle fs changes (i.e. locales added) then we can determine there is an update to be sent to clients. Though this can be added later too.
Done (etag)

Feel free to ask to do any change

mind-ar:
Hi, the last commits needs special attention:
* migrated supported languages and fallback information to backend
* dropped template generation for i18n index (`authelia-gen locales`)

This was necessary to allow support to custom locales that will next in a few days

Please, let me know if i must rollback these changes

internal/server/asset.go
```diff
@@ -248,3 +254,80 @@
...
+	}
+}
+
+func findLanguage(locale string, list []utils.Language) (*utils.Language, bool) {
+	for _, l := range list {
+		if l.Locale == locale {
+			return &l, true
+		}
+	}
+	return nil, false
```
autheliabot, line 332: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>return statements should not be cuddled if block has more than two lines (wsl)

mind-ar:
Hi! @james-d-elliott and @nightah  here's a functional version with custom locales support.
I think i messed it with custom locale support, i added this with the idea of allow using embedded and custom locales at the same time, but maybe it's better to allow only embedded or only custom

I'll let this as draft for now, I still need to do a general review/refactor and write some tests
Do not hesitate to request any modifications

web/src/utils/localStorage.ts
```diff
@@ -0,0 +1,28 @@
...
+export function localStorageAvailable() {
+    if (hasLocalStorageSupport !== null) return hasLocalStorageSupport;
+
+    if (typeof window !== "undefined" && window.localStorage !== null) {
+        hasLocalStorageSupport = true;
+
+        try {
+            window.localStorage.setItem(testKey, testValue);
+            window.localStorage.removeItem(testKey);
+        } catch (e) {
```
autheliabot, line 14: ⚠️ **[eslint]** <[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>'e' is defined but never used.

mind-ar:
Hi! removed custom locales from this PR. now this PR is only for language picker

Screenshot-2024-12-15 12-55-27
Screenshot-2024-12-15 12-55-37

Crowley723:
@coderabbitai What do you think about the sentence at web/src/components/AppBarItemLanguage.tsx Line 62?

internal/server/const.go
```diff
@@ -85,6 +85,10 @@
 	tmplCSPSwagger = "default-src 'self'; img-src 'self' https://validator.swagger.io data:; object-src 'none'; script-src 'self' 'nonce-%s'; style-src 'self' 'sha256-RL3ie0nH+Lzz2YNqQN83mnU0J1ot4QL7b99vMdIX99w='; base-uri 'self'"
 )
 
+const (
+	errCantLoadLocaleInfo = "fatal: can't load Locale Information"
```
autheliabot, line 89: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>const `errCantLoadLocaleInfo` is unused (unused)

internal/server/handlers.go
```diff
@@ -124,6 +132,9 @@
 
 	handlerPublicHTML := newPublicHTMLEmbeddedHandler()
 	handlerLocales := newLocalesEmbeddedHandler()
+	if handlerLocalesList, err = newLocalesListHandler(); err != nil {
```
autheliabot, line 135: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>only one cuddle assignment allowed before if statement (wsl)

internal/server/asset.go
```diff
@@ -80,58 +82,62 @@
...
-		language, namespace = ctx.UserValue("language").(string), ctx.UserValue("namespace").(string)
+		_, ok := aliases[v.Parent]
 
-		if !utils.IsStringInSlice(language, languages) {
-			return false, ""
+		if !ok {
+			aliases[v.Parent] = v.Locale
 		}
+	}
+	aliases["no"] = "nb"
```
autheliabot, line 135: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>assignments should only be cuddled with other assignments (wsl)

nightah:
@mind-ar and @james-d-elliott can we write some integration tests into the Standalone suite for this before 4.39 goes live?

james-d-elliott:
Odd, actually wrote one but never pushed it. I couldn't find much that would be reliable to test other than that a known commodity element changed when you picked a language. I'll push it in a few.

nightah:
> Odd, actually wrote one but never pushed it. I couldn't find much that would be reliable to test other than that a known commodity element changed when you picked a language. I'll push it in a few.

Yeah I think that suffices and even seeing that the language inside the language picker itself has changed would also be a good indication alongside it.

PR #6960, opened 2024-03-19
fix(logging): include logs about logging config

james-d-elliott:
Similar to fdb62daf9ea895e4e719c15dcee640e1f0e16e10 / #6935, it makes it hard to debug some configurations when various log configurations exist so logging the configuration at startup as a debug log message will assist the developers in debugging this kind of setup. This fixes the absence of this log message.

## Summary by CodeRabbit

- **Refactor**
	- Updated logging initialization to use a new configuration schema and added debug logging for settings initialization.


nightah (approved):
LGTM.

PR #6968, opened 2024-03-20
fix(configuration): missing trailing slash

james-d-elliott:
The trailing slash in the Authelia URL when missing in the path can cause situations where a user can provide an invalid configuration and not be aware which can be rather difficult to diagnose. This fixes that issue by ensuring if there is a path that it ends with a '/'.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved validation for session cookie URLs to ensure they are correctly formatted, enhancing the reliability of session management.
	- Updated error messages to include more accurate information when invalid portal URLs are detected, aiding in troubleshooting configuration issues.
	- Ignored `internal/suites/` directory in the `.yamllint.yml` configuration file.


nightah (approved):
LGTM.

PR #7105, opened 2024-04-07
update documentation about nextcloud integration

jvies:
hi,

Here is a PR about some update for the OpenID integration with Nextcloud.

I did not removed the information about the first app, I chose to keep it as it is probably still valuable.

If you prefer that I create another page.  I can do it.

Thanks a lot


james-d-elliott:
Please add the new example below the existing example as it's more thoroughly tested.

james-d-elliott (approved):

PR #7130, opened 2024-04-10
fix(web): adjust router basename behaviour

nightah:
This change adjusts the behaviour if Authelia is run with a sub-path and is visited without a trailing slash on the specified sub-path.
In 4.37.5 the base path would get normalized without a trailing slash, however, would cause issues when a refresh was completed while carrying a redirection query string.
In 4.38.x this was changed so the sub-path would not be normalized without the trailing slash and that it was therefore necessary. This change in behaviour would be observed as a regression by users with learned behaviours.

## Summary by CodeRabbit

- **New Features**
  - Enhanced URL handling in the application to improve user navigation and system logging.
  - Improved the Docker Compose setup for backend development, including refined volume mapping and reduced script sleep times.

- **Bug Fixes**
  - Corrected method handling for requests that are not allowed, ensuring proper redirection and error handling.

- **Refactor**
  - Optimized the authentication middleware and URL processing logic to streamline user interactions and system performance.

- **Chores**
  - Removed unnecessary delays in development scripts to speed up the backend development process.

- **Tests**
  - Added new tests to verify correct rendering of the frontend with and without trailing slashes.

### DIFF
--- a/internal/logging/const.go
+++ b/internal/logging/const.go
@@ -46,6 +46,7 @@ const (
 	FieldRemoteIP   = "remote_ip"
 	FieldMethod     = "method"
 	FieldPath       = "path"
+	FieldPathRaw    = "path_raw"
 	FieldStatusCode = "status_code"
 )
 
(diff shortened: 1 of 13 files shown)


james-d-elliott (approved):
LGTM

PR #7140, opened 2024-04-11
feat(logging): reopen on sighup

### LINKED ISSUE #4964: A way to tell authelia to reopen its logfile
### Description

In https://github.com/authelia/authelia/issues/3210#issuecomment-1101917497 I can see that is is recommended to use logrotate to rotate authelia's own logfile, which is fair enough.

But, unfortunately there is no way currently (not that I know of) to tell authelia to reopen its logfile after it has been rotated.
That means only the `copytruncate` option can be used instead of  `create` in logrotate, which has the downside of possibly losing some logs.

So I'd like to see an option to tell authelia to reopen its logfile to be able to use the `create` option in logrotate. (either by handling -HUP/-USR1/-USR2 signals, or any other method)

### Use Case

When using authelia's own mechanism to log to a file, that file grows forever.
It is recommended to use linux's logrotate tool to rotate it, but needs a bit more support from authelia's side.

### Details

_No response_

### Documentation

https://linux.die.net/man/8/logrotate

### Pre-Submission Checklist

james-d-elliott:
This implements a method to send Authelia a SIGHUP signal to indicate it should reload the configuration file. This can be used in two ways. If using an external logrotate service you can tell Authelia to reopen the file and create it if it doesn't exist. Secondly if you use the existing time date replacements this will create a brand new log file with the current time.

Closes #4964

### DIFF
--- a/docs/content/configuration/miscellaneous/logging.md
+++ b/docs/content/configuration/miscellaneous/logging.md
@@ -86,6 +86,10 @@ default uses the [RFC3339] layout, but optionally can be suffixed with the
 [Go Layout](https://pkg.go.dev/time#pkg-constants) semantics in the format of `{datetime:<layout>}` where `<layout>` is
 the layout supported by Go.
 
+When using a log file sending the Authelia process a SIGHUP will cause it to close and reopen the current log file and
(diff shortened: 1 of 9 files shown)


Crowley723 (dismissed):
#8577 
:+1:

PR #7251, opened 2024-04-28
feat(web):  introduction of an access denied page

### LINKED ISSUE #2319: Create a deny page
### Description

I would like that Authelia display a beautiful "Access denied" page when a user has no permission to access a resource, as defined by the rules.

### Use Case

Currently we configured HAProxy to return the default 403 page. We could create a custom 403 page, but it would make sense that this page use the same design as Authelia.

rocketproto:
When a user does not have the required permissions, a redirection url to a new 'Access Denied' page is passed back to the reverse proxy along with the 403 error code. This follows the same sequence as an unauthenticated user getting redirected to the login. The Access Denied page presents the user with an option to logout, as well as the details of the page they were denied access to.

Work includes a slight refactor of authz handlers to allow for code reuse.

Also includes a simple 'demo home' button injected into the development login page to allow for easier navigation back to home.example.com for the Standalone suite.

Fixes #2319.

### DIFF
--- a/internal/handlers/const.go
+++ b/internal/handlers/const.go
@@ -45,6 +45,7 @@ const (
 	queryArgConsentID  = "consent_id"
 	queryArgWorkflow   = "workflow"
 	queryArgWorkflowID = "workflow_id"
+	queryArgEC         = "ec"
 )
 
 var (
@@ -52,6 +53,14 @@ var (
 	qryArgRD        = []byte(queryArgRD)
 	qryArgAuth      = []byte(queryArgAuth)
 	qryArgConsentID = []byte(queryArgConsentID)
+)
+
+var (
+	baseErrorPath = "error"
+)
+
+var (
+	errorForbidden = "forbidden"
 )
 
 var (
@@ -91,7 +100,7 @@ const (
 )
 
 const (
-	logFmtAuthzRedirect = "Access to %s (method %s) is not authorized to user %s, responding with status code %d with location redirect to %s"
+	logFmtAuthzRedirect = "Access to %s (method %s) is not granted to user %s, responding with status code %d with location redirect to %s"
 
 	logFmtAuthorizationPrefix = "Authorization Request with id '%s' on client with id '%s' "
 
--- a/internal/handlers/handler_authz.go
+++ b/internal/handlers/handler_authz.go
@@ -93,8 +93,7 @@ func (authz *Authz) Handler(ctx *middlewares.AutheliaCtx) {
 
 	switch isAuthzResult(authn.Level, required, ruleHasSubject) {
 	case AuthzResultForbidden:
-		ctx.Logger.Infof("Access to '%s' is forbidden to user '%s'", object.URL.String(), authn.Username)
-		ctx.ReplyForbidden()
+		authz.handleForbidden(ctx, authn, authz.getErrorRedirectionURL(&object, autheliaURL, errorForbidden))
 	case AuthzResultUnauthorized:
 		var handler HandlerAuthzUnauthorized
 
@@ -134,12 +133,12 @@ func (authz *Authz) getAutheliaURL(ctx *middlewares.AutheliaCtx, provider *sessi
 	return nil, fmt.Errorf("authelia url lookup failed")
 }
 
-func (authz *Authz) getRedirectionURL(object *authorization.Object, autheliaURL *url.URL) (redirectionURL *url.URL) {
-	if autheliaURL == nil {
+func (authz *Authz) getRedirectionURL(object *authorization.Object, baseURL *url.URL) (redirectionURL *url.URL) {
+	if baseURL == nil {
 		return nil
 	}
 
-	redirectionURL, _ = url.ParseRequestURI(autheliaURL.String())
+	redirectionURL, _ = url.ParseRequestURI(baseURL.String())
 
 	if redirectionURL.Path == "" {
 		redirectionURL.Path = "/"
@@ -156,6 +155,34 @@ func (authz *Authz) getRedirectionURL(object *authorization.Object, autheliaURL 
 	redirectionURL.RawQuery = qry.Encode()
 
 	return redirectionURL
+}
+
+func (authz *Authz) getErrorRedirectionURL(object *authorization.Object, baseURL *url.URL, errorString string) (errorRedirectionURL *url.URL) {
+	if baseURL == nil {
+		return nil
+	}
+
+	baseErrorURL, _ := url.ParseRequestURI(baseURL.String())
+
+	if baseErrorURL.Path == "" {
+		baseErrorURL.Path = "/"
+	}
+
+	baseErrorURL.Path += baseErrorPath
+
+	qry := baseErrorURL.Query()
+
+	if object.Method != "" {
+		qry.Set(queryArgRM, object.Method)
+	}
+
+	qry.Set(queryArgEC, errorString)
+
+	baseErrorURL.RawQuery = qry.Encode()
+
+	errorRedirectionURL = authz.getRedirectionURL(object, baseErrorURL)
+
+	return errorRedirectionURL
 }
 
 func (authz *Authz) authn(ctx *middlewares.AutheliaCtx, provider *session.Session, object *authorization.Object) (authn *Authn, strategy AuthnStrategy, err error) {
--- a/internal/handlers/handler_authz_common.go
+++ b/internal/handlers/handler_authz_common.go
@@ -112,3 +112,27 @@ func hasInvalidMethodCharacters(v []byte) bool {
 
 	return false
 }
+
+func isNotRequestingWebpage(ctx *middlewares.AutheliaCtx) bool {
+	return ctx.IsXHR() || !ctx.AcceptsMIME("text/html")
+}
+
+func deriveStatusCodeFromAuthnMethod(authn *Authn) (statusCode int) {
+	switch authn.Object.Method {
+	case fasthttp.MethodGet, fasthttp.MethodOptions, fasthttp.MethodHead:
+		return fasthttp.StatusFound
+	default:
+		return fasthttp.StatusSeeOther
+	}
+}
+
+func handleAuthzSpecialRedirect(ctx *middlewares.AutheliaCtx, authn *Authn, redirectionURL *url.URL, statusCode int) {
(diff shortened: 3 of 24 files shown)


internal/handlers/handler_authz_impl_legacy.go
```diff
@@ -38,33 +38,54 @@
...
+
+func handleAuthzForbiddenLegacy(ctx *middlewares.AutheliaCtx, authn *Authn, redirectionURL *url.URL) {
+	var (
+		statusCode int
+	)
+
+	if authn.Type == AuthnTypeAuthorization {
+		ctx.Logger.Infof("Access to %s (method %s) is forbidden for user %s, responding with status code %d", authn.Object.URL.String(), authn.Method, authn.Username, fasthttp.StatusForbidden)
+		ctx.ReplyForbidden()
+		return
```
autheliabot, line 71: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>return statements should not be cuddled if block has more than two lines (wsl)

rocketproto:
For quick view: 

The access denied page: 
image

Noting the "Demo Home" link is only injected into the page by the vite config while being served in the demo container. It will not be in the production build.

Henkow:
Great work, just a question. If the user who got denied haven't setup TOTP and the denied resource mandates TOTP, will the user then have the ability to request a TOTP token? Because that is the case with the current implementation if you go back to the same URL  after the 403 and apply /authelia. 

I think the current implementation is incorrect. I have a user specifically for one_factor proxies, I don't want that user to have the ability to request a TOTP token if the user in question is not permitted to the resource to begin with.

james-d-elliott:
Thanks for this, we'll consider it for 4.39 if it's adjusted to be a non-breaking change.

rocketproto:
> Thanks for this, we'll consider it for 4.39 if it's adjusted to be a non-breaking change.

Hey, 

Thanks for considering, although I'd say calling the defect a breaking change is being very polite

The amount of free time I have to work on things has drastically reduced. If someone's able to offer some high level technical overview on the application flow at a higher level than working on authelia context, but lower level than the docs

That or if someone in the future was to use this as a starting point to build off, I'd also be happy :)

internal/handlers/handler_authz_impl_legacy.go
```diff
@@ -38,33 +38,55 @@ func handleAuthzUnauthorizedLegacy(ctx *middlewares.AutheliaCtx, authn *Authn, r
...
-			ctx.SpecialRedirect(redirectionURL.String(), statusCode)
-		}
+		handleAuthzSpecialRedirect(ctx, authn, redirectionURL, statusCode)
 	} else {
 		ctx.Logger.Infof("Access to %s (method %s) is not authorized to user %s, responding with status code %d", authn.Object.URL.String(), authn.Method, authn.Username, statusCode)
 		ctx.ReplyUnauthorized()
 	}
 }
+
+func handleAuthzForbiddenLegacy(ctx *middlewares.AutheliaCtx, authn *Authn, redirectionURL *url.URL) {
```
james-d-elliott, line 63: Why is this necessary?

internal/suites/example/compose/nginx/portal/nginx.conf
```diff
@@ -310,7 +310,7 @@ http {
             auth_request_set $redirection_url $upstream_http_location;
 
             ## Modern Method: When there is a 401 response code from the Authz endpoint redirect to the $redirection_url.
-            error_page 401 =302 $redirection_url;
+            error_page 401 403 =302 $redirection_url;
```
james-d-elliott: This implementation must not change for this to be accepted.

internal/handlers/const.go
```diff
@@ -99,7 +100,7 @@ const (
 )
 
 const (
-	logFmtAuthzRedirect = "Access to %s (method %s) is not authorized to user %s, responding with status code %d with location redirect to %s"
+	logFmtAuthzRedirect = "Access to %s (method %s) is not granted to user %s, responding with status code %d with location redirect to %s"
```
james-d-elliott, line 103:
```suggestion
	logFmtAuthzRedirect = "Access to %s (method %s) is not authorized to user %s, responding with status code %d with location redirect to %s"
```

internal/handlers/handler_authz_impl_legacy.go
```diff
@@ -38,33 +38,55 @@ func handleAuthzUnauthorizedLegacy(ctx *middlewares.AutheliaCtx, authn *Authn, r
 
 	if authn.Type == AuthnTypeAuthorization {
 		handleAuthzUnauthorizedAuthorizationBasic(ctx, authn)
-
 		return
 	}
 
 	switch {
-	case ctx.IsXHR() || !ctx.AcceptsMIME("text/html") || redirectionURL == nil:
+	case isNotRequestingWebpage(ctx) || redirectionURL == nil:
```
james-d-elliott, line 45:
```suggestion
	case ctx.IsXHR() || !ctx.AcceptsMIME("text/html") || redirectionURL == nil:
```

internal/suites/example/compose/nginx/portal/nginx.conf
james-d-elliott, line 313:
```suggestion
            error_page 401 =302 $redirection_url;
```

internal/suites/example/compose/nginx/portal/nginx.conf
```diff
@@ -212,7 +212,7 @@ http {
...
```
james-d-elliott, line 215:
```suggestion
            error_page 401 =302 $redirection_url;
```

internal/suites/example/compose/nginx/portal/nginx.conf
```diff
@@ -386,7 +386,7 @@ http {
...
```
james-d-elliott, line 389:
```suggestion
            error_page 401 =302 $redirection_url;
```

internal/handlers/handler_authz_common.go
```diff
@@ -112,3 +112,27 @@ func hasInvalidMethodCharacters(v []byte) bool {
 
 	return false
 }
+
+func isNotRequestingWebpage(ctx *middlewares.AutheliaCtx) bool {
+	return ctx.IsXHR() || !ctx.AcceptsMIME("text/html")
+}
+
```
james-d-elliott, line 119:
```suggestion

```

These kinds of refactors should be done outside of a change like this to make it easier to review.

internal/handlers/handler_authz.go
```diff
@@ -133,12 +133,12 @@ func (authz *Authz) getAutheliaURL(ctx *middlewares.AutheliaCtx, provider *sessi
 	return nil, fmt.Errorf("authelia url lookup failed")
 }
 
-func (authz *Authz) getRedirectionURL(object *authorization.Object, autheliaURL *url.URL) (redirectionURL *url.URL) {
-	if autheliaURL == nil {
+func (authz *Authz) getRedirectionURL(object *authorization.Object, baseURL *url.URL) (redirectionURL *url.URL) {
+	if baseURL == nil {
 		return nil
 	}
```
james-d-elliott, line 139:
```suggestion
func (authz *Authz) getRedirectionURL(object *authorization.Object, autheliaURL *url.URL) (redirectionURL *url.URL) {
	if autheliaURL == nil {
return nil
}
```

internal/handlers/handler_authz.go
```diff
@@ -133,12 +133,12 @@ func (authz *Authz) getAutheliaURL(ctx *middlewares.AutheliaCtx, provider *sessi
...
 
-	redirectionURL, _ = url.ParseRequestURI(autheliaURL.String())
+	redirectionURL, _ = url.ParseRequestURI(baseURL.String())
```
james-d-elliott, line 141:
```suggestion
	redirectionURL, _ = url.ParseRequestURI(autheliaURL.String())
```

internal/handlers/handler_authz.go
```diff
@@ -157,6 +157,34 @@ func (authz *Authz) getRedirectionURL(object *authorization.Object, autheliaURL
...
+	}
+
+	qry.Set(queryArgEC, errorString)
+
+	baseErrorURL.RawQuery = qry.Encode()
+
+	errorRedirectionURL = authz.getRedirectionURL(object, baseErrorURL)
+
+	return errorRedirectionURL
+}
```
james-d-elliott, line 186:
```suggestion
func (authz *Authz) getErrorRedirectionURL(object *authorization.Object, autheliaURL *url.URL, errorString string) (errorRedirectionURL *url.URL) {
	if autheliaURL == nil {
		return nil
	}

	autheliaErrorURL, _ := url.ParseRequestURI(autheliaURL.String())

	if autheliaErrorURL.Path == "" {
		autheliaErrorURL.Path = "/"
	}

	autheliaErrorURL.Path += autheliaErrorURL

	qry := autheliaErrorURL.Query()

	if object.Method != "" {
		qry.Set(queryArgRM, object.Method)
	}

	qry.Set(queryArgEC, errorString)

	autheliaErrorURL.RawQuery = qry.Encode()

	return authz.getRedirectionURL(object, autheliaErrorURL)
}
```

internal/handlers/handler_authz_impl_authrequest.go
```diff
@@ -37,12 +37,9 @@ func handleAuthzGetObjectAuthRequest(ctx *middlewares.AutheliaCtx) (object autho
...
 
-	switch authn.Object.Method {
-	case fasthttp.MethodHead:
-		ctx.SpecialRedirectNoBody(redirectionURL.String(), fasthttp.StatusUnauthorized)
-	default:
-		ctx.SpecialRedirect(redirectionURL.String(), fasthttp.StatusUnauthorized)
-	}
+func handleAuthzForbiddenAuthRequest(ctx *middlewares.AutheliaCtx, authn *Authn, redirectionURL *url.URL) {
+	handleAuthzSpecialRedirect(ctx, authn, redirectionURL, fasthttp.StatusForbidden)
 }
```
james-d-elliott, line 45: The changes here constitute a breaking change, in conjunction with the nginx suite changes.

james-d-elliott:
@rocketproto apologies for not replying, I could not quite understand what you were saying. The comment is not specifically about the TOTP comment, but the behavior of those API endpoints changed arbitrarily for everyone, ergo a breaking change.

At the time of the PR this would break every nginx user and potentially others.

PR #7650, opened 2024-08-08
refactor web: change 2fa method created_at last_used time to time-since-date from formatted date

Crowley723:
The existing date format for creation time/last used time for 2fa methods in the user settings menu is a bit lengthy and doesn't render too well even on desktop devices. This changes it to a delta-time format. Instead of showing the date, it shows the time since the last event e.g. just now, one day ago, 2 years ago etc.

Creation time/last used time doesn't need to be very specific, having the exact date and time when a credential was created/used isn't a super big issue.
For webauthn credentials, users still have the ability to see the exact date and time by inspecting the credential.

I think this should increase the readability of the settings page while also cutting down on the clutter, while maintaining clarity on when credentials were created/used.

### DIFF
--- a/web/src/views/Settings/TwoFactorAuthentication/CredentialItem.tsx
+++ b/web/src/views/Settings/TwoFactorAuthentication/CredentialItem.tsx
@@ -7,7 +7,6 @@ import Grid from "@mui/material/Unstable_Grid2/Grid2";
 import { useTranslation } from "react-i18next";
 
 import { useRelativeTime } from "@hooks/RelativeTimeString";
-// import { FormatDateHumanReadable } from "@i18n/formats";
 
 interface Props {
     id: string;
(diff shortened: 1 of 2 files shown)


Crowley723:
Before: 

After:

web/src/views/Settings/TwoFactorAuthentication/CredentialItem.tsx
```diff
@@ -6,7 +6,8 @@ import IconButton from "@mui/material/IconButton";
 import Grid from "@mui/material/Unstable_Grid2/Grid2";
 import { useTranslation } from "react-i18next";
 
-import { FormatDateHumanReadable } from "@i18n/formats";
+import { useRelativeTime } from "@hooks/RelativeTimeString";
+// import { FormatDateHumanReadable } from "@i18n/formats";
```
james-d-elliott, line 10:
```suggestion
// import { FormatDateHumanReadable } from "@i18n/formats";
```

james-d-elliott (changes requested):
I really love this change overall. There is one modification I've suggested.

Also we need to look at translations as I believe this will always be in English.

web/src/views/Settings/TwoFactorAuthentication/CredentialItem.tsx
Crowley723: Used [Intl.RelativeTimeFormat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) to properly translate the time strings.

It should be noted that without including `i18next.languages` as the list of default locales for `Intl.RelativeTimeFormat()`, RelativeTimeFormat will use the browser's language settings and ignore any language preferences set/used by react-i18next. Which can lead to the issue where everything is in the correct language except the string returned by RelativeTimeFormat.

james-d-elliott (approved):

james-d-elliott:
@all-contributors please add @Crowley723  for docs, design, and fixes

Issue #7467, opened 2024-06-30
Add --quiet option to `authelia crypto rand` command


jonasgeiler:
### Description

When running `authelia crypto rand --length 64`, the following is printed to the terminal:
```
Random Value: 2RUT9NzHzpXUp1fMDC2QaNnLk5ddYfoFwqjofwKhF0YFtaFnhu09eHkPzU6KsDHC
```

If I wanted to pipe this output into a file like `authelia crypto rand --length 64 > /data/secrets/jwt_secret`, the file will contain something like this:
```
Random Value: jn81Q2TardAO4EPcgG4mfAmvjtIDE8SYaZ3QgI9LjA8ZGSwMeLIMC61d08wKs4vn
```
As you can see the `Random Value: ` part is also written to the file.

It would be nice if I could provide a `--quiet` option like `authelia crypto rand --length 64 --quiet > /data/secrets/jwt_secret`, so the file will only contain this:
```
jn81Q2TardAO4EPcgG4mfAmvjtIDE8SYaZ3QgI9LjA8ZGSwMeLIMC61d08wKs4vn
```

Obviously you can remove the `Random Value: ` part yourself with a little bit of Bash scripting, but it would still be nice to have this option for simplicity and compatibility with any changes to the output in the future.

### Use Case

The main use case is using the `authelia crypto rand` command in Bash scripts and the-like. You could put the output into a Bash variable or pipe it into a file for use with Docker secrets and such. Or maybe someone wants just the random string without other stuff for easier copy-pasting?

### Details

The `authelia crypto rand` command should have a `--quiet` option which removes and redundant output from the command. Only the actual random string should be outputted.

### Pre-Submission Checklist

- [X] I agree to follow the [Code of Conduct](http://www.authelia.com/code-of-conduct)

- [X] I have checked for related issues and checked the documentation

jonasgeiler:
> However, you can achieve the desired result using standard shell tools to filter the output. For example, you can use `awk` to extract just the random string:
> 
> ```shell
> authelia crypto rand --length 64 | awk '{print $3}' > /data/secrets/jwt_secret
> ```

Of course this would be possible, but as I said the output might change in the future and could break your script, for example if the output changes from `Random Value: abc` to `Random Alphanumeric Value: abc` or something like that.

jonasgeiler:
I'd be ready to work on a PR myself to add this feature. Let me know if I can!

james-d-elliott:
[...]

PR #7683, opened 2024-08-15
refactor(commands): Add new crypto rand command options

Crowley723:
- Added ability to suppress string prefixes for cleaner output and easier file creation
- Implemented option to specify an output file for generated random strings

Add -q, --quiet to remove "Random String: " prefix
Add -o, --output-file to specify an output file.

It should be noted that because this command is most often run using the docker exec command that beyond redirecting output from stdout to file using '>' one must use a volume mount to write the output file outside of the container.
That looks like:
`docker run -u "$(id -u):$(id -g)" -v "$(pwd)":/keys authelia/authelia:latest authelia crypto rand --output-file /keys/random.txt`
to place the `random.txt` file in the current directory.

Resolves: [7467](https://github.com/authelia/authelia/issues/7467)


internal/commands/crypto.go
```diff
@@ -246,17 +250,45 @@ func newCryptoGenerateCmd(ctx *CmdCtx, category, algorithm string) (cmd *cobra.C
...
-	if value := url.QueryEscape(random); random != value {
-		fmt.Printf("Random Value (URL Encoded): %s\n", value)
+	if outputPath, err = flagsGetFilePath(cmd.Flags(), cmdFlagNameOutputFile, cmdFlagNameDirectory); err != nil {
+		return fmt.Errorf("error getting output file path: %v", err)
+	}
+
+	if outputPath != "" {
+		dir := filepath.Dir(outputPath)
+
+		if err = os.MkdirAll(dir, 0755); err != nil {
```
jonasgeiler, line 271: Not sure if creating all parent directories is an expected or common behavior in CLI programs. I think we should rather error when the output file parent directory doesn't exist by just using `os.WriteFile` and nothing else.

internal/commands/crypto.go
```diff
@@ -246,17 +250,45 @@ func newCryptoGenerateCmd(ctx *CmdCtx, category, algorithm string) (cmd *cobra.C
...
 	}
 
-	fmt.Printf("Random Value: %s\n", random)
+	quiet, _ = cmd.Flags().GetBool(cmdFlagNameQuiet)
 
-	if value := url.QueryEscape(random); random != value {
-		fmt.Printf("Random Value (URL Encoded): %s\n", value)
+	if outputPath, err = flagsGetFilePath(cmd.Flags(), cmdFlagNameOutputFile, cmdFlagNameDirectory); err != nil {
+		return fmt.Errorf("error getting output file path: %v", err)
+	}
```
jonasgeiler, line 266: Instead of writing a custom utility I would use the same snippet as here to get and check the file path:
https://github.com/authelia/authelia/blob/4dcf8936c22074ecb09b721a3306967121dee10a/internal/commands/storage_run.go#L439-L452

Or maybe even extract this snippet into a utility function and replace all other appearances of this snippet in the authelia CLI codebase with this new utility. But maybe that is outside the scope of this PR.

docs/content/reference/cli/authelia/authelia_crypto_rand.md
```diff
@@ -49,6 +51,8 @@ authelia crypto rand --characters 0123456789ABCDEF
   -x, --charset string      sets the charset for the random password, options are 'ascii', 'alphanumeric', 'alphabetic', 'numeric', 'numeric-hex', and 'rfc3986' (default "alphanumeric")
   -h, --help                help for rand
   -n, --length int          sets the character length for the random string (default 72)
+  -q, --quiet               suppress additional output
+  -o, --output-file string  file where the random string will be stored
```
jonasgeiler, line 55: I think to stay consistent with other commands, like [`authelia storage user totp export`](https://www.authelia.com/reference/cli/authelia/authelia_storage_user_totp_export/), [`authelia storage user webauthn export`](https://www.authelia.com/reference/cli/authelia/authelia_storage_user_webauthn_export/) and more, we should use `-f`/`--file` as the output file flag.
Especially since `-o` is being used as the "certificate organization" flag in the [`authelia crypto certificate rsa generate`](https://www.authelia.com/reference/cli/authelia/authelia_crypto_certificate_rsa_generate/) command.

internal/commands/crypto.go
```diff
@@ -53,6 +54,9 @@ func newCryptoRandCmd(ctx *CmdCtx) (cmd *cobra.Command) {
 	cmd.Flags().StringP(cmdFlagNameCharSet, "x", cmdFlagValueCharSet, cmdFlagUsageCharset)
 	cmd.Flags().String(cmdFlagNameCharacters, "", cmdFlagUsageCharacters)
 	cmd.Flags().IntP(cmdFlagNameLength, "n", 72, cmdFlagUsageLength)
+	cmd.Flags().BoolP(cmdFlagNameQuiet, "q", false, cmdFlagUsageQuiet)
+	cmd.Flags().StringP(cmdFlagNameOutputFile, "o", "", cmdFlagUsageOutputFile)
+	cmd.Flags().StringP(cmdFlagNameDirectory, "d", "", cmdFlagUsageDirectory)
```
jonasgeiler, line 59: I think having the file name and output directory as two separate flags in the same command could be confusing? The output directory or the full file path should be specified in the output file flag.

Just check out the behavior of the following commands:
```bash
authelia storage user totp export --file export.yaml
authelia storage user totp export --file nonexistant/export.yaml
mkdir existant && authelia storage user totp export --file existant/export.yaml
```

jonasgeiler (commented):
Added some suggestions for better UX and consistency. These are obviously opinionated and I'm open for discussion.

Crowley723:
> Added some suggestions for better UX and consistency. These are obviously opinionated and I'm open for discussion.

All your proposed changes sound good to me, I just want to make sure we can still use `docker run -u "$(id -u):$(id -g)" -v "$(pwd)":/keys authelia/authelia:latest authelia crypto rand --output-file /keys/random.txt`
and have it output to the user's current workong directory and not just the containers working directory.

The reason it creates all the directories in the path is because the container it's running in may or may not already contain the temporary directory used to place the secret file in the current working directory instead of the containers working directory.

jonasgeiler:
> @Crowley723 wrote:
> All your proposed changes sound good to me, I just want to make sure we can still use `docker run -u "$(id -u):$(id -g)" -v "$(pwd)":/keys authelia/authelia:latest authelia crypto rand --output-file /keys/random.txt` and have it output to the user's current working directory and not just the containers working directory.
> 
> The reason it creates all the directories in the path is because the container it's running in may or may not already contain the temporary directory used to place the secret file in the current working directory instead of the containers working directory.

When you add the `-v "$(pwd)":/keys` flag to your `docker run` command, Docker will automatically create the `/keys` directory for you.

Just try this out:
```bash
docker container run -it -u "$(id -u):$(id -g)" -v "$(pwd):/keys" alpine sh -c 'echo test > /keys/random.txt'
```
And then this, which prints an error:
```bash
docker container run -it -u "$(id -u):$(id -g)" -v "$(pwd):/keys" alpine sh -c 'echo test > /somethingelse/random.txt'
# sh: can't create /keysrand/random.txt: nonexistent directory
```

james-d-elliott:
Just closing this was superseded.

PR #7737, opened 2024-08-23
docs: Authelia Community Grafana Dashboard

hendrik1120:
Hi,

After changes to the metrics endpoint in v4.38, the current Grafana dashboard example stopped working (reported in #7003). 

To address this, I created a new dashboard designed to monitor the Authelia endpoints across multiple instances. Grafana can also be configured to send alerts if certain request thresholds are reached.

I tried to make this dashboard as future-proof as possible by minimizing overrides and using only basic Grafana features. A few people have tested the dashboard over the past few months, and I’ve incorporated their feedback.

The dashboard looks like this:
grafana

These dashboards can be subjective and vary significantly based on the use case. I hope this example serves as a good starting point for further customization. I am always happy to receive feedback.


james-d-elliott (approved):
LGTM! Thanks a lot.

PR #7798, opened 2024-09-01
fix(oidc): jwt assertions aud overly strict

james-d-elliott:
This fixes an issue where all JWT assertions are strictly checked against the Token URL. RFC7523 Section 3 states that the JWT must contain an 'aud' claim that identifies the authorization server and that the token endpoint URL may be used, not that it must be used. RFC9126 clarifies this that it should be the issuer value, and that both the token endpoint URL and pushed authorization request endpoint URL must also be accepted. This fix facilitate this. In addition it fixes a log message, and an edge case where using a JWT assertion could trigger a handled panic.

### DIFF
--- a/internal/oidc/config_test.go
+++ b/internal/oidc/config_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/authelia/authelia/v4/internal/configuration/schema"
+	"github.com/authelia/authelia/v4/internal/model"
 	"github.com/authelia/authelia/v4/internal/oidc"
 	"github.com/authelia/authelia/v4/internal/templates"
 )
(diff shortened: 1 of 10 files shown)


internal/oidc/config_test.go
```diff
@@ -320,11 +325,13 @@
 				Issuers: tc.have,
 			}
 
-			assert.Equal(t, tc.expectIntrospection, config.GetIntrospectionIssuer(tc.ctx))
-			assert.Equal(t, tc.expectIDToken, config.GetIDTokenIssuer(tc.ctx))
-			assert.Equal(t, tc.expectAccessToken, config.GetAccessTokenIssuer(tc.ctx))
-			assert.Equal(t, tc.expectAS, config.GetAuthorizationServerIdentificationIssuer(tc.ctx))
-			assert.Equal(t, tc.expectJARM, config.GetJWTSecuredAuthorizeResponseModeIssuer(tc.ctx))
+			ctx := context.WithValue(tc.ctx, 0, 1)
```
autheliabot, line 328: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>SA1029: should not use built-in type int as key for value; define your own type to avoid collisions (staticcheck)

PR #7824, opened 2024-09-05
chore: Docker - Remove `VOLUME` directive

polarathene:
The `VOLUME` directive provides **no value** to an image. It **only causes problems.**

For detailed justification, please see below context or my past writeup on this subject:
- https://github.com/kanidm/kanidm/pull/2948
- https://github.com/ory/hydra/issues/3683

---

Anonymous volumes are created when an image has a `VOLUME` directive. They will initialize by copying any of the existing content at the mount point, unlike a bind mount volume which replaces.
  - With `docker run --rm ...`, each container instance started will create a new anonymous volume. Without the `--rm`, these will accumulate pointlessly, the user may not be aware of it.
  - Docker Compose behaves differently, with additional logic to preserve the same anonymous volume across instances of the container for a given compose project.


james-d-elliott (approved):
LGTM

aeneasr (approved):

PR #7832, opened 2024-09-08
Documenting so optional config in Nextcloud that helps if you already have local users

jagg2:
Change is based on [this comment](https://github.com/nextcloud/user_oidc/issues/565#issuecomment-2120800504), as this is clearly an issue some people have had.

### DIFF
--- a/docs/content/integration/openid-connect/nextcloud/index.md
+++ b/docs/content/integration/openid-connect/nextcloud/index.md
@@ -189,6 +189,14 @@ To configure [Nextcloud] to utilize Authelia as an [OpenID Connect 1.0] Provider
 ],
 ```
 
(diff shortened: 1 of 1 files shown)


james-d-elliott:
Thanks for taking the time to be involved. I think we're unfortunately however going to not take it upon ourselves to document something that should realistically be fixed upstream.

Essentially the only claims a user should be linked to are the `sub` and `iss` claims. Relying Parties should ensure a flow exists which allows linking existing accounts to these claims to ensure the guarantees offered within OpenID Conect 1.0 around claim stability are met (only `iss` and `sub` are guaranteed to be the same, and the `preferred_username` and `email` claims can change without warning i.e. mutable).

PR #7882, opened 2024-09-17
Update example lite configuration.yml

darylf:
The mail domain was initialized as "exmaple.com", which is an actual domain owned by someone. It's better practice to use example.com, a reserved second-level domain


james-d-elliott:
This looked like a correct fix, why was it closed?

PR #7900, opened 2024-09-21
fix(web): authenticated layout inconsistent

james-d-elliott:
The upgrade to MUI6 caused a layout issue with the authenticated view in the instance of the WebAuthn option, this resolves that issue.

### DIFF
--- a/web/src/layouts/MinimalLayout.tsx
+++ b/web/src/layouts/MinimalLayout.tsx
@@ -36,8 +36,8 @@ const MinimalLayout = function (props: Props) {
 
     return (
         <Box>
-            <AppBar position="static" color="transparent" elevation={0}>
-                <Toolbar variant="regular">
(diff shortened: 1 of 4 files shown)


KonVei82:
image
This is what i am getting after running docker pull authelia/authelia:4.38.12
I did clear cache under dev tools in firefox

Crowley723:
> This is what i am getting after running docker pull authelia/authelia:4.38.12

How do you have authelia deployed, if your not using the "latest" image tag, it won't update to the latest version.
Additionally, you have to restart the container for the new image to be used.

KonVei82:
Running it as part of a docker compose stack with latest as a tag.
Did "sudo docker compose pull" followed by “docker compose down authelia” and “docker compose up -d authelia” which did indeed pull 4.38.12 as shown in the logs
time="2024-09-22T13:24:53+03:00" level=info msg="Authelia v4.38.12 is starting"

PR #7909, opened 2024-09-24
fix(web): inconsistent no 2FA authenticated layout

Crowley723:
inconsistent authenticated layout for authelia instances with no configured 2FA domains. 
`auth.example.com/authenticated`


james-d-elliott (approved):
LGTM

james-d-elliott (approved):

KonVei82:
I am still getting the same issue it seems.
This one is from using 2FA:
image
And here is without it:
image
This is using latest docker build v4.38.14 as confirmed by log files:
time="2024-09-29T16:57:34+03:00" level=info msg="Authelia v4.38.14 is starting"

I did clear browser cache prior to taking screenshots.

PR #7943, opened 2024-09-29
docs: update Caddy session header removal regex

deanrock:
Regex in documentation under `Removing the Authelia Session Header` only works if there are multiple cookies involved, and `authelia_session` is not the last one.

If `authelia_session` is the last one, tailing `;` is not present, which means that regex doesn't match it.

Updated regex seems to work if there is only authelia cookie present, and also if there are other cookies following it (which are correctly passed upstream).


james-d-elliott (approved):
LGTM

PR #7952, opened 2024-10-01
ci(buildkite): fix pre-exit hook conditionally being ignored

nightah:
This change ensures that the subsequent hooks after the `post-command` hook are not ignored due to a directory change which means hooks no longer can be found at `.buildkite/hooks`.


james-d-elliott (approved):
LGTM

PR #8021, opened 2024-10-13
fix(web): reset password absent progress feedback

Crowley723:
In the case where the password reset email takes longer than expected, a user should be given an indication as to what is happening.

This PR disables the login form and buttons when once a user initiates the password reset and shows a loading bar. The bar is visible and the form disabled until either the email succeeds or fails. This should also help minimize duplicate emails from clicking the button multiple times.

Related Reading: [3 Important (time) Limits for UX](https://www.nngroup.com/articles/response-times-3-important-limits/)


james-d-elliott (changes requested):
LGTM, I'd just suggest using the circular progress on the button similar to the other areas of the UI.

Crowley723:
> LGTM, I'd just suggest using the circular progress on the button similar to the other areas of the UI.
Sorry, would you mind pointing me to a place where there is circular progress on the buttons themselves?

james-d-elliott (changes requested):
One change I think makes sense also sorry I missed one the first time I think

Crowley723:
Moved the state logic into the async function - logic flow is much simpler now.

Also the text field error state should reset when password reset is triggered to fix the case where there is an error that persists even if an email is successfully sent.

james-d-elliott (approved):
LGTM

PR #8054, opened 2024-10-18
Outline: remove error-creating quotes

axelsimon:
Remove the double quotes around the OIDC scope, which cause an error (cf. https://github.com/authelia/authelia/discussions/5671)


james-d-elliott (approved):
LGTM

PR #8055, opened 2024-10-18
refactor(suites): remove commentary from suite based user databases

nightah:
This change ensures all user databases are consistent across the suites where utilised.


james-d-elliott (approved):

Crowley723 (approved):
👍

Issue #7918, opened 2024-09-25
Indicate why 2fa settings page is blank when no two factor domains are configured.


Crowley723:
### Description

There are a couple ways to go about this:
1. Disable/hide the two-factor settings page when there are no two factor domains for users with no credentials.
2. Display a message on the blank two factor page when there are no 2fa domains.

### Use Case

Currently when Authelia doesn't have any domains configured to use a second factor, no users are required/able to add a second factor to their account. Additionally, the Two Factor page in user settings is blank and shows no indication of why.

It could be a little confusing as to why the page is blank and might cause some confusion for administrators thinking there is a misconfiguration.

### Pre-Submission Checklist

- [X] I agree to follow the [Code of Conduct](http://www.authelia.com/code-of-conduct)

- [X] I have checked for related issues and checked the documentation

Issue #7660, opened 2024-08-10
Add inspect menu to totp credentials in user settings 


Crowley723:
### Description

Add "details" menu to totp credentials, similar to webauthn credentials.

### Use Case

#7650 
> Creation time/last used time doesn't need to be very specific, having the exact date and time when a credential was created/used isn't a super big issue. [isn't super helpful]
> For webauthn credentials, users still have the ability to see the exact date and time by inspecting the credential.

I think that it might be worth adding an inspect menu for the totp credentials as well, this would allow us to hide the algorithm type, digits, and time interval as well as give users the ability to see the long form date of when their credential was created.

I don't think people need to have immediate access to totp digital count, algorithm, and time interval and it clutters the widget. Additionally, users can have only one totp credential so (currently) there is no benefit to being able to distinguish between multiple (via digits, algorithm, interval)

This should be a fairly straightforward change.

### Details

Move 
- long form created date
- long form last used date
- totp digit count
- totp algorithm 
- totp time interval
- created by user(?)
- times used(?)
- any other relevant information
Into a info dialog similar to the webauthn credentials

This would help minimize clutter on the widget.

### Pre-Submission Checklist

- [X] I agree to follow the [Code of Conduct](http://www.authelia.com/code-of-conduct)

- [X] I have checked for related issues and checked the documentation

Crowley723:
Might be a good first issue.

james-d-elliott:
SGTM, can push it through as a ux fix if you ask me.

PR #8062, opened 2024-10-19
feat(web): add inspect menu to totp credential widget

Crowley723:
This moves Algorithm, Digits, and Period from the credential qualifier field to the new inspect dialog. This improves the appearance of the totp widget by removing unnecessary information. 
The new inspect dialog ensures the user doesn't lose access to any of the important information regarding the totp credential. This also allows users to view the explicit time and date the credential was created, as well as the last time it was used.

new otp widget
new otp dialog

Resolves: [#7660 ](https://github.com/authelia/authelia/issues/7660) and #7918


Crowley723:
This should also resolve [#7918 ](https://github.com/authelia/authelia/issues/7918)

james-d-elliott (approved):
LGTM

PR #8119, opened 2024-10-28
fix(configuration): utilise updated psl for domain validation

### LINKED ISSUE #8074: PSL updates
### Description

PSL (https://publicsuffix.org/) updates are extremely behind in authelia. Domains removed from the list may take months to become eligeble for protection under authelia. Can the list be updated on each authelia build?

### Use Case

A domain is listed on PSL. The domain is removed, authelia should acknowledge that.

### Details

_No response_

### Documentation

_No response_

### Pre-Submission Checklist

nightah:
`golang.org/x/net/publicsuffix` is updated infrequently and there have been [discussions over time](https://github.com/golang/go/issues/15518) which seem to imply that more frequent updates of the list are [unlikely to be considered](https://github.com/golang/go/issues/15518#issuecomment-217312171).

The proposed library [`github.com/weppos/publicsuffix-go`](https://github.com/weppos/publicsuffix-go) keeps itself updated with the list maintained at the PSL source and has a backwards compatible API which we can use as a drop-in replacement.

Fixes #8074.


Crowley723 (approved):
LGTM
The new library doesn't have a release that contains the updated list so it may still lag behind a bit in the future.

nightah:
> The new library doesn't have a release that contains the updated list so it may still lag behind a bit in the future.

It's usually updated daily so this shouldn't be the case. Looks like I was using the last tagged, @james-d-elliott picked that up and fixed it by pinning to the latest commit sha.

james-d-elliott (approved):
LGTM

PR #8130, opened 2024-10-29
fix(configuration): templating panic edge case

james-d-elliott:
This fixes an edge case for templating that causes a panic without helpful information.


nightah (approved):
LGTM.

PR #8146, opened 2024-11-01
fix(session): add connection timeout and retry options to redis

nightah:
This change adds the ability to configure the following options on the Redis connection:
* Configurable connection dial timeout (defaults to 5 seconds)
* Configurable max retries on failed commands (defaults to 0)


james-d-elliott (approved):
LGTM

PR #8318, opened 2024-11-24
fix(web): internalization

hendrik1120:
Hi,

this should fix the non internalized strings in key areas of the web frontend.

I tried to test this, but the standalone suite doesn't seem to recognize the translation files at all.
If this is not the correct workflow or there are other issues, feel free to use this as a bug report instead.


james-d-elliott (approved):
LGTM

PR #8321, opened 2024-11-24
feat(middlewares): tokenized bucket rate limit

### LINKED ISSUE #1947: Denial of Service prevention mechanism
## Feature Request

We'd need to reject some requests when there are too many of them (number being customizable). That would prevent DDoS attacks on the authentication system which could take down the entire infrastructure otherwise if not prevented at the proxy level.

We should also probably add a link to a doc explaining how to configure for DDoS prevention in the reverse proxy in the meantime.

### Description

- Make the number of requests the server can handle configurable. No limit if not configured.
- Start sending 429 responses once too many requests are being processed.

### Use Case

Prevent DDoS attacks both from inside and outside the cluster. For the outside, it's better configured in the reverse proxy but we'd rather protect Authelia anyway since it's a central piece of the infrastructure.

### LINKED ISSUE #7353: SMTP Rate Limiting
### Description

Allow administrators to set a rate limit for outgoing smtp traffic.

### Use Case

Gmail SMTP server only allows 2000 messages per 24 hour rolling period. This amounts to about 83 emails/hour.
On Authelia implementations with larger amounts of users and multiple services using the same Gmail account for notifications its possible that this limit could be reached.

### Details

Rate limiting can be a complex problem. One user could conceivably cause other user's otp emails to be dropped by hitting the rate limit. Is the rate per user? Is there a global rate for all traffic?

### Documentation

Cloudflare has a [good blog post](https://blog.cloudflare.com/counting-things-a-lot-of-different-things/) on the algorithm used on their edge servers to help mitigate attacks while minimizing the affect on legitimate traffic.

I think that ideally, requests should be limited on a per distinct user basis. This way, malicious traffic can be mitigated while legitimate traffic can be passed with minimal latency.

Mentioned in the cloudflare post is the method they use to cache currently enforced mitigations to maximize speed.

Here is the diagram from cloudflare modified for an example email request.
rate limit cacheing

### Pre-Submission Checklist

james-d-elliott:
This adds rate limits to the TOTP second factor endpoint, the Duo second factor endpoint, Session Elevation endpoint, and the Reset Password endpoint. This protection exists as several configurable tokenized buckets anchored to the users remote IP address. In the event the rate limit is exceeded by the user the middleware will respond with a 429 status, a Retry-After header, and JSON body indicating it's rate limited, which the UI will gracefully handle. This has several benefits that compliment the 1FA regulation, specifically in simple architectures it limits the number of SMTP sends a unique client can make, as well as the number of requests a particular client can make in general on specific endpoints where too many requests may indicate either a fault or some form of abuse.

Closes #7353, Closes #1947


internal/middlewares/rate_limiting.go
```diff
@@ -0,0 +1,88 @@
...
+// The RateLimitBucket describes a limit (number of seconds), and a burst (number of events) that can occur for a given
+// rate limiter.
+type RateLimitBucket struct {
+	Limit float64
+	Burst int
+}
+
+// NewIPRateLimit given a series of RateLimitBucket items produces an AutheliaMiddleware which handles requests based
+// on the IPRateLimitBucket.
+func NewIPRateLimit(bs ...RateLimitBucket) AutheliaMiddleware {
```
autheliabot, line 19: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>unnecessary leading newline (whitespace)

web/src/views/Settings/Common/SecondFactorMethodOneTimePassword.tsx
```diff
@@ -50,17 +52,27 @@
...
+                setState(State.Failure);
+            }
         } catch (err) {
             console.error(err);
             setState(State.Failure);
         }
 
         setPasscode("");
-    }, [passcode, config, props]);
+    }, [passcode, config, translate]);
```
autheliabot, line 75: ⚠️ **[eslint]** <[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>React Hook useCallback has a missing dependency: 'createErrorNotification'. Either include it or remove the dependency array.

internal/configuration/validator/server.go
```diff
@@ -158,6 +161,43 @@
...
+	config.Buckets = make([]schema.ServerEndpointRateLimitBucket, len(defaults.Buckets))
+
+	for i := range defaults.Buckets {
+		config.Buckets[i] = schema.ServerEndpointRateLimitBucket{
+			Period:   weight * defaults.Buckets[i].Period,
+			Requests: defaults.Buckets[i].Requests,
+		}
+	}
+
+	return
```
autheliabot, line 198: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>S1023: redundant `return` statement (gosimple)

internal/middlewares/rate_limiting.go
```diff
@@ -0,0 +1,191 @@
...
+	defer l.mu.Unlock()
+
+	var ok bool
+
+	if limiter, ok = l.bucket[key]; !ok {
+		limiter = l.new(key)
+	}
+
+	return limiter
+}
```
Crowley723, line 135: Question: Does holding the lock for the entire function make sense?

I am curious if the  `l.new()` call is fast enough that holding the lock while it executes is okay. Potentially this slows down bucket-miss handling. Every new bucket would have to wait for the currently executing one to complete memory allocation and initialization.
Would it be better to do an initial lock, check, unlock. Check if it already exists, if so, return. Otherwise create a new limiter (`l.NewLimiter()`) then lock, check again (in case another fetch added it already) then add the newLimiter and return.

Something like this
```go
func (l *IPRateLimitBucket) Fetch(key string) (limiter *RateLimiter) {
    l.mu.Lock()
    limiter, ok := l.bucket[key]
    l.mu.Unlock()
    
    if ok {
        return limiter
    }
    // pre-allocate and initialize non-existent limiter
    newLimiter := l.newLimiter(key);
    
    l.mu.Lock()
    defer l.mu.Unlock()
    
    //check again
    if limiter, ok = l.bucket[key]; ok {
        return limiter
    }
    
    l.bucket[key] = newLimiter
    return newLimiter
}
```

internal/middlewares/rate_limiting.go
```diff
@@ -0,0 +1,191 @@
...
+		return func(ctx *AutheliaCtx) {
+			var (
+				retryAfter time.Duration
+			)
+
+			for i, bucket := range buckets {
+				limiter := bucket.FetchCtx(ctx)
+
+				if !limiter.Allow() {
+					reservation := limiter.ReserveN(time.Now().UTC(), 1)
```
Crowley723, line 70: Question: Would it be beneficial to move the `reservation` logic to a helper function? 

Since its just here to check when the next token is available and not actually being used, its feels weird to have it here.

Something like `timeToNextToken()` might make more sense and would improve readability.

Crowley723 (changes requested):
Just a couple questions.

internal/middlewares/rate_limiting.go
james-d-elliott: So I've tested something identical to this before, the benchmarks indicated at most a 3% difference (something like 10ns). I'd consider it a premature optimization in this situation since you can get even more performance by using a RWMutex (about 5-10% depending) and avoiding defer.

It's more likely to cause the introduction of bugs when making enhancements or fixes unless the performance impact is significant, at least in my opinion.

internal/middlewares/rate_limiting.go
```diff
@@ -0,0 +1,191 @@
...
+
+	ticker := time.NewTicker(time.Minute * 30)
+
+	go func() {
+		for range ticker.C {
+			for _, bucket := range buckets {
+				bucket.GC()
+			}
+		}
+	}()
```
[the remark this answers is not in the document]
james-d-elliott: It's a good question but technically go will end all goroutines without calling defer. At least the last time I checked.

web/src/views/Revoke/RevokeResetPasswordTokenView.tsx
```diff
@@ -25,12 +25,16 @@ const RevokeResetPasswordTokenView = function () {
...
 
         if (ok) {
             createSuccessNotification(translate("Successfully revoked the Token"));
         } else {
-            createErrorNotification(translate("Failed to revoke the Token"));
+            if (status === 429) {
+                createErrorNotification(translate("You have made too many requests"));
+            } else {
+                createErrorNotification(translate("Failed to revoke the Token"));
+            }
```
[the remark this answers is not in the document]
james-d-elliott: Yeah it'd be better in this instance, think this was introduced when refactoring and simplifying the API.

internal/middlewares/rate_limiting.go
james-d-elliott:
```go
func NewRateLimiter(newBucket func(bucket RateLimitBucketConfig) RateLimitBucket, handler RequestHandler, bs ...RateLimitBucketConfig) AutheliaMiddleware {
	buckets := make([]RateLimitBucket, len(bs))

	for i, b := range bs {
		buckets[i] = newBucket(b)
	}

	if handler == nil {
		handler = HandlerRateLimitAPI
	}

	ticker := time.NewTicker(time.Minute * 30)

	go func() {
		for range ticker.C {
			for _, bucket := range buckets {
				bucket.GC()
			}
		}
	}()

	return func(next RequestHandler) RequestHandler {
		return func(ctx *AutheliaCtx) {
			var (
				retryAfter time.Duration
			)

			for i, bucket := range buckets {
				limiter := bucket.FetchCtx(ctx)

				if !limiter.Allow() {
					reservation := limiter.ReserveN(time.Now().UTC(), 1)
					limiter.updated = time.Now().UTC()

					ctx.Logger.WithFields(map[string]any{"bucket": i + 1, "delay": reservation.Delay().Seconds()}).Warn("Rate Limit Exceeded")

					if reservation.Delay() > retryAfter {
						retryAfter = reservation.Delay()
					}

					reservation.Cancel()
				}
			}

			if retryAfter > 0 {
				ctx.Response.Header.SetBytesK(headerRetryAfter, time.Now().UTC().Add(retryAfter).Format(http.TimeFormat))
				ctx.SetStatusCode(fasthttp.StatusTooManyRequests)

				handler(ctx)

				return
			}

			next(ctx)
		}
	}
}
```

vs

```go
func NewRateLimiter(newBucket func(bucket RateLimitBucketConfig) RateLimitBucket, handler RequestHandler, bs ...RateLimitBucketConfig) AutheliaMiddleware {
	buckets := make([]RateLimitBucket, len(bs))

	for i, b := range bs {
		buckets[i] = newBucket(b)
	}

	if handler == nil {
		handler = HandlerRateLimitAPI
	}

	ticker := time.NewTicker(time.Minute * 30)

	go func() {
		for range ticker.C {
			for _, bucket := range buckets {
				bucket.GC()
			}
		}
	}()

	timeToNextToken := func(ctx *AutheliaCtx, i int, bucket RateLimitBucket) (delay time.Duration) {
		limiter := bucket.FetchCtx(ctx)

		if !limiter.Allow() {
			reservation := limiter.ReserveN(time.Now().UTC(), 1)
			limiter.updated = time.Now().UTC()

			ctx.Logger.WithFields(map[string]any{"bucket": i + 1, "delay": reservation.Delay().Seconds()}).Warn("Rate Limit Exceeded")

			delay = reservation.Delay()

			reservation.Cancel()
		}

		return delay
	}

	return func(next RequestHandler) RequestHandler {
		return func(ctx *AutheliaCtx) {
			var (
				retryAfter time.Duration
			)

			for i, bucket := range buckets {
				delay := timeToNextToken(ctx, i, bucket)
				
				if delay > retryAfter {
					retryAfter = delay
				}
			}

			if retryAfter > 0 {
				ctx.Response.Header.SetBytesK(headerRetryAfter, time.Now().UTC().Add(retryAfter).Format(http.TimeFormat))
				ctx.SetStatusCode(fasthttp.StatusTooManyRequests)

				handler(ctx)

				return
			}

			next(ctx)
		}
	}
}
```

Crowley723: Fair enough. Considering the overall length of the function, it doesn't make much difference.

Crowley723 (approved):
👍

internal/middlewares/rate_limiting.go
james-d-elliott: Yeah I'm 50/50 on it, it would probably be a more obvious choice if it was reused. We can probably refactor it later if necessary.

PR #8380, opened 2024-12-04
fix(configuration): oidc subject not validated

james-d-elliott:
This fixes an issue where the OIDC subject was not validated in the authz policies correctly.


internal/configuration/validator/const.go
```diff
@@ -195,10 +195,12 @@ const (
...
-	errFmtOIDCPolicyMissingOption        = "identity_providers: oidc: authorization_policies: policy '%s': option '%s' is required"
-	errFmtOIDCPolicyRuleMissingOption    = "identity_providers: oidc: authorization_policies: policy '%s': rules: rule #%d: option '%s' is required"
+	errFmtOIDCPolicyInvalidName         = "identity_providers: oidc: authorization_policies: authorization policies must have a name but one with a blank name exists"
+	errFmtOIDCPolicyInvalidNameStandard = "identity_providers: oidc: authorization_policies: policy '%s': option '%s' must not be one of %s but it's configured as '%s'"
+	errFmtOIDCPolicyMissingOption       = "identity_providers: oidc: authorization_policies: policy '%s': option '%s' is required"
+	errFmtOIDCPolicyRuleMissingOption   = "identity_providers: oidc: authorization_policies: policy '%s': rules: rule #%d: option '%s' is required"
+	errFmtOIDCPolicyRuleSubjectInvalid  = "identity_providers: oidc: authorization_policies: policy '%s': rules: rule #%d: 'subject' option '%s' is " +
+		"invalid: must start with 'user:' or 'group:'"
```
Crowley723, line 203: Curious why this has two strings? Is it for word wrap?

Crowley723 (dismissed):

internal/configuration/validator/const.go
james-d-elliott: Generally to keep the line length similar throughout the codebase. It should allow a tri-view comparison with a QHD+ screen in most scenarios.

PR #8470, opened 2024-12-17
docs: update portainer oidc config

leedotpang:
While implementing the OIDC config from the docs I ran in to the below error and was able to quickly fix it by updating the `token_endpoint_auth_method` explained in the error. 

> time="2024-12-17T10:00:41-05:00" level=error msg="Access Request failed with error: Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method). The request was determined to be using 'token_endpoint_auth_method' method 'client_secret_post', however the OAuth 2.0 client registration does not allow this method. The registered client with id 'portainer' is configured to only support 'token_endpoint_auth_method' method 'client_secret_basic'. Either the Authorization Server client registration will need to have the 'token_endpoint_auth_method' updated to 'client_secret_post' or the Relying Party will need to be configured to use 'client_secret_basic'." method=POST path=/api/oidc/token remote_ip=172.18.0.1 stack="github.com/authelia/authelia/v4/internal/handlers/handler_oidc_token.go:25


docs/content/integration/openid-connect/portainer/index.md
```diff
@@ -88,6 +89,7 @@ To configure [Portainer] to utilize Authelia as an [OpenID Connect 1.0] Provider
    9. Redirect URL: `https://portainer.{{< sitevar name="domain" nojs="example.com" >}}`
    10. User Identifier: `preferred_username`
    11. Scopes: `openid profile groups email`
+   12. Token Endpoint Auth Method: `client_secret_post`
```
james-d-elliott, line 92: Please show the screenshot of this in Portainer using the example values.

james-d-elliott (changes requested):
Also please update the version, looks like a breaking change from portainer's end.

docs/content/integration/openid-connect/portainer/index.md
leedotpang: Actually, that was a mistake! This isn't a portainer setting. Well, at least not directly. 

It looks like [they recently added](https://github.com/portainer/portainer/commit/ffc66647f867393f6274dd747c47494eb54f53f1#diff-6c81c4be8002000c7704490af244731ad3de1e41e161a26606e95c2ef1b11bcaR91) a new "Auth Style" setting that allows users to select "In Headers" or "In Params" which toggles between `client_secret_basic` and `client_secret_post` respectively. Since the default is "Auto Detect" and it defaulted to "In Params" I've updated this checklist and screenshot accordingly.

james-d-elliott: I think we should set it to `In Params` explicitly. Auto Detection mechanisms can cause a lot of confusion for users when debugging as they can produce multiple errors rather than single errors.

leedotpang: great callout. updated!

james-d-elliott (approved):
LGTM

hendrik1120:
Funny, I am on the way newer 2.21.4 BE and this has never been an issue as the `Auth Style` defaults to `Auto Detect`.
@leedotpang Did you change the default by any chance?

hendrik1120:
Ok, that could be the case. I am unable to even find a docker image of 2.12.4, which is why I tested the latest STS version instead of yours. Can you maybe share link where I can find an image of that version?

Your change is not wrong by any means, but 2.12.4 is at least 2 years old and without any support. We could also set `Auth Style` to `In header` because `client_secret_basic` is the default and more commonly used if `Auto Detect` has issues.

leedotpang:
@hendrik1120 You're right! I must have had a dyslexic moment because I updated that version to `2.12.4` instead of `2.21.4`. Also, great callout with the last point -- especially if `client_secret_basic` is the most common setting, maybe that was the right decision.

PR #8473, opened 2024-12-18
docs: fix portainer version

leedotpang:
I updated the semver string incorrectly for the portainer oidc docs -- the `latest` release on docker [and github](https://github.com/portainer/portainer/releases/tag/2.21.4) is version `2.21.4`.


leedotpang:
Also, I'd hate to undo all my _"hard work"_ but I'd be easily convinced to rollback the additional token auth endpoint setting and set the Portainer input to `In Headers` if `client_secret_basic` is the [more common implementation](https://github.com/authelia/authelia/pull/8470#issuecomment-2551354993)

james-d-elliott:
While it may be more common my personal issue with it is it's not as secure (though not by much) and the encoding of this value is more commonly incorrectly implemented. The method `client_secret_basic` is effectively discouraged. So I'm quite content with the new guide as is.

leedotpang:
I'm convinced. Thanks for confirming -- otherwise, this is a relatively minor change. Apologies for the over-sight 😬

james-d-elliott (approved):

PR #8502, opened 2024-12-23
Add missing parameter in express integration

Tchoupinax:
I copied the example as minimum example and redirect_url param is required. Also, this parameter is not super understandable for a beginner so I think it is relevant to add it in the example.


james-d-elliott:
Pretty sure the baseurl just needs to be adjusted. Can you try fixing that instead?

james-d-elliott:
Pretty sure the linked PR fixes the actual issue. Let me know if not.

PR #8530, opened 2024-12-29
refactor: docker-compose begone

james-d-elliott:
This removes any reference and usage of the deprecated docker-compose python tool.


archef2000:
In the setup.sh file there is still:
* a check for the docker-compose command: [examples/compose/local/setup.sh#L30](https://github.com/authelia/authelia/blob/5fa95165ddc0113f0dc99ca0eaf58e7623067f41/examples/compose/local/setup.sh#L30)
* Here the file is not renamed: [examples/compose/local/setup.sh#L47](https://github.com/authelia/authelia/blob/5fa95165ddc0113f0dc99ca0eaf58e7623067f41/examples/compose/local/setup.sh#L47) + L48
* The two sed commands also have the old filename: [examples/compose/local/setup.sh#L67](https://github.com/authelia/authelia/blob/5fa95165ddc0113f0dc99ca0eaf58e7623067f41/examples/compose/local/setup.sh#L67) + L69

The only other reference of docker-compose is in the ".renovaterc" file, but that seems ok looking at the documentation: https://docs.renovatebot.com/modules/manager/#supported-managers

PR #8547, opened 2025-01-02
refactor(configuration): include defaults map source

james-d-elliott:
This includes a source which can be used to set defaults from a map. This will be used in various upcoming PR's.


internal/configuration/defautls.go
Crowley723, line 1: There is a typo in the name of the file, it should be `defaults.go`.

Crowley723 (approved):

PR #8549, opened 2025-01-02
Update Jellyfin SSO documentation

Skyb0rg007:
Integration with Jellyfin SSO requires a `/r/<provider>/` redirect URI in addition to the `/redirect/<provider>` URI (docs source: https://github.com/9p4/jellyfin-plugin-sso#examples). This PR adds the other URI to the `required_uris` field of the Authelia configuration.

Edit: You may also want to update the URIs to use `/Authelia` (capitalized) because it needs to match the name in Jellyfin and the docs use "Authelia" as the example.

### DIFF
--- a/docs/content/integration/openid-connect/jellyfin/index.md
+++ b/docs/content/integration/openid-connect/jellyfin/index.md
@@ -61,6 +61,7 @@ identity_providers:
         require_pkce: true
         pkce_challenge_method: 'S256'
         redirect_uris:
+          - 'https://jellyfin.{{< sitevar name="domain" nojs="example.com" >}}/sso/OID/r/authelia'
           - 'https://jellyfin.{{< sitevar name="domain" nojs="example.com" >}}/sso/OID/redirect/authelia'
         scopes:
           - 'openid'


hendrik1120:
Hi, I recently tested this integration, did you observe any issues without the additional `redirect_uri`present?
As far as I know, the `/r/<provider>/` was only needed for legacy support.
If you take a look here: https://github.com/9p4/jellyfin-plugin-sso/blob/main/providers.md only the new style is present.

If I remember correctly, as soon you are using `/sso/OID/start/PROVIDER_NAME` the redirect URI should always be `/redirect/<provider>/`

I missed the capitalization error, thanks for pointing that out.

james-d-elliott:
I'm closing this due to lack of a response. We can reopen it if there's a response and it can be justified. We do however really appreciate your involvement in contributing regardless and do not want you to feel this devalues that in any way.

Skyb0rg007:
Sorry for the lack of a response. I did have an issue without the additional `redirect_uri`, though I believe that this was only for the initial setup of the OIDC and not required for logins after the configuration settings are setup in Jellyfin. I unfortunately don't have time to test this out this week but can comment / create a new PR once I can reproduce.

PR #8736, opened 2025-02-12
feat: properly set lang attribute for document root

### LINKED ISSUE #8729: Google Translate translates the already translated user interface
### Version

v4.38.17
v4.37.5

### Browser

Chrome

### Deployment Method

Docker

### Reverse Proxy

NGINX

### Reverse Proxy Version

1.27.1
1.21.4

### Description

Google Translate processes the `<html lang="en"> ` tag, which causes it to translate the already translated user interface again. This results in a very random translation.

### Reproduction

Set German as the default browser language. 

Image

The buttons are correct translated by Authelia. 

If you configure the website with Google Translate settings to "translate English by default":

Image

The buttons are translated randomly

Image

"Zurücksetzen" (en: reset) is translated as "Speichern" (en: save)
"Abbrechen" (en: cancel) is translated as "Reservierung" (en: reservation)

### Expectations

_No response_

### Configuration (Authelia)

```yaml
not relevant
```

### Build Information

```shell
not relevant
```

### Logs (Authelia)

```shell
not relevant
```

### Logs (Proxy / Application)

```shell
not relevant
```

### Documentation

_No response_

### Generative AI

No

### Pre-Submission Checklist

- [x] This is a bug report and not a support request

- [x] I have read the security policy and this bug report is not a security issue or security related issue

- [x] I have either included the complete configuration file or I am sure it's unrelated to the configuration

- [x] I have either included the complete debug / trace logs or the output of the build-info command if the logs are not relevant

- [x] I have provided all of the required information in full with the only alteration being reasonable sanitization in accordance with the [Troubleshooting Sanitization](https://www.authelia.com/r/sanitize) reference guide

- [x] I have checked for related proxy or application logs and included them if available

Crowley723:
Sets root element's lang attribute based on i18n.language and updates it on language changes.

Resolves #8729

### DIFF
--- a/cmd/authelia-gen/templates/web_i18n_index.ts.tmpl
+++ b/cmd/authelia-gen/templates/web_i18n_index.ts.tmpl
@@ -56,6 +56,16 @@ i18n.use(Backend)
             escapeValue: false,
         },
         debug: false,
+    })
+    .then(() => {
+        const rootElement = document.getElementById("root");
+        if (rootElement) {
+            rootElement.lang = i18n.language;
+
+            i18n.on("languageChanged", (lng) => {
+                rootElement.lang = lng;
+            });
+        }
     });
 
 export default i18n;
--- a/web/src/i18n/index.ts
+++ b/web/src/i18n/index.ts
@@ -225,6 +225,16 @@ i18n.use(Backend)
             escapeValue: false,
         },
         debug: false,
+    })
+    .then(() => {
+        const rootElement = document.getElementById("root");
+        if (rootElement) {
+            rootElement.lang = i18n.language;
+
+            i18n.on("languageChanged", (lng) => {
+                rootElement.lang = lng;
+            });
+        }
     });
 
 export default i18n;


hendrik1120:
This does not resolve the issue. Tested on Chrome, Firefox, Safari.

Crowley723:
> This does not resolve the issue. Tested on: Chrome, Firefox, Safari.

I probably should have conveyed what was discussed elsewhere.

We are going to maintain `lang="en"` on the top level HTML element because we can't guarantee translations for the [index.html](https://github.com/authelia/authelia/blob/42f2bb82ace502b88b8810b67551feca4c9b53b2/web/index.html) (meta descriptions, noscript text, etc) Per WCAG 3.1.1

As a compromise, we are setting the lang attribute on the [root div](https://github.com/authelia/authelia/blob/42f2bb82ace502b88b8810b67551feca4c9b53b2/web/index.html#L27) element to reflect the current i18n language as we can make a best effort to ensure the content of the element matches the declared language.

Supporting a single translation service's behavior would create an implicit commitment to support other browser translation implementations, which is outside the cope of our accessibility goals. Removing the lang attribute completely is not an option as it violates the WCAG requirements.

We are still looking to see if there is a solution but its difficult because the documentation from Google is thin in this respect.

### References:
[Understanding language of a page](https://www.w3.org/WAI/WCAG21/Understanding/language-of-page.html)
[Test Cases](https://www.w3.org/WAI/standards-guidelines/act/rules/ucwvc8/proposed/)

hendrik1120:
> We are going to maintain lang="en" on the top level HTML element because we can't guarantee translations for the [index.html](https://github.com/authelia/authelia/blob/42f2bb82ace502b88b8810b67551feca4c9b53b2/web/index.html) (meta descriptions, noscript text, etc).

I disagree with this approach, as it simply doesn't work with any browser I tested.

We are talking about 3 sentences in `index.html` which no user during normal operation will see. Yes, a buggy screen reader might read the title `Login - Authelia`, but that should be all. I see enough wiggle room in the specification since 99.9% of the page is correctly translated.

PR #8813, opened 2025-02-23
fixed homarr oidc connect

DaniW42:
the oidc config for homarr was outdated and didn't worked anymore. the part with the automatic permission assignment isn't available anymore. instead a note to the permission sync was added.

### DIFF
--- a/docs/content/integration/openid-connect/homarr/index.md
+++ b/docs/content/integration/openid-connect/homarr/index.md
@@ -21,9 +21,9 @@ seo:
 ## Tested Versions
 
 * [Authelia]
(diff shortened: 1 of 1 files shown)


james-d-elliott (approved):

DaniW42:
That coderabbitai thing made me aware of @rkj's PR regarding the NEXTAUTH_URL... I am using Nginx Proxy Manager with Authelia and Homarr. For me it works without the env variable and I dropped it because the official Homarr docs do not mention it either.

Although, I asked the Homarr Docs AI and it mentioned that "Several support threads mention the importance of setting NEXTAUTH_URL". So now I am a little bit confused if it should be mentioned or not.

james-d-elliott:
It's fine we can add it back if it is later.

PR #8876, opened 2025-03-04
docs: add missing middleware in traefik example

Lynilia:
The Docker Compose Traefik example references the authelia-basic@docker middleware which is not defined.


james-d-elliott (changes requested):
The /api/verify endpoint is deprecated.

PR #8884, opened 2025-03-05
feat: build from authelia/base base image

nightah:
This change builds based off of our own base libc image created from scratch, we are also now deprecating our musl builds.


james-d-elliott (approved):
LGTM

PR #8899, opened 2025-03-07
fix: add more detailed errors for password-change failures

Crowley723:


Crowley723:
@hendrik1120 Would you mind giving `authelia/authelia:fix-password-change-error` a try when you have a moment.

hendrik1120:
Yeah, this is much more helpful:
```
time="2025-03-07T09:18:43+01:00" level=error msg="unable to change password for user: 'hendrik': operation failed: LDAP Result Code 50 \"Insufficient Access Rights\": User `authelia` cannot modify the password of user `hendrik`" method=POST path=/api/change-password remote_ip=192.168.5.74
```

james-d-elliott:
We'll need to see before and after to make a decision on this.

hendrik1120:
before:
```
time="2025-03-07T01:23:46+01:00" level=debug msg="Unable to change password for user 'hendrik'" caller="github.com/authelia/authelia/v4/internal/handlers/handler_change_password.go:40 ChangePasswordPOST" error="operation failed" method=POST path=/api/change-password remote_ip=192.168.5.74
```

internal/handlers/handler_change_password.go
```diff
@@ -37,19 +37,21 @@ func ChangePasswordPOST(ctx *middlewares.AutheliaCtx) {
 	}
 
 	if err = ctx.Providers.UserProvider.ChangePassword(username, requestBody.OldPassword, requestBody.NewPassword); err != nil {
-		ctx.Logger.WithError(err).Debugf("Unable to change password for user '%s'", username)
-
 		switch {
 		case errors.Is(err, authentication.ErrIncorrectPassword):
+			ctx.Logger.Debug(fmt.Sprintf("Incorrect password attempt for user %s", username))
```
james-d-elliott, line 42:
```suggestion
			ctx.Logger.WithError(err).WithFields(map[string]any{"username": username}).Debug("Unable to change password for user as their old password was incorrect")
```

internal/handlers/handler_change_password.go
```diff
@@ -37,19 +37,21 @@ func ChangePasswordPOST(ctx *middlewares.AutheliaCtx) {
...
 			ctx.SetJSONError(messageIncorrectPassword)
 			ctx.SetStatusCode(http.StatusUnauthorized)
 		case errors.Is(err, authentication.ErrPasswordWeak):
+			ctx.Logger.Debug(fmt.Sprintf("Weak password attempt for user %s", username))
```
james-d-elliott, line 46:
```suggestion
			ctx.Logger.WIthError(err).WithFields(map[string]any{"username": username}).Debug("Unable to change password for user as their new password was weak or empty")
```

internal/handlers/handler_change_password.go
```diff
@@ -37,19 +37,21 @@ func ChangePasswordPOST(ctx *middlewares.AutheliaCtx) {
...
 			ctx.SetJSONError(messagePasswordWeak)
 			ctx.SetStatusCode(http.StatusBadRequest)
 		case errors.Is(err, authentication.ErrAuthenticationFailed):
+			ctx.Error(fmt.Errorf("unable to change password for user: '%s': %w", username, err), messageOperationFailed)
```
james-d-elliott, line 50:
```suggestion
			ctx.Logger.WIthError(err).WithFields(map[string]any{"username": username}).Error("Unable to change password for user as authentication failed for the user")
```

internal/handlers/handler_change_password.go
```diff
@@ -37,19 +37,21 @@ func ChangePasswordPOST(ctx *middlewares.AutheliaCtx) {
...
 			ctx.SetJSONError(messageOperationFailed)
 			ctx.SetStatusCode(http.StatusUnauthorized)
 		default:
+			ctx.Error(fmt.Errorf("unable to change password for user: '%s': %w", username, err), messageOperationFailed)
```
james-d-elliott, line 54:
```suggestion
			ctx.Error(fmt.Errorf("unable to change password for user: '%s': %w", username, err), messageOperationFailed)
			ctx.Logger.WIthError(err).WithFields(map[string]any{"username": username}).Error("Unable to change password for user for an unknown reason")
```

james-d-elliott (changes requested):
We should avoid the use of ctx.Error() where we can, and always include the `err` in the fields. We should probably also add tests for the log messages including the fields, have a look at the passkey stuff if you're unsure.

Crowley723:
> have a look at the passkey stuff if you're unsure.

I ran into issues doing it the same as the passkey tests you mentioned. The mocks package has a dependency on the authentication package which causes an import loop if I try to use mocks in authentication.

I was able to do it although it doesn't look amazing and doesn't directly test the resulting logs, rather the returned errors.

james-d-elliott (approved):

PR #8960, opened 2025-03-14
Homarr instructions fix

AbelLykens:
These settings actually work 😉

### DIFF
--- a/docs/content/integration/openid-connect/homarr/index.md
+++ b/docs/content/integration/openid-connect/homarr/index.md
@@ -76,14 +76,11 @@ To configure [Homarr] to utilize Authelia as an [OpenID Connect 1.0] Provider:
 1. Include the [Homarr] environment variables for [OpenID Connect 1.0] configuration:
 
 ```env
-AUTH_PROVIDERS=oidc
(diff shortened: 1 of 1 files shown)


james-d-elliott (changes requested):
The one in master has been verified to work as of today. Please articulate the issue with it clearly.

james-d-elliott:
@coderabbitai title

james-d-elliott:
@coderabbitai Listen up, fool! You didn’t read this right! We ain't no optometrists, and even if we were, we ain't got time to hand out prescriptions—especially not for free! That header? It ain't valid. Check this out. https://github.com/authelia/authelia/blob/master/web/.commitlintrc.cjs#L6

james-d-elliott:
@coderabbitai Hey, fool! You havin’ a stroke or somethin’? The line I linked to says the header must be all lowercase; always, not sometimes! Get it right!

Crowley723:
> The one in master has been verified to work as of today.

It should be noted that the current authelia.com version does not match `master` as we are preparing for a release.

james-d-elliott:
Let us know if https://github.com/authelia/authelia/pull/8960#pullrequestreview-2687103262 and/or https://github.com/authelia/authelia/pull/8960#issuecomment-2726247839 (which is no longer the case) are unclear. I don't think this is necessary considering, let us know if you're still having issues with the current docs and exactly what issue you're having.

PR #8079, opened 2024-10-21
feat(oidc): authorization policy network criteria

james-d-elliott:
This adds a network criteria to the Authorization Policies for OpenID Connect 1.0 clients.

PR #8998, opened 2025-03-17
fix matching for multiple oidc rule subjects

lhns:
The subject matching in oidc rules is broken since v4.39.0 by https://github.com/authelia/authelia/pull/8079.
This PR reverts the behavour to match the documentation again.

## Expected Behaviour
```yaml
    authorization_policies:
      my_policy:
        default_policy: 'deny'
        rules:
          - policy: 'two_factor'
            subject:
              - 'group:admin'
              - 'group:my_group'
```
Given this configuration I would expect the user to have access if it is either in the admin or my_group group.
Since the latest update this is not the case anymore and if there is any subject that doesn't match, access is denied.
The required subject list changed from an OR/ANY to a AND/ALL requirement.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the evaluation process for client access checks so that all applicable rules are fully reviewed before a decision is made. This update prevents premature rejections and ensures a more reliable authorization outcome.

- **New Features**
  - Enhanced test coverage by adding a new user, `bob`, to the authorization policy and verifying the expected authorization level for this user.
  - Introduced a comprehensive test for matching subjects against authorization policy rules.


james-d-elliott (changes requested):
Please add a test.

lhns:
Hi, the added user subject "bob" should already cover this issue. Are there some other cases I missed?

Crowley723:
> Hi, the added user subject "bob" should already cover this issue. Are there some other cases I missed?

We want to avoid this issue cropping up in the future. Since the tests were already passing previously (but the issue was there), we want to add new test(s) that will detect regressions.
While your test does have the potential to show regressions, it doesn't test the underlying issue.
Instead of only testing the AuthorizationPolicy that is created, we should also test the MatchesSubject() function directly.

If you want to see what we are looking for you can take a look at the [TestNewClientRequestedAudienceMode](https://github.com/authelia/authelia/blob/d43c119f612c4d4d1d979c90bcf1dc92305aa122/internal/oidc/client_policy_test.go#L143) test.

Some good additional test cases might include:
- A subject that matches none of the rules (should return false)
- A subject that matches one but not all rules (should return true with the fix, would fail with the old logic)
- A subject that matches all rules (should return true)
- A rule with no subjects defined (should match based on network check)


Thanks for catching this issue!

internal/oidc/client_policy_test.go
```diff
@@ -171,3 +176,126 @@
...
+func TestMatchesSubjects(t *testing.T) {
+	lanip, lan, err := net.ParseCIDR("192.168.2.1/24")
+
+	require.NoError(t, err)
+
+	lanip2, _, err := net.ParseCIDR("192.168.3.1/24")
+
+	require.NoError(t, err)
+
+	subjectAnonymous   := authorization.Subject{}
```
autheliabot, line 189: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>File is not properly formatted (gofmt)

internal/oidc/client_policy_test.go
```diff
@@ -171,3 +176,126 @@
...
+				Policy: authorization.OneFactor,
+			},
+			[]authorization.Subject{subjectGroupANet, subjectGroupBNet, subjectGroupABNet},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			matchingSubjects := []authorization.Subject{}
+			for _, subject := range allSubjects {
```
autheliabot, line 293: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>ranges should only be cuddled with assignments used in the iteration (wsl)

internal/oidc/client_policy_test.go
```diff
@@ -171,3 +176,126 @@
...
+				if tc.rule.MatchesSubjects(subject) {
+					matchingSubjects = append(matchingSubjects, subject)
+				}
+			}
+			assert.Equal(t, tc.subjects, matchingSubjects)
```
autheliabot, line 298: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>expressions should not be cuddled with blocks (wsl)

internal/oidc/client_policy_test.go
```diff
@@ -171,3 +176,128 @@
...
+				},
+				Policy: authorization.OneFactor,
+			},
+			[]authorization.Subject{subjectGroupA, subjectGroupB, subjectGroupAB, subjectGroupANet, subjectGroupBNet, subjectGroupABNet, subjectGroupABNet2},
+		},
+		{
+			"ShouldMatchRuleNet",
+			oidc.ClientAuthorizationPolicyRule{
+				Networks: []*net.IPNet{lan},
+				Policy: authorization.OneFactor,
```
autheliabot, line 255: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>File is not properly formatted (gofmt)

james-d-elliott (approved):
LGTM

james-d-elliott:
In go the `_test.go` files are included in all test runs, the functions that accept a `*testing.T` or related test structs will all be tested. Conventional names are Test(Purpose/etc).

Crowley723:
@all-contributors please add @lhns for code, bug, test

PR #9036, opened 2025-03-19
correct jellyfin oidc documentation

ctrenthem159:
Simple one-line change to the configuration instructions for Jellyfin.

The current instructions provide an incorrect redirect_uri. This change keeps that uri (for backwards compatability) and adds the new, working one. There may be a few other URIs that also need to be included depending on the user's specific setup, but I haven't tested any of them on mine, only the `/sso/OID/r/authelia` one (and the original `/sso/OID/redirect/authelia`)

## Summary by CodeRabbit

- **New Features**
  - Enhanced the Jellyfin authentication experience by adding an additional redirect option, providing users with more flexibility during the login process.

### DIFF
--- a/docs/content/integration/openid-connect/jellyfin/index.md
+++ b/docs/content/integration/openid-connect/jellyfin/index.md
@@ -62,6 +62,7 @@ identity_providers:
         pkce_challenge_method: 'S256'
         redirect_uris:
           - 'https://jellyfin.{{< sitevar name="domain" nojs="example.com" >}}/sso/OID/redirect/authelia'
+          - 'https://jellyfin.{{< sitevar name="domain" nojs="example.com" >}}/sso/OID/r/authelia'
         scopes:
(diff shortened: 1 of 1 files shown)


hendrik1120:
There has been some ongoing debate about this, and I have been trying to reproduce the issue for some time. Answering the following questions would help us understand the problem better:

- How are you starting your OIDC flow from jellyfin?
- Could you provide the full URL you are using to initiate the flow? (You can redact your domain for privacy.)
- Where did you find the information that the /sso/OID/r/authelia redirect URL should be used? Did you follow other sources, or only the Authelia integration guide?
- Does this problem only occur on one device or does it not work at all with the documented url?

hendrik1120:
Can be closed if no further response: https://github.com/9p4/jellyfin-plugin-sso/pull/243#issuecomment-2749320553

PR #9077, opened 2025-03-23
docs: correct mixup between implicit and explicit TLS

avesst:
Implicit TLS is when the TLS encryption is handled implicitly at the beginning of the connection.
Explicit TLS is when the connection starts unencrypted, and explicitly is negotiated to upgrade to encrypted.

## Summary by CodeRabbit

- **Documentation**
	- Updated TLS descriptions in the configuration reference to clarify the usage distinctions across various connection schemes.


james-d-elliott:
I understand the motivation here but I'm in two minds as I don't think normal users would think of it this way as the do not understand the technical frame of reference when using that language with TLS. In this instance the scheme itself explicitly communicates the TLS requirement. I'll have a think about it.

avesst:
@james-d-elliott Sorry but I have honestly no idea what you're saying here.

Are you talking about not mentioning implicit/explicit altogether?

james-d-elliott:
I'm saying that the terminology is used in this way in the technical sphere, but I believe most non-technical users will interpret `explicit` from the admin perspective as explicitly requesting a TLS socket. So I understand why this would confuse someone in the technical space. 

So it's true that maybe changing the wording entirely would be more appropriate and clear to more users. Though I need to think about the wording here to make sure it is crystal clear to the majority of users.

james-d-elliott (commented):
LGTM

PR #9241, opened 2025-04-12
fix(configuration): yescrypt not configurable

james-d-elliott:
This fixes an issue where the yescrypt variant can't be configured.

## Summary by CodeRabbit

- **New Features**
  - Added a new option allowing users to select the cryptographic hashing variant ("scrypt" or "yescrypt") for generating hash digests.
  - Enforced validation rules ensuring variant-specific parameter constraints are met.

- **Documentation**
  - Updated command-line and configuration guides to include details of the new variant option, with clear defaults and accepted values.
  - Enhanced password hashing documentation to clearly distinguish between the supported variant options.


atalii:
Arguably yescrypt should be the default; it's much more resistant to GPU-based cracking and so forth, as well as being the default for Fedora, Ubuntu, and some other big distros.

It's also bugging me a little that we're using \$y\$ for yescrypt and \$scrypt\$ for scrypt (the corresponding identifier would be \$7\$), but that's mostly an aesthetic nitpick on my part.

Thanks a ton for working on this!

james-d-elliott:
There's no way we can make it the default. Even if we could I'd be cautious about jumping into something like this without some significant analysis. The `scrypt` KDF itself has a significant disadvantage where the N parameter controls both CPU and memory cost which also affects yescrypt by the look of it. This issue is not the same for `argon2id`. 

Also as far as the identifier most implementations use `scrypt`, I believe it's only the crypt implementation on linux that doesn't. In any case we can't change that now other than to support decoding type 7.

atalii:
Sorry for the late response - I replied yesterday via email but it doesn't look like it went through.

I meant that yescrypt should likely be the default variant when picking between yescrypt and scrypt, not that it ought to supersede  argon2id - sorry for the confusion there! Supporting type 7 would be a welcome addition if you think it's worth the hassle. By the way, thanks for making the yescrypt params compatible with how crypt() does things, that makes things much easier for me :)

james-d-elliott:
> I meant that yescrypt should likely be the default variant when picking between yescrypt and scrypt

This can't be done though without a breaking change.

PR #9257, opened 2025-04-15
docs: update minio docs

Crowley723:
This attempts to make the relationship between Authelia groups and MinIO policies more clear.

It should also be noted that with the addition of claims policies and the additional user attributes, it should be possible to specify MinIO policies using a custom user attribute and claim that is provided to MinIO.

## Summary by CodeRabbit

- **Documentation**
  - Updated instructions for configuring MinIO with Authelia as an OpenID Connect provider, including enabling the "Claim User Info" option and clarifying how to associate user groups with policies. Emphasized the need for exact group and policy name matching and specified limitations on available default policies.


james-d-elliott (approved):
LGTM

PR #9327, opened 2025-04-23
docs: refactor ldap documentation

hendrik1120:
Proposal to move the LDAP documentation from the reference section to the integration section and split it into individual providers.

## Summary by CodeRabbit

- **Documentation**
  - Added new integration guides for Active Directory, FreeIPA, GLAuth, LLDAP, and RFC2307bis LDAP backends.
  - Revised LDAP integration introduction to provide a conceptual overview, binding methods, supported implementations, group search modes, and filter replacements.
  - Updated LDAP integration documentation structure and metadata for improved organization and navigation.
  - Updated internal documentation links to point to the new LDAP integration guide location.
  - Removed the previous comprehensive LDAP reference guide and replaced it with focused, implementation-specific documentation.


james-d-elliott (dismissed):
The application section for each app should be removed and just be placed in the introduction as a generic guide on a user an the relevant general permissions they should have i.e. read user attributes, change password if the ability to reset passwords is enabled, create user/group/container objects if they want to support future versions which support user management. 

Users should refer to the user guide and support channels of the relevant application for how to achieve these goals.

Also the weight should be reverted I think, was it an intentional change?

hendrik1120:
> Also the weight should be reverted I think, was it an intentional change?

It was an intentional change, otherwise it will be below the oidc integration guides and basically invisible on all screen sizes without scrolling to the very bottom

james-d-elliott:
I think if anything moves it's trusted header sso (to the bottom under LDAP) based on importance. We can realistically adjust the individual application guides for OIDC to be in a collapsed menu in a separate PR too.

Considering integration of LDAP backends is `set the implementation` it'd be rare users would have to do more than look at the configuration guide, whereas elements like proxies, k8s, docker,  and OIDC will nearly always require someone to look at a guide until we do a similar profile-based configuration, and even then they'll need to know the specifics about configuring the app itself. The trusted header SSO is okay but it's rarely used.

PR #9339, opened 2025-04-25
amend the wrong userinfo_signed_response_alg for the correct id_token…

xhemp:
userinfo_signed_response_alg is the wrong configuration for Home Assistant, id_token_signed_response_alg is the right one.
Spent some time troubleshooting this and thought would be good to get a PR to amend this on Authelia's docs.

## Summary by CodeRabbit

- **Documentation**
  - Updated the Home Assistant integration guide to use the correct OpenID Connect parameter for configuring the ID token signature algorithm.

### DIFF
--- a/docs/content/integration/openid-connect/home-assistant/index.md
+++ b/docs/content/integration/openid-connect/home-assistant/index.md
@@ -69,7 +69,7 @@ identity_providers:
           - 'openid'
           - 'profile'
           - 'groups'
-        userinfo_signed_response_alg: 'none'
+        id_token_signed_response_alg: 'none'
 ```
 
 ### Application


james-d-elliott (changes requested):
Please show the error that occurs without `id_token_signed_response_alg` set to none.

james-d-elliott:
So why are you setting it to a different value than what he set it to?

xhemp:
My one is actually set to RS256, just saw that, what I really wanted to change was from userinfo_signed_response_alg to id_token_signed_response_alg. I could only log in after I changed that.

james-d-elliott:
Alright, well the authelia configs are the same then. `userinfo_signed_response_alg` defaults to `none` and `id_token_signed_response_alg` defaults to `RS256`. We only include `userinfo_signed_response_alg` set to `none` for the sole purpose of the fact that has historically been changed a lot by third parties.

xhemp:
Ran a bit more tests here and indeed, If I try with exactly the same config on Authelia's page, it works fine so no need for this one! Thanks!

christiaangoossens:
> Alright, well the authelia configs are the same then. `userinfo_signed_response_alg` defaults to `none` and `id_token_signed_response_alg` defaults to `RS256`. We only include `userinfo_signed_response_alg` set to `none` for the sole purpose of the fact that has historically been changed a lot by third parties.
> 
> Without the error we probably can't action this.

If I'm correct this all started when the original doc PR for Authelia on the HA integration accidentally set the userinfo endpoint to be signed.

Up to then it seemed the defaults just worked and I would expect the id_token to be `RS256`.

For future readers, DO NOT set the id token to be unsigned, as this is unsafe. The defaults are fine.

PR #9359, opened 2025-04-28
feat: adding the token itself as a template variable for template use

centerorbit:
We had an instance when making an override template that we wanted access to the token directly. This exposes that.

## Summary by CodeRabbit

- **New Features**
  - Identity verification emails now include a signed token for improved verification processes.

### DIFF
--- a/internal/middlewares/identity_verification.go
+++ b/internal/middlewares/identity_verification.go
@@ -97,6 +97,7 @@ func IdentityVerificationStart(args IdentityVerificationStartArgs, delayFunc Tim
 
 		data := templates.EmailIdentityVerificationJWTValues{
 			Title:              args.MailTitle,
+			Token:              signedToken,
 			LinkURL:            linkURL.String(),
 			LinkText:           args.MailButtonContent,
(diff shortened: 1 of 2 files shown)


james-d-elliott:
Thanks for this. Please clearly explain the use case.

centerorbit:
> Thanks for this. Please clearly explain the use case.

Hey James!

Our use-case is this that we needed more control over where the token is being used in the email template.

With our case, Authelia is an independent backend service, not publically exposed at all, and so these tokens must go to our web service to then get forwarded to our Authelia internally via APIs.

This change also allows us to use custom app url schemes for these emails to recieve the reset request. Particularly useful for apps that are primarily web-backed 'hybrid' apps. We'll want our templates to go to `app://reset?token=<token>` for example so that the email client can invoke the "native" app instead of the browser.

james-d-elliott:
I think #9468 will cover this use case sufficiently.

centerorbit:
> I think #9468 will cover this use case sufficiently.

I believe you're right! And I see it recently was merged back. I'll close out this ticket and we'll upgrade to the newest templating version.

Thank you!

PR #9586, opened 2025-05-27
docs: fix domain_regex example

cschug:
The second example regarding the `domain_regex` option in the access control documentation wants to demonstrate the combined use of `domain` and `domain_regex` at the same time in a single rule but rather lists `domain` and `domain_regex` as separate rules (with the first one actually being invalid because of a missing `policy` option).

## Summary by CodeRabbit

- **Documentation**
  - Corrected YAML indentation in the access control configuration guide to improve clarity and accuracy. No changes to the actual configuration or policies.


james-d-elliott (changes requested):
Thanks for choosing to contribute, however no domain will match all of these criteria as you can't both be `protected.example.com` and match the regex pattern. It's just missing a policy.

PR #9698, opened 2025-06-21
docs: add bookstack integration considerations for authelia v4.39

LenardHess:
This PR adds configurations needed with Authelia v4.39, specifically the configuration of a custom `claim_policy`.

## Summary by CodeRabbit

- **Documentation**
  - Updated integration guide for BookStack to support Authelia version 4.39.4.
  - Added instructions for configuring claims policies to ensure the email claim is included for BookStack compatibility.
  - Introduced a section explaining changes in Authelia v4.39 affecting ID Token claims.


james-d-elliott (changes requested):
Thanks for choosing to contribute. See another example of how to document this client behavior that defies the spec: https://github.com/authelia/authelia/blob/8173ba4f69ac90c013f7dc6dc0d5cb0d446a01a0/docs/content/integration/openid-connect/romm/index.md?plain=1#L47

LenardHess:
Changed the description text to use the template.

I left in the configuration snippets - is that alright with you?

Also - the Claims Conformance shared section does not refer to any Authelia version - would you accept a PR that adds relevant versions where Authelia behavior changed to that shared section - similar to my first draft of this PR?

james-d-elliott (approved):
LGTM

james-d-elliott:
> Also - the Claims Conformance shared section does not refer to any Authelia version - would you accept a PR that adds relevant versions where Authelia behavior changed to that shared section - similar to my first draft of this PR?

This is a workaround and/or escape hatch for a categorically incorrect behavior for bookstack which they should fix. They are not requesting the claims be present in the ID Token, they're just asking for the Access Token to have the required access to obtain them, and then not using that Access Token to retrieve them. Additionally even though they are not asking for the claims to be present in the ID Token, they're checking the ID Token for them even though there is no reasonable expectation that they should be there.

TL;DR; Bookstack **_does not support_** OpenID Connect 1.0 even though they claim to.

The linked reference guide includes this information as well, I don't think we need to duplicate it. But take a closer look once it's merged and if you feel the clarity is lacking then feel free to PR it.

james-d-elliott:
Adjusted as necessary and clarified the bugs on their end: https://www.authelia.com/integration/openid-connect/bookstack/

LenardHess:
I see you removed the old authelia version on the bookstack integration guide - in that case, the integration guide itself is clear.

My line of thinking on adding an authelia version hint to the claims bug banner was just to ease authelia migrations. It would admittedly be redundant information on top of the changelog itself - but (from my sample size of 1) piecing it all together without being versed in the topic took a bit. Hopefully the updated integration guide means anyone else doing this migration path will have an easier time.

PR #9711, opened 2025-06-25
docs(oidc): Add Headscale documentation

fredrikekre:


docs/content/integration/openid-connect/headscale/index.md
```diff
@@ -0,0 +1,116 @@
...
+          - 'email'
+          - 'profile'
+          - 'groups'
+        response_types:
+          - 'code'
+        grant_types:
+          - 'authorization_code'
+        access_token_signed_response_alg: 'none'
+        userinfo_signed_response_alg: 'none'
+        token_endpoint_auth_method: 'client_secret_basic'
```
fredrikekre, line 78: I included these for consistency with other integrations (https://github.com/authelia/authelia/pull/9672). Out of curiosity, is there a reason these options are explicitly included although they simply use default values (and in some of the cases the example configuration warns against setting them)?

james-d-elliott: They are explicitly included as they are options that third parties often change or users want to change.

docs/content/integration/openid-connect/headscale/index.md
```diff
@@ -0,0 +1,116 @@
...
+- __Client Secret:__ `insecure_secret`
+
+Some of the values presented in this guide can automatically be replaced with documentation variables.
+
+{{< sitevar-preferences >}}
+
+## Configuration
+
+### Authelia
+
```
Crowley723, line 46:
```suggestion

{{% oidc-conformance-claims %}}

```

docs/content/integration/openid-connect/headscale/index.md
```diff
@@ -0,0 +1,116 @@
...
+operate with the application example:
+
+```yaml {title="configuration.yml"}
+identity_providers:
+  oidc:
+    ## The other portions of the mandatory OpenID Connect 1.0 configuration go here.
+    ## See: https://www.authelia.com/c/oidc
+    claims_policies:
+      headscale:
+        id_token: ['groups', 'email', 'email_verified', 'alt_emails', 'preferred_username', 'name']
```
Crowley723:
```suggestion
```

We have decided to take a pretty strict line on the use of `claims_policies` in our docs, since the only reason its required is that client applications aren't doing the bare minimum as an OIDC client.

docs/content/integration/openid-connect/headscale/index.md
```diff
@@ -0,0 +1,116 @@
...
+      - client_id: 'headscale'
+        client_name: 'Headscale'
+        client_secret: '$pbkdf2-sha512$310000$c8p78n7pUMln0jzvd4aK4Q$JNRBzwAo0ek5qKn50cFzzvE9RXV88h1wJn5KGiHrD0YKtZaR/nCb2CJPOsKaPK0hjf.9yHxzQGZziziccp6Yng'  # The digest of 'insecure_secret'.
+        public: false
+        authorization_policy: 'two_factor'
+        require_pkce: true
+        pkce_challenge_method: 'S256'
+        redirect_uris:
+          - 'https://headscale.{{< sitevar name="domain" nojs="example.com" >}}/oidc/callback'
+        claims_policy: 'headscale'
```
Crowley723:
```suggestion
```

Crowley723 (changes requested):
Just a few changes.

Thanks for choosing to contribute!

docs/content/integration/openid-connect/headscale/index.md
fredrikekre: Okay that's fair, but this is *the* thing I wished was here. Took me half a day to figure out. The rest is pretty easy to extrapolate from all the other integrations. I added this (and in fact created the PR) so that someone else doesn't have to waste the time I did. Perhaps the `oidc-conformance-claims` you suggested above is enough, although not as explicit.

fredrikekre:
Thanks for the comments, updated the PR.

docs/content/integration/openid-connect/headscale/index.md
james-d-elliott: We appreciate the PR a lot, this is not a criticism on you. Let me explain.

The issue is that Headscale **_does not_** actually support OpenID Connect 1.0 but claims to. The `id_token` key of the claims policy is an escape hatch for these buggy clients, reasonably this is also something we may choose to remove in the future as this option works around a failure to follow the most basic spec requirements. The absolute bare minimum required by them under the spec regarding claims is that they are required to check the UserInfo endpoint for the claims and assume they are not in the ID Token except when they expressly request the claims via the claims parameter (they're using the scope parameter which does not affect the ID Token; it affects the Access Token) or via the `response_type=id_token` i.e. Implicit Flow without an Access Token.

A particular advantage of keeping this specific element separate is the ease of removal once these clients fix the bug. This is incredibly appealing to the maintainability of all of these integrations as it's quite a complex task even though the community mostly does it.

It might be worth raising a bug report with Headscale since this is an undeniable bug on their end. If you don't I will be making a very detailed report to multiple providers in the coming weeks.

Also I've been thinking about this a lot and just updated the layout to be clearer. I'll merge this and update the relevant parts. See it here: https://www.authelia.com/integration/openid-connect/opkssh/#authelia

james-d-elliott (approved):
LGTM

james-d-elliott:
In case you missed my reply: https://github.com/authelia/authelia/pull/9711#discussion_r2168554997

docs/content/integration/openid-connect/headscale/index.md
fredrikekre: Yea I understand. I think this is sufficiently tracked by https://github.com/juanfont/headscale/issues/2655 on the Headscale side. Thanks!

PR #9803, opened 2025-07-03
Allow for unencrypted smtp auth when config option `disable_require_tls` is set to true

bishtawi:
Hi,

I want to be able to send emails via an unencrypted smtp connection but with auth (smtp username and password). I know this is considered insecure as it leaks credentials over an unencrypted connection but both Authelia and the email service are hosted on the same machine. Also, if Authelia is going to support a config option to disable smtp TLS, I think you should fully support that mode of operation.

At first I tried setting `disable_require_tls` config option to `true` but when I attempt to send a password reset email I saw an error in the logs:
```
level=error msg="notifier: smtp: failed to dial connection: SMTP AUTH failed: unencrypted connection" method=POST path=/api/reset-password/identity/start remote_ip=192.168.x.x
```

Did a little digging and saw that for SMTP PlainAuth and LoginAuth its hardcoded to set the argument `allowUnenc` to `false`.

While this is great for security, there are situations where you still want SMTP AUTH but dont need TLS. The `gomail` dependency you are using explicitly allows unencrypted SMTP auth for connections to localhost which makes sense.

In my case I am selfhosting Authelia along with an email service on the same machine. My TLS terminates at the reverse proxy so service to service communication within the same machine is all done without TLS. Both services are dockerized and thus do not share the same localhost so the explicit exception that `gomail` makes for localhost does not help me.

I need the Authelia `disable_require_tls` option to also work for SMTP PlainAuth and LoginAuth by passing it through to `gomail`.

`disable_require_tls` defaults to false so if someone is setting it to true then they are accepting the decreased security with sending smtp creds over an unencrypted connection.

## Summary by CodeRabbit

* **New Features**
  * Added support for conditionally allowing unencrypted SMTP authentication based on configuration settings. This provides greater flexibility for users managing email notification security requirements.
* **Documentation**
  * Enhanced warning in SMTP configuration about the risks of disabling TLS, emphasizing the potential exposure of sensitive data during email notifications.
  * Added a prominent security warning highlighting the critical risks of disabling TLS enforcement for SMTP notifications.


james-d-elliott (approved):
LGTM

PR #9807, opened 2025-07-04
docs: complete i18n on roadmap

Crowley723:


Crowley723:
@james-d-elliott I wasn't sure if i18n should be moved to completed since it has an ongoing maintenance task (ongoing translation) but I put it there because it doesn't require any additional work.

james-d-elliott (approved):

PR #9390, opened 2025-05-03
docs(oidc): fix immich example config

Bhambya:
Immich [v1.132.3](https://github.com/immich-app/immich/releases/tag/v1.132.3) fixes an issue where the mobile app cannot log in with an instance using Authelia for OAuth. They suggest setting 
```
token_endpoint_auth_method: "client_secret_post"
```

## Summary by CodeRabbit

- **Documentation**
  - Updated the OpenID Connect integration guide for Immich to reflect a change in the recommended authentication method for the token endpoint.
  - Updated tested versions of Authelia and Immich in the integration documentation.

PR #9887, opened 2025-07-16
Update oauth docs for latest immich

mikhail5555:
A new option `TOKEN_ENDPOINT_AUTH_METHOD` is now available which is set to `client_secret_post` by default which leads to errors

## Summary by CodeRabbit

* **Documentation**
  * Updated integration guide with newer tested versions of Authelia and Immich.
  * Revised OAuth client configuration instructions to use uppercase parameter names.
  * Changed `AUTO REGISTER` parameter guidance to specify using the string "Enable".
  * Adjusted formatting of OAuth client parameters for clarity and consistency.

### DIFF
--- a/docs/content/integration/openid-connect/immich/index.md
+++ b/docs/content/integration/openid-connect/immich/index.md
@@ -91,7 +91,6 @@ To configure [Immich] to utilize Authelia as an [OpenID Connect 1.0] Provider, u
     - ISSUER_URL: `https://{{< sitevar name="subdomain-authelia" nojs="auth" >}}.{{< sitevar name="domain" nojs="example.com" >}}/.well-known/openid-configuration`.
     - CLIENT_ID: `immich`.
     - CLIENT_SECRET: `insecure_secret`.
-    - TOKEN_ENDPOINT_AUTH_METHOD: `client_secret_basic`
     - SCOPE: `openid profile email`.
     - BUTTON TEXT: `Login with Authelia`.
     - AUTO REGISTER: `Enable` (if desired).


james-d-elliott:
Can you take a screenshot of the GUI settings using the immich.example.com and auth.example.com domains?

mikhail5555:
> Can you take a screenshot of the GUI settings using the immich.example.com and auth.example.com domains?

You beat me to it, i was JUST about to add it to the PR. Is this what you mean, or do you want a screenshot of something else?
image

mikhail5555:
> Can you take a screenshot of the GUI settings using the immich.example.com and auth.example.com domains?

Ah i think i understand what you mean, give me a second
image


Is this what you want? (to possibly add it to the documentation?)

james-d-elliott:
Yep bet you've got it now, we use `example.com` as default example values so that's what we include in the screenshots, it's not easy to do substitution in a way that most users will be able to do PR's so I've avoided performing the image magic to make that occur (you can technically fairly easily render text within an image at exact locations).

~~Can you update the other value instructions to match the GUI?~~ just noticed you did this, awesome!

james-d-elliott:
Also just realized this means it now supports `client_secret_post` which I prefer as it has a marginal security benefit over `client_secret_basic` (headers are much easier to capture and much more often logged by enterprise firewalls) and also is less often butchered by third parties which do not encode the header correctly.

If it's not too much trouble could you update it to use `client_secret_post` both in the authelia config, the immich config, then test it? If that works can you update those two in the PR and the screenshot.

mikhail5555:
> Also just realized this means it now supports `client_secret_post` which I prefer as it has a marginal security benefit over `client_secret_basic` (headers are much easier to capture and much more often logged by enterprise firewalls) and also is less often butchered by third parties which do not encode the header correctly.
> 
> If it's not too much trouble could you update it to use `client_secret_post` both in the authelia config, the immich config, then test it? If that works can you update those two in the PR and the screenshot.

Will do tiny bit later, i tried it with 'post' first, but it didnt seem authelia supported it without extra changes (I left `token_endpoint_auth_method` on default), So i guess it would also need to explicitly add it to authelia config.
```
time="2025-07-16T08:54:28+02:00" level=error msg="Access Request failed with error: Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method). The request was determined to be using 'token_endpoint_auth_method' method 'client_secret_post', however the OAuth 2.0 client registration does not allow this method. The registered client with id 'immich' is configured to only support 'token_endpoint_auth_method' method 'client_secret_basic'. Either the Authorization Server client registration will need to have the 'token_endpoint_auth_method' updated to 'client_secret_post' or the Relying Party will need to be configured to use 'client_secret_basic'." method=POST path=/api/oidc/token remote_ip=10.95.0.9
```

mikhail5555:
> Also just realized this means it now supports `client_secret_post` which I prefer as it has a marginal security benefit over `client_secret_basic` (headers are much easier to capture and much more often logged by enterprise firewalls) and also is less often butchered by third parties which do not encode the header correctly.
> 
> If it's not too much trouble could you update it to use `client_secret_post` both in the authelia config, the immich config, then test it? If that works can you update those two in the PR and the screenshot.

Seems in #9390 the example config already got changed to use `client_secret_post`. So i updated it to just leave it 'default' in the UI:
image

edit: I guess this PR now doesnt really add anything anymore haha

hendrik1120:
It's more likely that the capitalization changes are a bug, because this is only place in the frontend with ENV like naming.

mikhail5555:
> It's more likely that the capitalization changes are a bug, because this is only place in the frontend with ENV like naming.

I can likely close this whole PR all together, since it seems i am not actually changing anything of use with it.

mikhail5555:
Since i am not really adding anything of value anymore, this pr can be considered void. thank you guys for your time

hendrik1120:
Thanks for pointing out the changes, I have opened an issue at Immich, so the docs on their side get updated and the naming gets sorted out: https://github.com/immich-app/immich/issues/19959

PR #9897, opened 2025-07-17
Fix existing user linking in example Jellyfin openid configuration

Steve-Brule:
Per jellyfin-plugin-sso documentation:

"Note that you should add both "/r/" and "/redirect/" paths to your SSO provider's configuration!"

It seems that without the /r/ path in redirect_uris, linking existing user with authelia fails.

### DIFF
--- a/docs/content/integration/openid-connect/jellyfin/index.md
+++ b/docs/content/integration/openid-connect/jellyfin/index.md
@@ -77,6 +77,7 @@ identity_providers:
         pkce_challenge_method: 'S256'
         redirect_uris:
           - 'https://jellyfin.{{< sitevar name="domain" nojs="example.com" >}}/sso/OID/redirect/authelia'
+          - 'https://jellyfin.{{< sitevar name="domain" nojs="example.com" >}}/sso/OID/r/authelia'
         scopes:
           - 'openid'
           - 'profile'


james-d-elliott:
Thanks for the PR but this needs to be resolved by the plugin author as per https://github.com/9p4/jellyfin-plugin-sso/pull/243 and https://github.com/authelia/authelia/pull/9036. It appears the usage of this URI is deprecated on their end (see https://github.com/9p4/jellyfin-plugin-sso/commit/defab9c4e8558d347314bb9a8b74ce673c5dede4).

They also document the /redirect/ variant 7 times more than the older variant (which is not properly documented at all).

Steve-Brule:
In that case, perhaps a note about this would be nice to have in the docs until the issue is fixed.

james-d-elliott:
From what I understand the only time the `/r/` URI is used is with deprecated setups that don't use current documentation. In particular if the `/start/` URL's are used to initiate the flows the `/redirect/` URI's will always be used.

My current leaning is to freeze the guide with a YMMV-style warning or remove it entirely. The plugin is in "very early alpha" and the developer doesn't seem to have time to maintain it. Not that I'm being critical of the maintainer life happens, I'm just not looking to adopt yet another unmaintained project; and this plugin is starting to become a maintenance burden.

Crowley723:
> YMMV-style warning

I agree with this, jellyfin oidc plugins have had multiple discussions/prs opened directly related to the uri handling for oidc. Until such a time as the upstream docs are clear on the matter I don't think we should expend more energy trying to support it.

hendrik1120:
Agreed, this is really getting too out of hand. There is too much old documentation flowing around using the old paths, which brings this issue up again and again. I doubt that removing it from their official docs will help much, but I created a PR for that.

james-d-elliott:
Thanks a lot for choosing to contribute @Steve-Brule. While I'm closing this but we're open to a [Discussion](https://github.com/authelia/authelia/discussions/new?category=troubleshooting) to troubleshoot the reasons why you're having this issue, and we generally appreciate your choice to contribute regardless and would not want to discourage that. 

I suspect one of the URL's configured in Jellyfin are not the correct URL's and this is what caused your experience.

@all-contributors please add @Steve-Brule  for docs

PR #9985, opened 2025-07-28
docs: grafana oidc integration

vehagn:
Remove claims hydration bug. Grafana is able to fetch user details from the UserInfo endpoint.

I've tested Grafana integration in my Homelab where I don't need to specify the _escape hatch_ with extra claims in the ID Token. Grafana appears to be fetching name and email from the UserInfo endpoint.

[Authelia configuration](https://github.com/vehagn/homelab/blob/285d9b075e72e8f8985ea826fba1ab956ba112e6/k8s/infra/auth/authelia/values.yaml#L161-L177)
[Grafana configuration](https://github.com/vehagn/homelab/blob/285d9b075e72e8f8985ea826fba1ab956ba112e6/k8s/infra/monitoring/prometheus-stack/values.yaml#L48-L66)

### DIFF
--- a/docs/content/integration/openid-connect/clients/grafana/index.md
+++ b/docs/content/integration/openid-connect/clients/grafana/index.md
@@ -3,6 +3,7 @@ title: "Grafana"
 description: "Integrating Grafana with the Authelia OpenID Connect 1.0 Provider."
 summary: ""
 date: 2024-03-14T06:00:14+11:00
+lastmod: 2025-07-28T23:00:00+02:00
 draft: false
 images: []
 weight: 620
@@ -23,11 +24,11 @@ seo:
 ## Tested Versions
 
 - [Authelia]
-  - [v4.38.17](https://github.com/authelia/authelia/releases/tag/v4.38.17)
+  - [v4.39.5](https://github.com/authelia/authelia/releases/tag/v4.39.5)
 - [Grafana]
-  - [v11.4.0](https://github.com/grafana/grafana/releases/tag/v11.4.0)
+  - [v12.0.2](https://github.com/grafana/grafana/releases/tag/v12.0.2)
 
-{{% oidc-common bugs="claims-hydration" %}}
+{{% oidc-common %}}
 
 ### Assumptions
 
@@ -45,12 +46,6 @@ Some of the values presented in this guide can automatically be replaced with do
 ## Configuration
 
 ### Authelia
-
-{{< callout context="caution" title="Important Note" icon="outline/alert-triangle" >}}
-At the time of this writing this third party client has a bug and does not support [OpenID Connect 1.0](https://openid.net/specs/openid-connect-core-1_0.html). This
-configuration will likely require configuration of an escape hatch to work around the bug on their end. See
-[Configuration Escape Hatch](#configuration-escape-hatch) for details.
-{{< /callout >}}
 
 The following YAML configuration is an example __Authelia__ [client configuration] for use with [Grafana] which will
 operate with the application example:
@@ -79,18 +74,13 @@ identity_providers:
           - 'code'
         grant_types:
           - 'authorization_code'
-        access_token_signed_response_alg: 'none'
         userinfo_signed_response_alg: 'none'
         token_endpoint_auth_method: 'client_secret_basic'
 ```
 
-#### Configuration Escape Hatch
-
-{{% oidc-escape-hatch-claims-hydration client_id="grafana" claims="email,name,groups,preferred_username" %}}
-
 ### Application
 
-To configure [Grafana] there are two methods, using the [Configuration File](#configuration-file), or using
+To configure [Grafana], there are two methods, using the [Configuration File](#configuration-file), or using
 [Environment Variables](#environment-variables).
 
 #### Configuration File
@@ -120,6 +110,7 @@ login_attribute_path = preferred_username
 groups_attribute_path = groups
 name_attribute_path = name
 use_pkce = true
+auth_style = InHeader
 role_attribute_path =
 ```
 
@@ -145,6 +136,7 @@ GF_AUTH_GENERIC_OAUTH_LOGIN_ATTRIBUTE_PATH=preferred_username
 GF_AUTH_GENERIC_OAUTH_GROUPS_ATTRIBUTE_PATH=groups
 GF_AUTH_GENERIC_OAUTH_NAME_ATTRIBUTE_PATH=name
 GF_AUTH_GENERIC_OAUTH_USE_PKCE=true
+GF_AUTH_GENERIC_OAUTH_AUTH_STYLE=InHeader
 GF_AUTH_GENERIC_OAUTH_ROLE_ATTRIBUTE_PATH=
 ```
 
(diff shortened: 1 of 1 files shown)


docs/content/integration/openid-connect/clients/grafana/index.md
```diff
@@ -120,6 +110,7 @@ login_attribute_path = preferred_username
 groups_attribute_path = groups
 name_attribute_path = name
 use_pkce = true
+auth_style = InHeader
 role_attribute_path =
```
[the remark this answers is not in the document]
vehagn: **No, you stop taking hallucinogens and chill!**
Grafana capitalises it as `InHeader` in their documentation: https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/generic-oauth/#configuration-options

docs/content/integration/openid-connect/clients/grafana/index.md
```diff
@@ -79,18 +74,13 @@ identity_providers:
           - 'code'
         grant_types:
           - 'authorization_code'
-        access_token_signed_response_alg: 'none'
```
vehagn: I'm able to use `access_token_signed_response_alg: ES256` (and `id_token_signed_response_alg: ES256`) given

```yaml
  identity_providers:
    oidc:
      jwks:
        - key_id: default
          algorithm: RS256
          use: sig
          key: { path: /secrets/jwk-rsa/tls.key }
          certificate_chain: { path: /secrets/jwk-rsa/tls.crt }
        - key_id: ecdsa256
          algorithm: ES256
          use: sig
          key: { path: /secrets/jwk-ecdsa/tls.key }
          certificate_chain: { path: /secrets/jwk-ecdsa/tls.crt }
```

So I assume Grafana handles the default RS256 signing algorithm as well.

vehagn: I re-read the section on [`access_token_signed_response_alg`](https://www.authelia.com/configuration/identity-providers/openid-connect/clients/#access_token_signed_response_alg) and I see that the default value is `none`. I (incorrectly?) assumed it's safer to sign the Access Token as well.
I have no qualms about changing this back.

james-d-elliott: Yeah it's left as it's a common option people want to change or clients incorrectly think needs to be changed. In reality the Access Token format is none of their business or concern.

vehagn: Should I just leave it as-is then?

Do you also want me to update the commit-comment to e.g. `docs(oidc): remove grafana claims-hydration bug disclamer`?

I see Coderabbit didn't enforce this

> Learnt from: james-d-elliott
PR: authelia/authelia#0
:0-0
Timestamp: 2025-03-15T23:14:37.691Z
Learning: PR titles for the authelia/authelia repository must follow the commitlint rules defined in web/.commitlintrc.cjs, which require all text to be lowercase. For documentation changes related to OpenID Connect, the format should be "docs(oidc): [description of changes]".

hendrik1120:
The integration guide requests the `groups` scope, if I try to use said scope to map the grafana roles without a claims policy explicitly placing it in the ID-Token, my admin role gets removed. I have a very similar `role_attribute_path` to yours:
```
contains(groups[*], 'admins') && 'GrafanaAdmin' || contains(groups[*], 'editor') && 'Editor' || 'Viewer'
```
You are using a custom `grafana` scope. In theory, it shouldn't make a difference, but please make sure everything works with what is configured in the guide.

Edit: This feature is also explicitly mentioned https://www.authelia.com/integration/openid-connect/clients/grafana/#role-attribute-path

vehagn:
@hendrik1120 Thank you for properly testing my proposed change! You are correct in arresting me for not properly checking with the groups claim. I assumed that it would behave in a similar fashion to my custom claim. Testing ([commit](https://github.com/vehagn/homelab/commit/d63982ddd00d068413b155d5efac129a8173ee2d)) with the `groups` claim I unfortunately get the same results as you.

It appears that Grafana (or the library they're relying on) is treating the `groups` claim in a special way. You got me to double check the presence of my custom `grafana` claim the _ID Token_ (using [these](https://gitlab.com/vehagn/blog/-/blob/main/content/articles/2025/06/authelia-oidc/resources/scripts/fetch_tokens.sh) [two](https://gitlab.com/vehagn/blog/-/blob/main/content/articles/2025/06/authelia-oidc/resources/scripts/inspect_tokens.sh) bash scripts). The custom claim is not present in the _ID Token_!

Auth URL:

```
https://authelia.stonegarden.dev/api/oidc/authorization?
response_type=code&
client_id=grafana&
redirect_uri=http://localhost:8080&
scope=openid profile offline_access groups grafana&
code_challenge_methis=...
```

ID Token payload:

```json
{
  "amr": [
    "pwd",
    "kba",
    "pop",
    "swk",
    "user",
    "mfa"
  ],
  "at_hash": "PpsUD910jerm_Sjf1cNN3w",
  "aud": [
    "grafana"
  ],
  "auth_time": 1753805632,
  "azp": "grafana",
  "exp": 1753809245,
  "iat": 1753805645,
  "iss": "https://authelia.stonegarden.dev",
  "jti": "<>",
  "sub": "<>"
}
```

Access token introspection:

```json
{
  "active": true,
  "client_id": "grafana",
  "exp": 1753809245,
  "iat": 1753805644,
  "scope": "openid profile offline_access groups grafana",
  "sub": "<>",
  "username": "<>"
}
```

UserInfo endpoint:

```json
{
  "grafana": [
    "grafana_admin"
  ],
  "groups": [
    "k8s:cluster_admin",
    "lldap_admin"
  ],
  "name": "Vegard",
  "preferred_username": "<>",
  "rat": 1753805625,
  "sub": "<>",
  "updated_at": 1753805646
}
```

I simply assumed that since the `grafana` claim "behaves" similar to the "group" claim (not present in _ID Token_, available as a list of strings under UserInfo endpoint), that they would be the same! This only fuels my annoyance with half-assed implementations of the OAuth/OIDC spec! Treating the `grafana` and `groups` claims the same would surely be easier to implement?

james-d-elliott:
Does this mean the adjustments work as is or needs something more?

hendrik1120:
Grafana does not seem to fetch or use the `groups` claim from the UserInfo Endpoint when mapping a role to a group using the `role-attribute-path`.
The usage of this setting has been explicitly documented in the guide, this PR as is will break the role assignment.
There are multiple ways to fix or workaround this issue.

Option 1: Use a custom `grafana` scope with a custom role claim directly from LDAP or file.
Option 2: Use a custom `grafana` scope with a custom role claim derived using CEL from the groups attribute
Option 3: Escape hatch using a claims policy to force the `groups` claim to the ID-Token

I am not sure if this is a bug on the grafana side, or if the `groups` claim is being treated differently on purpose, in which case options 1&2 are clearly better as it won't be fixed.

Whatever you think is best, I can create a follow-up PR or update the existing guide

vehagn:
@hendrik1120 I'd be interested in option 2 just because I haven't seen that one before. I planned to take a look on the Grafana source code (and hopefully fix it there), but I haven't had the time yet.

james-d-elliott:
Not entirely sure the state of this. Is this realistically an incorrect change or?

vehagn:
As it stands now it's an incorrect change I think. Grafana apparently (incorrectly) treats the group claim as a special case.

nightah:
Just a final confirmation, can this be closed?
Has anybody lodged an issue upstream with Grafana?

PR #10007, opened 2025-08-03
add instructions how to use Seafile with existing users

grasshide:
There is no documentation on how to switch existing users to OAuth in Seafile. There is only little information that explains how this has to be done and how it works. Therefore it was added to the documentation.

## Summary by CodeRabbit

* **Documentation**
  * Updated OAuth attribute mapping instructions for Seafile, requiring "sub" and making "email" optional.
  * Added guidance on migrating existing users from local authentication to external OAuth.
  * Included a link to the official Seafile migration guide in the "See Also" section.


docs/content/integration/openid-connect/clients/seafile/index.md
```diff
@@ -113,15 +113,70 @@ OAUTH_SCOPE = [
...
++-----------------------------+-----------------------------------------+
+| email                       | passwd                                  |
++-------------+---------------------------------------------------------+
+| email@{{< sitevar name="domain" nojs="example.com" >}}        | !                                       |
++-----------------------------+-----------------------------------------+
+```
+
+How it works:
+- If the user logs in with OAuth, the id attribute "sub" from authelia is mapped to "uid" via the setting in `seahub_settings.py`.
+- Seafile now looks up the social_auth_usersocialauth.username for the uid from the request and maps it to EmailUser.email.
```
james-d-elliott:
```suggestion
### Existing Users

When using [Seafile] with external authentication you may have to perform manual steps to achieve this. 

The [See Also](#see-also) has a link to the [Seafile] `migrating from local user database to external authentication` guide which has been verified to work.  
```

james-d-elliott (changes requested):
Since they already maintain a guide and may change things on their end, to keep our docs fresh this is probably the best choice.

docs/content/integration/openid-connect/clients/seafile/index.md
grasshide: Sorry, I hope I did not mess up anything by clicking the merge button. I've not worked with this kind of workflow before.I thought that would add your suggestions to my branch.

I found the manual in the Seafile guide a bit confusing. They create this new "*@auth.local" account that Seafile will not be able to link to an existing **real** email. It's tricky to spot but  they actually address this in point two of the "general procedure" that older users should just use their email.

The main Issue for me was the attribute mapping in `seahub_settings.py` that currently is definitely wrong and is now corrected.

A hint like you proposed that existing users need to be migrated will help others understand why the config itself is not sufficient. (And hopefully save them a lot of frustration)

So I'm happy with your proposal! Thanks for looking into this.

grasshide:
I accepted your changes and merged all recent changes.

james-d-elliott (approved):
LGTM

Issue #10280, opened 2025-09-10
Need to know why the user file reload was skipped


robinlioret:
### Description

When triggering a reload of the user file, we often have a the log "Reload was triggered but it was skipped" which is not really useful by itself.
We need more information.

### Use Case

Troublehsoot why hot reload of the user sometimes failed (from the admin POV).

### Details

The issue can be addressed by adding logging to the [Reload](https://github.com/authelia/authelia/blob/910f283dacb43a60034221f542f227ff6e3341e6/internal/authentication/file_user_provider.go#L55) function.

There is 3 scenarios when the reload can be skipped:

1. mutex timeout
2. failed to load database (ErrNoContent)
3. Failed to reload for any other error

### Generative AI

No

### Pre-Submission Checklist

- [x] I agree to follow the [Code of Conduct](http://www.authelia.com/code-of-conduct)

- [x] I have checked for related issues and checked the documentation

PR #10281, opened 2025-09-10
refactor: enhance file reload logs

robinlioret:
Addresses  #10280 

## Summary by CodeRabbit

* New Features
  * Added a distinct "cooldown" indicator to show when a reload is skipped due to recent activity.
* Bug Fixes
  * Reload now surfaces explicit reasons (e.g., cooldown or no file content) instead of appearing as silent successes.
  * Improved consistency of reload outcomes and messages so users see accurate reasons when reloads are skipped or fail.


robinlioret:
I agree. But it's weird with the whole "skipped" behaviour. Maybe it should be success or fail, but no skip. What do you think ?

PR #10340, opened 2025-09-18
docs: fix bad path prefix for oidc-common links

gene1wood:
The `$config` value points to the wrong location in the docs tree and the specific links which use the `$config` value point to `.md` files which are incorrect. This results in these two links in every client OIDC doc page to be broken and lead to 404 errors.

* Change the `$config` value to the right location and remove the trailing slash
* Change the specific links to the non `.md` suffixed values

Note that this commit is in the authelia repo instead of the authelia website repo which is where the READMEs indicate this content is meant to be. It's unclear if the READMEs are incorrect or if this file is in the incorrect location. See https://github.com/authelia/authelia/discussions/10339


james-d-elliott (approved):
LGTM

PR #10370, opened 2025-09-24
docs(envoy-gateway): fix wrong syntax on SecurityPolicy

fma965:
After much trial and error i found the reason why my auth headers wasn't working.

remote-* does not seem to work, whether it did previously or not i don't know? but it doesn't any more, i also tried Remote-* incase it was case sensitive

also headersToExtAuth and failOpen belongs under extAuth not http, they fail validation being under http

```
dry-run failed: failed to create typed patch object (gateway.envoyproxy.io/v1alpha1, Kind=SecurityPolicy): .spec.extAuth.http.failOpen: field not declared in schema

dry-run failed: failed to create typed patch object (gateway.envoyproxy.io/v1alpha1, Kind=SecurityPolicy): .spec.extAuth.http.headersToExtAuth: field not declared in schema
```

### DIFF
--- a/docs/content/integration/kubernetes/envoy/gateway.md
+++ b/docs/content/integration/kubernetes/envoy/gateway.md
@@ -74,22 +74,24 @@ spec:
       kind: 'Gateway'
       name: 'eg'
   extAuth:
+    headersToExtAuth:
+      - 'accept'
+      - 'cookie'
+      - 'authorization'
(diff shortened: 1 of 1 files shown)


fma965:
With it set to `remote-*`
image

with it set manually to the 4 values
image

Crowley723 (approved):

PR #10383, opened 2025-09-26
refactor(authentication): ldap pool logging

james-d-elliott:
This adds additional logging to the Pooled LDAP Client Factory.


internal/authentication/ldap_client_factory.go
```diff
@@ -110,8 +113,10 @@ func NewPooledLDAPClientFactory(config *schema.AuthenticationBackendLDAP, certs
 }
 
 // PooledLDAPClientFactory is a LDAPClientFactory that takes another LDAPClientFactory and pools the
-// factory generated connections using a channel for thread safety.
+// factory generated connections using a buffered channel for thread safety.
```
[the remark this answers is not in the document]
thielj: @james-d-elliott "or replace the mutex+channel mix with a single concurrency primitive (e.g., manage the pool from a dedicated goroutine or use a sync.Cond) so no blocking op occurs
while a mutex is held." 

That stupid rabbit is right every now and then...

internal/authentication/ldap_client_factory.go
Crowley723: Nit: At first glance the way `f.next` is being used appears thread-unsafe, but its not because the function calling `new()` is holding a lock for the client factory.

Might be worth noting that while the function in isolation is thread-unsafe, the way it is used is not necessarily unsafe.

```suggestion
// The new function creates a pool based client. This function is not thread safe but is called from a function that requires holding a lock. The intended usage of this function IS thread safe.
```

internal/authentication/ldap_client_factory.go
```diff
@@ -209,35 +247,36 @@ func (f *PooledLDAPClientFactory) acquire(ctx context.Context) (client *LDAPClie
 
 	if cap(f.pool) != f.config.Pooling.Count {
 		if err = f.Initialize(); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("error acquiring client: error initializing buffer: %w", err)
```
Crowley723:
```suggestion
			return nil, fmt.Errorf("error acquiring client: failed to initialize pool: %w", err)
```

PR #10392, opened 2025-09-26
issue 9936 - contention in LDAP pool

thielj:
See https://github.com/authelia/authelia/issues/9936

AI summary follows.

If you're allergic to comments and logging, I can make this as bare as you prefer.

### **New PooledLDAPClientFactory**
- **Thread-safe connection pool** with configurable min/max sizes and client lifetime management
- **Channel-based architecture** using buffered channels for available clients and wakeup signaling  
- **Dedicated pool manager goroutine** that maintains pool size, handles demand calculation, and performs graceful cleanup
- **Defensive channel capturing** pattern prevents race conditions and panics from concurrent access

### **Key Features**
- **Health & expiration checking** - clients validated on acquisition/release with randomized lifetime fuzzing
- **Retry logic with exponential backoff** for transient connection failures
- **Graceful shutdown** with timeout-based cleanup and context cancellation as fallback
- **Robust error handling** for edge cases like duplicate releases, late calls, and pool closure scenarios

### **Production Hardening**
- **Atomic operations** for closing flag and active count tracking
- **Comprehensive logging** for debugging and monitoring pool behavior
- **Configuration validation** with warnings for insufficient timeout settings
- **Metrics support** for observability (active/available client counts)

### **Implementation Details**
- Pool primed on initialization with readiness check
- Demand calculation balances minimum pool size vs. burst limiting
- Client lifetime fuzzed 10% to prevent thundering herd expiration
- Thread-safe methods handle uninitialized/closed states gracefully

The implementation provides enterprise-grade LDAP connection pooling with excellent concurrency safety

## Summary by CodeRabbit

- New Features
  - Added readiness/health checks and basic operational metrics for LDAP connections.
  - Introduced graceful startup and shutdown of the LDAP connection pool.

- Bug Fixes
  - Prevents hangs by disposing unhealthy or expired LDAP connections.
  - Ensures connections are properly closed to reduce resource leaks.

- Refactor
  - Revamped LDAP connection pooling with smart sizing, non-blocking acquisition, soft expiration, and enhanced logging/instrumentation.

- Tests
  - Adjusted LDAP tests to assert health checks and tolerate clean shutdown behavior.


internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+				logging.Logger().Info("Received invalid client from LDAP pool, disposing")
+				_ = f.disposeClient(client)
+				continue
+			}
+			// NOTE: we do NOT check for expiration here - it's a randomized "soft" target.
+			// Rationale: expiration is ony checked on release to avoid potential starvation.
+			// Clients may be used slightly beyond their expiration time, but this is acceptable.
+			logging.Logger().Tracef("Successfully acquired valid client from LDAP pool after %s", time.Since(now))
+			return client, nil
 		}
```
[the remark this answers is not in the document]
thielj: Dear rabbit.... In the majority of cases the request is served in a couple nanoseconds. That's one timer. If it takes a bit longer - while the pool is still filling up - we're looking at a wait of typically less than 300ms: 30 timers. Worst case, we're hitting a ~30 seconds timeout: 3000 timers.

But if this makes you happy bunny, you can get your ticker...

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 
-	if pool, ok := client.(*LDAPClientPooled); !ok || cap(f.pool) == len(f.pool) {
-		// Prevent extra or non-pool connections from being returned into the pool.
-		return client.Close()
-	} else {
-		f.pool <- pool
+	if !pooled.IsHealthy() || pooled.IsExpired() {
+		logging.Logger().Info("Pooled LDAP client is not healthy or has expired, disposing")
+		return f.disposeClient(pooled)
 	}
```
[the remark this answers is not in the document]
thielj: I appreciate that you share my thoughts about tracking lease status!

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+	minPoolSize    int32                  // Minimum number of pool connections to maintain (soft target)
+	maxPoolSize    int32                  // Maximum number of pool connections (hard target)
+	clientLifetime time.Duration          // Maximum lifetime for a pooled client (soft target)
+
+	// Synchronization
+	wakeup  chan struct{}      // Channel for client requests
+	closing atomic.Bool        // Atomic flag indicating if the pool is closing
+	closed  chan error         // Signals when poolManager has completed cleanup
+	ctx     context.Context    // Context for cancellation
+	cancel  context.CancelFunc // Function to cancel the context
```
james-d-elliott: Remove all these comments as they are unnecessary.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+// initializes the pool and starts the pool manager goroutine
+// must be called only once
 func (f *PooledLDAPClientFactory) Initialize() (err error) {
-	f.mu.Lock()
-
-	defer f.mu.Unlock()
-
-	if f.pool != nil {
-		return nil
+	if f.pool != nil { // Already initialized
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		return fmt.Errorf("LDAP pool is already initialized")
 	}
 
-	f.pool = make(chan *LDAPClientPooled, f.config.Pooling.Count)
-
-	var (
-		errs   []error
-		client *LDAPClientPooled
-	)
+	minRecommendedTimeout := f.config.Timeout * time.Duration(f.config.Pooling.Retries+2) // +2 to account for initial attempt and backoff
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -4,15 +4,16 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
-	"errors"
 	"fmt"
+	"math/rand"
 	"net"
-	"sync"
+	"sync/atomic"
```
james-d-elliott: Don't use this package please. This introduces a significant unnecessary complexity, while this package is thread safe it does not guarantee order which can have deadlock interactions with channels.

internal/authentication/ldap_client_factory.go
```diff
@@ -98,15 +99,18 @@ func NewPooledLDAPClientFactory(config *schema.AuthenticationBackendLDAP, certs
...
-		tls:    tlsc,
-		opts:   opts,
-		dialer: dialer,
-		sleep:  sleep,
+	factory = &PooledLDAPClientFactory{
+		config:        	config,
+		tls:           	tlsc,
+		opts:          	opts,
+		dialer:        	dialer,
+		// these could possibly be configured, or initialized differently
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -98,15 +99,18 @@ func NewPooledLDAPClientFactory(config *schema.AuthenticationBackendLDAP, certs
...
+		minPoolSize:    int32(max(2, config.Pooling.Count/2)),
+		maxPoolSize:    int32(max(2, config.Pooling.Count)),
+		clientLifetime: time.Hour, // this is a soft target, actual lifetime is fuzzed slightly
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
-			continue
-		}
-
-		f.pool <- client
-	}
+	f.ctx, f.cancel = context.WithCancel(context.Background())
 
-	if len(errs) == f.config.Pooling.Count {
-		return fmt.Errorf("errors occurred initializing the client pool: no connections could be established")
+	// prime the pool to test connectivity and run a quick self-check
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+	if !f.tryAddPooledClient(f.pool) {
+		err = fmt.Errorf("LDAP pool couldn't acquire initial client")
+	} else if rerr := f.ReadinessCheck(); rerr != nil {
+		err = rerr // failed readiness check
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+	} else {
+		go f.poolManager() // transfer responsibility to the pool manager
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
-	return nil
+	logging.Logger().Debug("LDAP pool initialization failed, aborting...")
+	err = fmt.Errorf("LDAP pool initialization failed: %w", err)
+	f.closed<-err
+	_ = f.Close()
+	return err
 }
 
-// GetClient opens new client using the pool.
+// thread-safe, opens new client using the pool.
```
james-d-elliott: Revert this change, it doesn't comply with our style guide.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 	}
 
-	return f.acquire(context.Background())
+	return f.acquire()
 }
 
+// TODO: Clarify why this method is not thread-safe - or does this only apply to the returnd client?
+
 // The new function creates a pool based client. This function is not thread safe.
+// NOTE: this function adheres to config.Timeout, not config.Pooling.Timeout!
```
james-d-elliott: Remove unnecessary additional comment. This function creates a new client, the pooling timeout has nothing to do with it.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
-		return nil, fmt.Errorf("error occurred establishing new client for the pool: %w", err)
-	}
+	client, err := getLDAPClient(f.config.Address.String(), f.config.User, f.config.Password, f.config.Timeout, f.dialer, f.tls, f.config.StartTLS, f.opts)
+	if err != nil { return nil, err }
 
 	return &LDAPClientPooled{Client: client}, nil
 }
 
-// ReleaseClient returns a client using the pool or closes it.
+// thread-safe, returns a client using the pool or closes it.
```
james-d-elliott: Revert this change as it doesn't adhere to our style guide.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 
-	return nil
+	pC := f.pool
+	if pC == nil {
+		logging.Logger().Warning("LDAP pool is not initialized or was closed, disposing client")
+		return f.disposeClient(pooled)
+	}
+
+	// NOTE: clients can still be returned while the pool is closing.
+	// Rationale: cleanup in poolManager() handles this gracefully.
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+
+	select {
+	case pC <- pooled:
+		logging.Logger().Trace("Successfully returned client to LDAP pool")
+		return nil
+	default: // shouldn't happen, unless returning more clients than acquired
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		logging.Logger().Debug("LDAP pool is full, disposing client")
+		return f.disposeClient(pooled)
+	}
 }
 
-func (f *PooledLDAPClientFactory) acquire(ctx context.Context) (client *LDAPClientPooled, err error) {
-	f.mu.Lock()
+// thread-safe, closes the client and decrements active count
```
james-d-elliott: Comments should start with capitals and end with periods.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 
-	if cap(f.pool) != f.config.Pooling.Count {
-		if err = f.Initialize(); err != nil {
-			return nil, err
+	now := time.Now()
+	deadline := now.Add(f.config.Pooling.Timeout)
+	interval := time.Millisecond * 10
+
+	for time.Now().Before(deadline) {
+		f.wakeupPoolManager() // request more clients if needed
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		select {
+		case <-time.After(interval): // this avoids the blocking time.Sleep()
+			continue // do nothing, except to loop again to signal the pool manager
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
-		if err = f.Initialize(); err != nil {
-			return nil, err
+	now := time.Now()
+	deadline := now.Add(f.config.Pooling.Timeout)
+	interval := time.Millisecond * 10
+
+	for time.Now().Before(deadline) {
+		f.wakeupPoolManager() // request more clients if needed
+		select {
+		case <-time.After(interval): // this avoids the blocking time.Sleep()
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+			continue // do nothing, except to loop again to signal the pool manager
+		case client = <-pC:
+			if !client.IsHealthy() { // NOTE: we dispose expired clients only on release
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+	cancel  context.CancelFunc // Function to cancel the context
+}
 
-	closing bool
+func (f *PooledLDAPClientFactory) isClosing() bool {
+	return f.closing.Load()
 }
 
+// initializes the pool and starts the pool manager goroutine
+// must be called only once
```
james-d-elliott: Should start with the function name and end in a period.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+			continue // do nothing, except to loop again to signal the pool manager
+		case client = <-pC:
+			if !client.IsHealthy() { // NOTE: we dispose expired clients only on release
+				logging.Logger().Info("Received invalid client from LDAP pool, disposing")
+				_ = f.disposeClient(client)
+				continue
+			}
+			// NOTE: we do NOT check for expiration here - it's a randomized "soft" target.
+			// Rationale: expiration is ony checked on release to avoid potential starvation.
+			// Clients may be used slightly beyond their expiration time, but this is acceptable.
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+			return client, nil
 		}
 	}
+	return nil, fmt.Errorf("timeout waiting for client from LDAP pool")
+}
+
+// must be called EXCLUSIVELY by Initialize() or poolManager()
+func (f *PooledLDAPClientFactory) tryAddPooledClient(pC chan *LDAPClientPooled ) (bool) {
+	attempts := f.config.Pooling.Retries
+	sleep := f.config.Timeout / 8 // initial sleep duration, trade-off between responsiveness and overload
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+			// Rationale: expiration is ony checked on release to avoid potential starvation.
+			// Clients may be used slightly beyond their expiration time, but this is acceptable.
+			logging.Logger().Tracef("Successfully acquired valid client from LDAP pool after %s", time.Since(now))
+			return client, nil
 		}
 	}
+	return nil, fmt.Errorf("timeout waiting for client from LDAP pool")
+}
+
+// must be called EXCLUSIVELY by Initialize() or poolManager()
```
james-d-elliott: Must begin with a capital and end in a period, also there is no need to put other emphasis or punctuation in comments, please make `EXCLUSIVELY` become `exclusively`.  It may also be beneficial to name the function after the function that should be calling it.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+func (f *PooledLDAPClientFactory) tryAddPooledClient(pC chan *LDAPClientPooled ) (bool) {
+	attempts := f.config.Pooling.Retries
+	sleep := f.config.Timeout / 8 // initial sleep duration, trade-off between responsiveness and overload
+
+	for !f.isClosing() {
+		start := time.Now()
+		client, err := f.new() // NOTE: new() adheres to config.Timeout, not config.Pooling.Timeout!
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		if err == nil {
+			// Shorten client lifetime by up to 10% to avoid mass exodus events
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+			fuzzFactor := 0.90 + 0.1*rand.Float64() // [0.90..1.00)
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+			client.expiresAt = time.Now().Add(time.Duration(float64(f.clientLifetime) * fuzzFactor))
+
+			select {
+			case pC <- client:
+				active := f.activeCount.Add(1)
+				logging.Logger().Debugf("Added new client to LDAP pool. Active: %d, Available: %d", active, len(pC))
+				return true
+			default: // shouldn't happen
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		}
 
-	ctx, cancel := context.WithTimeout(ctx, f.config.Pooling.Timeout)
-	defer cancel()
+		logging.Logger().Tracef("Sleeping %s before next attempt to create pooled LDAP client", sleep)
+		time.Sleep(sleep)
+		sleep *= 2
+		attempts--
+	}
+	return false // pool is closing
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		return f.disposeClient(pooled)
+	}
 }
 
-func (f *PooledLDAPClientFactory) acquire(ctx context.Context) (client *LDAPClientPooled, err error) {
-	f.mu.Lock()
+// thread-safe, closes the client and decrements active count
+func (f *PooledLDAPClientFactory) disposeClient(client *LDAPClientPooled) error {
+	if client == nil {
+		logging.Logger().Warn("Cannot dispose nil LDAP client")
```
james-d-elliott: Please remove call to `logging.Logger()` and initialize a fielded logger like in #10383

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 
-	f.pool = make(chan *LDAPClientPooled, f.config.Pooling.Count)
-
-	var (
-		errs   []error
-		client *LDAPClientPooled
-	)
+	minRecommendedTimeout := f.config.Timeout * time.Duration(f.config.Pooling.Retries+2) // +2 to account for initial attempt and backoff
+	if f.config.Pooling.Timeout <= minRecommendedTimeout {
+		logging.Logger().Warnf("LDAP pooling timeout (%v) may be insufficient for retry scenarios (recommended: >%v)",
```
james-d-elliott: Please remove call to `logging.Logger()` and initialize a fielded logger like in #10383, include the braces in another calll to WithFields, and avoid using Warnf and use Warn instead.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		err = fmt.Errorf("LDAP pool couldn't acquire initial client")
+	} else if rerr := f.ReadinessCheck(); rerr != nil {
+		err = rerr // failed readiness check
+	} else {
+		go f.poolManager() // transfer responsibility to the pool manager
+		return nil
 	}
 
-	return nil
+	logging.Logger().Debug("LDAP pool initialization failed, aborting...")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 	return &LDAPClientPooled{Client: client}, nil
 }
 
-// ReleaseClient returns a client using the pool or closes it.
+// thread-safe, returns a client using the pool or closes it.
 func (f *PooledLDAPClientFactory) ReleaseClient(client ldap.Client) (err error) {
-	f.mu.Lock()
-
-	defer f.mu.Unlock()
+	logging.Logger().Trace("Releasing LDAP client")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 
-	if f.closing {
+	pooled, ok := client.(*LDAPClientPooled)
+	if !ok {
+		logging.Logger().Debug("LDAP client is not pooled, closing directly")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 		return client.Close()
 	}
 
-	if pool, ok := client.(*LDAPClientPooled); !ok || cap(f.pool) == len(f.pool) {
-		// Prevent extra or non-pool connections from being returned into the pool.
-		return client.Close()
-	} else {
-		f.pool <- pool
+	if !pooled.IsHealthy() || pooled.IsExpired() {
+		logging.Logger().Info("Pooled LDAP client is not healthy or has expired, disposing")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+	remaining := f.activeCount.Add(-1)
+	logging.Logger().Tracef("Pooled LDAP client disposed; remaining: %d", remaining)
+	return client.Close()
+}
 
-	if f.closing {
-		return nil, fmt.Errorf("error acquiring client: the pool is closed")
+// thread-safe, returns a client using the pool or an error
+func (f *PooledLDAPClientFactory) acquire() (client *LDAPClientPooled, err error) {
+	logging.Logger().Trace("Acquiring client from LDAP pool")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		client, err := f.new() // NOTE: new() adheres to config.Timeout, not config.Pooling.Timeout!
+		if err == nil {
+			// Shorten client lifetime by up to 10% to avoid mass exodus events
+			fuzzFactor := 0.90 + 0.1*rand.Float64() // [0.90..1.00)
+			client.expiresAt = time.Now().Add(time.Duration(float64(f.clientLifetime) * fuzzFactor))
+
+			select {
+			case pC <- client:
+				active := f.activeCount.Add(1)
+				logging.Logger().Debugf("Added new client to LDAP pool. Active: %d, Available: %d", active, len(pC))
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+				return true
+			default: // shouldn't happen
+				logging.Logger().Warn("LDAP pool closed or full, closing newly created client")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+				_ = client.Close()
+				return false
+			}
+		}
+		logging.Logger().Debugf("Failed to create new client for LDAP pool. Attempt %d/%d, Elapsed: %s: %v",
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+			f.config.Pooling.Retries - attempts + 1, f.config.Pooling.Retries + 1, time.Since(start), err)
+
+		if attempts == 0 {
+			logging.Logger().Warningf("Exceeded maximum attempts (%d) for creating pooled LDAP client; last error: %v",
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+				f.config.Pooling.Retries + 1, err)
+			return false
+		}
 
-	ctx, cancel := context.WithTimeout(ctx, f.config.Pooling.Timeout)
-	defer cancel()
+		logging.Logger().Tracef("Sleeping %s before next attempt to create pooled LDAP client", sleep)
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		time.Sleep(sleep)
+		sleep *= 2
+		attempts--
+	}
+	return false // pool is closing
+}
 
+// thread-safe, does not block
```
james-d-elliott: Please ensure this starts with a capital and ends with a period, it should also be a sentence or be removed.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+func (f *PooledLDAPClientFactory) wakeupPoolManager() {
+	wakeC := f.wakeup
+	if wakeC == nil { return } // pool is closing or not initialized
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
-					if client, err = f.new(); err != nil {
-						time.Sleep(f.sleep)
-
-						continue
-					}
-
-					return client, nil
+	case wakeC <- struct{}{}:
+	default: // Already signaled, do nothing
+	}
```
james-d-elliott: Remove unnecessary comment and add a break.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+}
+
+// goroutine that maintains the pool of LDAP clients
+func (f *PooledLDAPClientFactory) poolManager() {
+
+	pCtx, pC, wakeC, retC := f.ctx, f.pool, f.wakeup, f.closed
+	if pC == nil || pCtx == nil || wakeC == nil || retC == nil {
+		logging.Logger().Error("LDAP pool manager cannot start: some components are not initialized")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
-						continue
-					}
-
-					return client, nil
+	case wakeC <- struct{}{}:
+	default: // Already signaled, do nothing
+	}
+}
+
+// goroutine that maintains the pool of LDAP clients
```
james-d-elliott: Comment style, should start with a capital end with a period.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+func (f *PooledLDAPClientFactory) poolManager() {
+
+	pCtx, pC, wakeC, retC := f.ctx, f.pool, f.wakeup, f.closed
+	if pC == nil || pCtx == nil || wakeC == nil || retC == nil {
+		logging.Logger().Error("LDAP pool manager cannot start: some components are not initialized")
+		close(f.closed)
+		return
+	}
+
+	logging.Logger().Infof("LDAP Pool Manager started: minPoolSize=%d, maxPoolSize=%d, retries=%d, timeout=%v",
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		f.minPoolSize, f.maxPoolSize, f.config.Pooling.Retries, f.config.Pooling.Timeout)
+
+	f.wakeupPoolManager() // initial wakeup to prime the pool
+
+	for !f.isClosing() {
+		select {
+		case <-pCtx.Done():
+			logging.Logger().Warn("LDAP pool context cancelled, exiting...")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
-		f.pool <- pool
+	if !pooled.IsHealthy() || pooled.IsExpired() {
+		logging.Logger().Info("Pooled LDAP client is not healthy or has expired, disposing")
+		return f.disposeClient(pooled)
 	}
 
-	return nil
+	pC := f.pool
+	if pC == nil {
+		logging.Logger().Warning("LDAP pool is not initialized or was closed, disposing client")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		return f.disposeClient(pooled)
+	}
+
+	// NOTE: clients can still be returned while the pool is closing.
+	// Rationale: cleanup in poolManager() handles this gracefully.
+
+	select {
+	case pC <- pooled:
+		logging.Logger().Trace("Successfully returned client to LDAP pool")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+// thread-safe, closes the client and decrements active count
+func (f *PooledLDAPClientFactory) disposeClient(client *LDAPClientPooled) error {
+	if client == nil {
+		logging.Logger().Warn("Cannot dispose nil LDAP client")
+		return nil
+	}
 
-	defer f.mu.Unlock()
+	remaining := f.activeCount.Add(-1)
+	logging.Logger().Tracef("Pooled LDAP client disposed; remaining: %d", remaining)
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 
-	if f.closing {
-		return nil, fmt.Errorf("error acquiring client: the pool is closed")
+// thread-safe, returns a client using the pool or an error
+func (f *PooledLDAPClientFactory) acquire() (client *LDAPClientPooled, err error) {
+	logging.Logger().Trace("Acquiring client from LDAP pool")
+
+	pC := f.pool
+	if pC == nil || f.isClosing() {
+		logging.Logger().Error("Cannot acquire client: LDAP pool is not initialized or closing")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		case client = <-pC:
+			if !client.IsHealthy() { // NOTE: we dispose expired clients only on release
+				logging.Logger().Info("Received invalid client from LDAP pool, disposing")
+				_ = f.disposeClient(client)
+				continue
+			}
+			// NOTE: we do NOT check for expiration here - it's a randomized "soft" target.
+			// Rationale: expiration is ony checked on release to avoid potential starvation.
+			// Clients may be used slightly beyond their expiration time, but this is acceptable.
+			logging.Logger().Tracef("Successfully acquired valid client from LDAP pool after %s", time.Since(now))
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+	logging.Logger().Infof("LDAP Pool Manager started: minPoolSize=%d, maxPoolSize=%d, retries=%d, timeout=%v",
+		f.minPoolSize, f.maxPoolSize, f.config.Pooling.Retries, f.config.Pooling.Timeout)
+
+	f.wakeupPoolManager() // initial wakeup to prime the pool
+
+	for !f.isClosing() {
+		select {
+		case <-pCtx.Done():
+			logging.Logger().Warn("LDAP pool context cancelled, exiting...")
+			return // This exits the goroutine 'ungracefully', without further cleanup
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+			request = min( request, f.maxPoolSize - active)		// Goal 3, never exceed maxPoolSize
+			if request <= 0 { continue }
+			request = min(request, (f.minPoolSize + 1) / 2)		// Goal 4, limit bursts to half minPoolSize (rounded up)
+
+			logging.Logger().Debugf("LDAP Pool clients available: (%d/%d), requesting +%d increase", available, active, request)
+
+			for range request {
+				if !f.tryAddPooledClient(pC) {
+					break 																				// Goal 5: stop if we can't add more clients
+					// TODO: consider a more generalized backoff strategy here
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+	// no timeout here, Close() handles cancelling f.ctx if necessary
+	for f.activeCount.Load() > 0 {
+		select {
+		case <-pCtx.Done():
+			logging.Logger().Warn("LDAP pool context cancelled during cleanup")
+			retC <- fmt.Errorf("LDAP pool context cancelled during cleanup")
+			return
+		case client, ok := <-pC:
+			if !ok {
+				logging.Logger().Warn("LDAP pool channel unexpectedly closed during cleanup")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
-		return client, nil
+poolCleanup:
+
+	logging.Logger().Debugf("LDAP pool is closing, cleaning up %d pooled clients", f.activeCount.Load())
+
+	// no timeout here, Close() handles cancelling f.ctx if necessary
+	for f.activeCount.Load() > 0 {
+		select {
+		case <-pCtx.Done():
+			logging.Logger().Warn("LDAP pool context cancelled during cleanup")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+					// TODO: consider a more generalized backoff strategy here
 				}
 			}
 		}
+	}
 
-		return client, nil
+poolCleanup:
+
+	logging.Logger().Debugf("LDAP pool is closing, cleaning up %d pooled clients", f.activeCount.Load())
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+			retC <- fmt.Errorf("LDAP pool context cancelled during cleanup")
+			return
+		case client, ok := <-pC:
+			if !ok {
+				logging.Logger().Warn("LDAP pool channel unexpectedly closed during cleanup")
+				retC <- fmt.Errorf("LDAP pool channel unexpectedly closed during cleanup")
+				return
+			}
+			_ = f.disposeClient(client)
+			logging.Logger().Tracef("Closed pooled LDAP client, remaining clients: %d", f.activeCount.Load())
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+	close(retC) // don't send an error value, just close to signal completion
 }
 
+// Cooperates with the pool manager goroutine to cleanup and exit gracefully
+// must be called only once - subsequent calls are ignored
 func (f *PooledLDAPClientFactory) Close() (err error) {
-	f.mu.Lock()
-
-	defer f.mu.Unlock()
+	logging.Logger().Trace("Closing LDAP connection pool")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+			}
+			_ = f.disposeClient(client)
+			logging.Logger().Tracef("Closed pooled LDAP client, remaining clients: %d", f.activeCount.Load())
+		}
 	}
+	close(retC) // don't send an error value, just close to signal completion
 }
 
+// Cooperates with the pool manager goroutine to cleanup and exit gracefully
+// must be called only once - subsequent calls are ignored
```
james-d-elliott: Should start with Close, shouldn't include hyphens, and should end in a period.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 func (f *PooledLDAPClientFactory) Close() (err error) {
-	f.mu.Lock()
-
-	defer f.mu.Unlock()
+	logging.Logger().Trace("Closing LDAP connection pool")
 
-	f.closing = true
+	if f.pool == nil {
+		logging.Logger().Warn("LDAP pool is not initialized, nothing to close")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+		return nil
+	}
 
-	close(f.pool)
+	if !f.closing.CompareAndSwap(false, true) {
+		logging.Logger().Warn("LDAP pool already closing, ignoring request")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+	cleanupCtx, cancelCleanup := context.WithTimeout(context.Background(), 2 * f.config.Timeout)
+	defer cancelCleanup()
 
-		if err = client.Close(); err != nil {
-			errs = append(errs, err)
+	select {
+	case cleanupErr, hasError := <-f.closed:
+		err = cleanupErr // maybe nil, if there was no error
+		if !hasError {
+			logging.Logger().Debug("LDAP pool cleanup complete, all clients closed")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 		}
+		f.closed = nil
+	case <-cleanupCtx.Done(): // timeout
+		f.cancel() // shoot the pool manager, if it hasn't exited gracefully
+		logging.Logger().Warn("LDAP pool cleanup exceeded timeout")
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
-			errs = append(errs, err)
+	select {
+	case cleanupErr, hasError := <-f.closed:
+		err = cleanupErr // maybe nil, if there was no error
+		if !hasError {
+			logging.Logger().Debug("LDAP pool cleanup complete, all clients closed")
 		}
+		f.closed = nil
+	case <-cleanupCtx.Done(): // timeout
+		f.cancel() // shoot the pool manager, if it hasn't exited gracefully
```
james-d-elliott: Remove unnecessary comments.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
 }
 
+// thread-safe, returns ( active: 0, available: 0) if pool is not initialized or closed
+func (f *PooledLDAPClientFactory) Metrics() (active, available int) {
+	pC := f.pool
+	if pC == nil {
+		return 0, 0
+	}
+	return int(f.activeCount.Load()), len(pC)
+}
```
james-d-elliott:
```suggestion
```

This is not used.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+
+
+
+
 // LDAPClientPooled is a decorator for the ldap.Client which handles the pooling functionality. i.e. prevents the client
 // from being closed and instead relinquishes the connection back to the pool.
 type LDAPClientPooled struct {
 	ldap.Client
+	expiresAt time.Time
+	// leased bool // prevent double leasing / double returns
```
james-d-elliott: Remove unnecessary comment.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+}
+
+func (c *LDAPClientPooled) IsExpired() bool {
+	return time.Now().After(c.expiresAt)
+}
```
james-d-elliott:
```suggestion
```

Clients should not expire.

internal/authentication/ldap_client_factory.go
```diff
@@ -117,162 +121,393 @@ type PooledLDAPClientFactory struct {
...
+func (c *LDAPClientPooled) IsHealthy() bool {
+	return c.Client != nil && !c.Client.IsClosing()
+}
+
+func (c *LDAPClientPooled) Close() error {
+	if c.Client != nil {
+		logging.Logger().Trace("Closing LDAP client")
+		err := c.Client.Close()
+		if err != nil {
+			logging.Logger().Warnf("Error closing LDAP client: %v", err)
```
james-d-elliott: Please remove call to logging.Logger() and initialize a fielded logger like in #10383.

james-d-elliott (changes requested):
Thanks for choosing to use your time to contribute! This is my initial review. 

In general these changes are too vast without any additional testing, and a lack of design consultation with maintainers. We would need this design consultation process to occur before we move ahead at all.

Could you please indicate if you made these changes yourself, based on another reference design, or using other tools like AI/ML?

Could you please detail the critical measures you used to prevent channel deadlocks?

internal/authentication/ldap_client_factory.go
thielj: Then default them to 100 years. Some servers don't like it when connections are held too long as their strategy is to release resources only when connections end.

Similar with load balancing; sessions that last forever won't migrate to a different server unless they are forcefully killed - which means Authelia loses all connections at once.

internal/authentication/ldap_client_factory.go
thielj: You consider this unnecessary?

internal/authentication/ldap_client_factory.go
thielj: I can just use sync and write this - if it makes you happy. It's slower and awkward.

```
func (f *PooledLDAPClientFactory) activeCountLoad() int32 {
	return f.activeCountAdd(0)
}

func (f *PooledLDAPClientFactory) activeCountAdd(v int32) int32 {
	f.mu.Lock()
	defer f.mu.Unlock()
	f.activeCount += v
	return f.activeCount
}
```

But seriously, I'm certainly not going to sprinkle locks and unlocks over code that is perfectly safe without.

Point me to an actual issue and we can discuss this.

internal/authentication/ldap_client_factory.go
thielj: I know, and it has been replaced on my side already, generating a full set of metrics.

```
======== LDAP Pool Metrics Report ========
 Report Time: 2025-09-27 22:45:48 UTC
 Collection Duration: 0.05 hours
 Current State: 31 active, 17 available
 Max Active Clients: 31
 Avg Active Clients: 31.00

 Successful Acquisitions by Time Range:
    <10us: 1154   avg: 1.92μs
   <100us: 18
     <1ms: 5
    <10ms: 17
   <100ms: 44
      <1s: 16
      ≥1s: 0
    TOTAL: 1254

Failures:
   Timeout failures: 0

Client Management:
            Clients created: 31
        Unhealthy disposals: 0
          Avg creation time: 1442.44ms
          Creation failures: 0
   Creation failed attempts: 0
        Pool drained events: 76
       Pool manager wakeups: 22
==========================================
```

So maybe we can steer this discussion towards how I can make those metrics available?

internal/authentication/ldap_client_factory.go
thielj: this i now implemented.

thielj:
The actual code isn't that much once you remove logging and error handling.

I've implemented a lock free connection pool, using "modern" concurrency patterns like channels and atomic operations. Modern, as in, Windows and Linux had support for interlocked ("atomic") operations since the 90s and .NET since v1.0 (early 2000s). They simply translate into [platform specific CPU instructions](https://github.com/golang/go/tree/34e67623a81e9e93c3d7d0f0cb257d7d722939f2/src/internal/runtime/atomic). Lock-free queues (similar to Go's channels) also appeared in the early 2000s and were standardized by 2010. And yes, I am that old to have done all this.

I like Authelia, but the current pool deadlocks and I need proper, pooled LDAP access. 

Below you will find all information you requested.

---

## Code walk-through

`Initialize()` probably doesn't need much explanation. It calls `tryAddPooledClient()` to create a connection which is safe as long as the `poolManager()` isn't running. It then does a readiness check that uses `acquire()` and `ReleaseClient()` intrnally. All this works without `poolManager` running. If init fails (most likely because a client couldn't be created), an error is put into the `closed` channel to prevent the following `Close()` call from blocking until timeout.

All relevant action happens on the `poolManager()` goroutine and there isn't much synchronization required. Clients are added to `pool`. The `wakeup` channel requests action from the `poolManager`. `isClosing()` is checked before e.g. creating new connections which could take a full `config.Timeout` to complete. 

The goroutine has three distinct phases: init, event loop and cleanup. It's using the `closed` channel to signal completion or an error status, received by `Close()`. You will find the mgmt heuristics in the event loop. The pool manager is currently 'pressured' from `acquire()` and `disposeClient()` for potential top-ups. `Close()` signals the pool manager to start cleanup. 

`tryAddPooledClient()` implements exponential back-off and doesn't require any synchronization beyond the `pool` channel and interlocked increments of `activeCount`.

Clients are leased through `acquire()`. There's no magic there either, just being careful not to block acquisition with `time.Sleep(sleep)` for 10 second "chunks". Synchronization is through the `pool` channel. 

Leased clients are returned through `ReleaseClient()`. If they are suitable for re-use, they are sent to the `pool` channel - no further synchronization required. Junk goes to `disposeClient()`, which is using an interlocked decrement for `activeCount`. Again, no further synchronization required.

Finally, `Close()` triggers a cleanup in `poolManager()` and waits for graceful completion. If this doesn't happen - callers not returning leased clients - the context is cancelled, and we call it a day.

And that's it.

---

I'll remove verbose comments and logging as the final step. And yes, I'll use AI for stupid chores like that.

Please don't bother me with remarks on piffle paffle.

---

There was no reference required. It's a very simple, clean, straight-forward design with one worker thread (aka goroutine) and no complex synchronization requirements. 

> Could you please detail the critical measures you used to prevent channel deadlocks? While they don't necessarily directly lead to them they often indirectly can.

There are none. Nothing. Nada. This is becoming like Uber Eats asking for a picture of the non-delivered pizza. If you find something that could cause "channel deadlocks", I'll buy beers for the team. 

See walkthrough of the code above for details on what synchronization patterns are used and where.

> Generally channels and sync/atomic should never be used together

This is wrong. Careless locks and channels are a recipe for disaster though. Listen to the rabbit. 

Please, read up what "atomic" or interlocked operations do!

---

If you really want to talk about "ordering", look at my code in `metricsCalculateSnapshot()`:

```go
  bin_10us := f.metricsSuccessfulBin0.Load()
  bin_10us_avg = float64(f.metricsBin0TimeSum.Load()) / float64(bin_10us) / 1e3
```

`metricsSuccessfulBin0` and `metricsBin0TimeSum` are atomically incremented in `acquire()`. What can happen without additional synchronization is the following:

| acquire() | metricsCalculateSnapshot() |
|---------------------|----------------------------|
| | f.metricsSuccessfulBin0.Load() |
| f.metricsSuccessfulBin0.Add(1) | |
| f.metricsBin0TimeSum.Add(t) | |
| | f.metricsBin0TimeSum.Load() |

The divisor would be off by 1 and the average slightly wrong. 

What still can't happen here is memory corruption, panics, lock-ups or anything like this. 

Now... We might disagree, but I rather prioritize to not lock and unlock a Mutex just to get the 100% perfect metrics and prefer low latency pool operations instead. Hence the comment:

```
// thread-safe, creates a snapshot of current metrics with all calculations performed
// NOTE: loads may overlap with updates - this is considered acceptable
func (f *PooledLDAPClientFactory) metricsCalculateSnapshot() *metricsSnapshot {
```

That is: this function is safe to call from different threads (goroutines) at the same time without further synchronization. It doesn't affect or even slow down pool operations at all. It doesn't corrupt anything. The use of atomic types here only guarantees that we don't get torn reads. 

This, for example can NOT happen:

| Memory | Operation |
|--------|-----------|
| BAAD *BABE* | Thread 1 reading lower word |
| BAAD *BEEF* | Thread 2 writing lower word |
| *DEAD* BEEF | Thread 2 writing upper word |
| *DEAD* BEEF | Thread 1 reading upper word |

The resulting "DEADBABE" is a "torn read", and this can happen if memory isn't properly aligned, or interlocked/atomic operations aren't used where necessary.

What does this have to do with Go channels? Nothing. Except, when you look into `runtime/chan.go` you will find atomic operations as well. They are fast, lightweight, non-locking and consistency is guaranteed by hardware.

And as I said, this does the same, just much, much slower:

```go
func (f *PooledLDAPClientFactory) activeCountLoad() int32 {
	return f.activeCountAdd(0)
}

func (f *PooledLDAPClientFactory) activeCountAdd(v int32) int32 {
	f.mu.Lock()
	defer f.mu.Unlock()
	f.activeCount += v
	return f.activeCount
}
```

james-d-elliott:
At this stage I'm not satisfied with the answers or the lack of collaboration here. If the goal here is to resolve 9936 as described there are several additional features here that have been added that wont be merged with a fix (metrics, client lifetimes, change to the fundamental operation, etc). In addition the original code was relatively simple, this adds ~700 lines to a ~200 line design, this is clearly several times more complex.

1. The `sync/atomic` package does introduce a non-insignificant maintenance burden.
2. There has been an attempt to make something faster with zero evidence there is a meaningful performance issue related to 1.
3. There have been no tests added that showcase the issue attempting to be fixed, and a general unwillingness to cooperate with reproducing the bug.
4. While you claim to answer my questions there are a few that you clearly didn't, and I don't agree with several key aspects that have been answered, for example the performance "issue" has no demonstrated evidence that it is true, there is a performance cost to mutexes but they are negligible for the given task on modern hardware. Combining that with the maturity of this solution, and a lack of evidence backing this being an actual issue, this is a clear sign of premature optimization; people are going to want to add features to this as you clearly do already.
5. The ordering issue really is important, currently the design appears to allow sending to a full buffered channel due to the use of atomic; specifically the value returned using atomic is not guaranteed to be the current value when sending to the channel. This will more than likely deadlock that go routine. 

I'm quite agreeable to adding metrics, but this should be a separate feature PR, and it should be implemented via prometheus. I'm also quite agreeable to trying to find the source of the issue and just implementing the actual fix. But I'm not convinced as of now that a complete redesign is necessary let alone a good choice.

thielj:
> the original code was relatively simple, this adds ~700 lines to a ~200 line design, this is clearly several times more complex.

It's now less than 200 lines, plus a few extra to replace sync/atomic with a Mutex. All additional features removed.

> The sync/atomic package does introduce a non-insignificant maintenance burden.

How? I hope you realize that `ldapClient.IsClosing()` returns an atomic.Bool? Anyway, removed.  

> The ordering issue really is important, currently the design appears to allow sending to a full buffered channel due to the use of atomic; specifically the value returned using atomic is not guaranteed to be the current value when sending to the channel. This will more than likely deadlock that go routine.

Show me where and how. This will be fun. 

> While you claim to answer my questions there are a few that you clearly didn't, 

I no longer won't. If you have actual technical questions, fine. But I won't bother with your nitpicking on comments and the constant whataboutism.

thielj:
Here's your challenge:

> The ordering issue really is important, currently the design appears to allow sending to a full buffered channel due to the use of atomic; specifically the value returned using atomic is not guaranteed to be the current value when sending to the channel. This will more than likely deadlock that go routine.

1. Demonstrate how my implementation can actually create more clients than the channel's capacity; and how, even if this would happen, this "will more than likely deadlock that go routine" or cause invalid operations, memory corruption, panics, memory leaks, etc.

2. Assuming that someone calls `ReleaseClient` with additional clients obtained through other means: demonstrate how this "will more than likely deadlock that go routine" or cause invalid operations, memory corruption, panics, memory leaks, etc.

I'll go and get the popcorn...

PS: Even if the go routine would "block" - the pool operations would simply continue until all client connections have closed. Take the readiness check done in `Initialize()` - it runs before the goroutine has even started.

thielj:
And if you're feeling adventurous:

1. Comment out the start of the goroutine in `Initialize()`. I haven't tried this, but Authelia should still work (with just one pooled connection, obviously) and even shutdown semi-clean with a timeout in `Close()`

2. Start 2 goroutines in `Initialize()`. Or 200 if your LDAP server can handle this. I haven't tried this either, but as long as you give each one its own return channel (like an `f.closed2`), Authelia should still work. You'll probably see the occasional parallel connection attempts and the pool might reach maxSize faster than planned as the heuristics are off, but all connections should be cleaned up on shutdown. `Close()` might timeout as above.

3. Change the heuristics calculations in `poolManager` to always request 2*maxPoolSize (or 200) *additional* connections. I haven't tried this either, but the worst thing that can happen is that you get a couple extra connections that will be properly disposed. Authelia will still work, and you will still get a 100% clean shutdown.

thielj:
Maybe my (non-native) English isn't good enough. I asked Claude 4.0 to summarize and explain the core design points, the problems in the previous implementation and asked for a neutral judgement of production readiness.

If this doesn't help, please remove my PR.

---

# LDAP Pool Redesign: Lock-Free Architecture Analysis

## 1. Client Leasing Operations (acquire/ReleaseClient)

The new design fundamentally separates client leasing from pool management through lock-free operations. Client acquisition and release work independently of the pool manager goroutine, ensuring that basic pooling operations cannot be blocked by background maintenance tasks.

```go
func (f *PooledLDAPClientFactory) acquire() (client *LDAPClientPooled, err error) {
    pool := f.pool
    if pool == nil {
        return nil, fmt.Errorf("error acquiring client: the pool is unitialized or closed")
    }

    for !f.closing_Load() {
        select {
        case client = <-pool: // fast path - direct channel operation
            // ...
        default: // slow path - pool empty
            f.wakeupPoolManager() // non-blocking signal
            select {
            case client = <-pool:
                // ...
            case <-time.After(time.Millisecond * 10):
                // timeout handling with retries
            }
        }
        // ... health checks and validation
    }
}
```

All synchronization between leasing operations occurs through the pool channel itself, which provides thread-safe FIFO access to available clients. In worst-case scenarios where the pool is empty, the system operates on a one-in-one-out principle, and uses non-blocking signals to the pool manager to signal pressure.

All operations are designed to be non-blocking from the perspective of other goroutines. Client acquisition either succeeds immediately, waits with a timeout, or fails gracefully. Client release either returns the client to the pool immediately or disposes of it if the pool is full or the client is unhealthy.

Communication with the pool manager occurs solely through the wakeup channel using non-blocking sends. The leasing operations never create new connections directly but will dispose of unhealthy or disconnected clients to maintain pool integrity.

## 2. Pool Manager Design

The pool manager operates as a completely independent background goroutine that has no role in the actual client leasing process and cannot block those operations. It responds only to signals received through the wakeup channel, making independent assessments of the current pool situation.

```go
func (f *PooledLDAPClientFactory) poolManager(ctx context.Context, pool chan *LDAPClientPooled, wakeup chan struct{}, result chan error) {
    for !f.closing_Load() {
        select {
        case <-wakeup:
            if f.closing_Load() {
                goto poolCleanup
            }
            active := f.activeCount_Load()
            // Heuristic calculation for required clients
            for range min(min(max(f.minPoolSize-active, 2-int32(len(pool))), f.maxPoolSize-active), max(2, (f.maxPoolSize+3)/4)) {
                if !f.tryAddPooledClient() {
                    break
                }
            }
        }
    }
    // ... cleanup logic
}
```

The pool manager uses sophisticated heuristics to determine how many additional client connections are needed. It considers the current minimum pool size requirements, immediate pool pressure (when available clients drop below 2), and respects the maximum pool size limit. The heuristics also include burst limiting to prevent overwhelming the LDAP server with too many simultaneous connection attempts.

The pool manager cannot be blocked by leasing operations because it never directly interacts with the acquire/release code paths. It operates asynchronously and uses non-blocking operations when adding new clients to the pool. Even if the pool is full when trying to add a new client, for example due to logic errors outside the pool implementation, the operation simply fails gracefully without affecting ongoing operations.

The pool manager is also responsible for handling the shutdown signal from Close() and performing complete cleanup of all active clients before signaling completion.

## 3. The Role of activeCount and closing

Atomic operations are both necessary and sufficient for managing the shared state between multiple goroutines. The `activeCount` and `closing` variables represent simple integer state that multiple goroutines need to read and modify concurrently.

```go
// These operations provide atomic read-modify-write semantics
func (f *PooledLDAPClientFactory) activeCount_Add(v int32) int32 {
    f.mu.Lock()
    defer f.mu.Unlock()
    f.activeCount += v
    return f.activeCount
}

func (f *PooledLDAPClientFactory) closing_CompareAndSwap(old, new bool) bool {
    f.mu.Lock()
    defer f.mu.Unlock()
    if f.closing == old {
        f.closing = new
        return true
    }
    return false
}
```

The ordering of operations doesn't matter in this design because each atomic operation represents a complete, independent state change. Unlike complex multi-variable updates that might require sequential consistency, these single-variable updates can occur in any order without affecting correctness.

The `activeCount` variable is only incremented by the pool manager when successfully adding a new client to the pool, and only decremented by `disposeClient` when a client is being removed from active service. This provides the critical information that `len(pool)` cannot: the total number of active connections, which is essential for enforcing the `maxPoolSize` limit.

The `closing` flag uses compare-and-swap semantics to ensure that Close() can only succeed once, preventing race conditions during shutdown.

## 4. Initialization Process

The initialization process creates one connection synchronously to validate that the pool can establish LDAP connectivity before completing initialization. This fail-fast approach prevents runtime failures that would be difficult to diagnose.

```go
func (f *PooledLDAPClientFactory) Initialize() (err error) {
    if f.pool != nil {
        return fmt.Errorf("LDAP pool is already initialized")
    }
    f.pool = make(chan *LDAPClientPooled, f.maxPoolSize)
    // ... channel setup

    if !f.tryAddPooledClient() {
        err = fmt.Errorf("LDAP pool couldn't acquire initial client")
    } else if rerr := f.ReadinessCheck(); rerr != nil {
        err = rerr
    } else {
        go f.poolManager(f.ctx, f.pool, f.wakeup, f.closed)
        return nil
    }

    // Cleanup on any failure
    err = fmt.Errorf("LDAP pool initialization failed: %w", err)
    f.closed <- err
    _ = f.Close()
    return err
}
```

If initialization fails at any step, the system uses Close() to perform complete cleanup and de-initialization, ensuring no resources are leaked. The ReadinessCheck performs an actual acquire/release cycle to verify the pool is functional.

## 5. Shutdown Process

The shutdown process uses atomic state transitions to ensure Close() can only execute once successfully, preventing race conditions during concurrent shutdown attempts.

```go
func (f *PooledLDAPClientFactory) Close() (err error) {
    if !f.closing_CompareAndSwap(false, true) {
        return nil // already closing
    }

    f.wakeupPoolManager() // signal cleanup start

    timeout := max(time.Second, 2*f.config.Timeout)
    cleanupCtx, cancelCleanup := context.WithTimeout(context.Background(), timeout)
    defer cancelCleanup()

    select {
    case cleanupErr := <-f.closed:
        err = cleanupErr
    case <-cleanupCtx.Done():
        f.cancel() // force termination
        err = fmt.Errorf("pool cleanup exceeded timeout")
    }
    // ... resource cleanup
}
```

The pool manager handles the actual cleanup by methodically draining all active clients before signaling completion through the closed channel. If cleanup takes too long, the context cancellation provides a forced termination path.

## 6. Synchronization Primitives and Safety

The design uses three primary synchronization primitives: channels for client pooling and signaling, atomic operations for shared counters and flags, and context cancellation for coordinated shutdown.

Channels provide inherently thread-safe communication with FIFO semantics that are guaranteed by the Go runtime. The pool channel ensures that clients are distributed fairly among competing goroutines, while the wakeup channel provides non-blocking signaling to the pool manager.

The atomic operations eliminate race conditions for shared state without requiring locks. These operations are implemented using the same compare-and-swap and memory barrier primitives that mutex implementations use internally, but without the contention and deadlock risks of traditional locking.

```go
func (f *PooledLDAPClientFactory) wakeupPoolManager() {
    select {
    case f.wakeup <- struct{}{}:  // non-blocking signal
    default:                      // already signaled, continue
    }
}
```

The design is safe under high load scenarios because client leasing uses lockless channel operations that scale linearly with the number of CPU cores, while the pool manager processes signals asynchronously without blocking the critical paths.

All critical code sections are defensive, using default cases in select statements and handling unexpected conditions gracefully. The system can handle logic errors such as more or fewer clients being returned than were leased without corruption or failure.

## 7. Design Assessment

### Original Design Analysis

The previous mutex-based design contained several fundamental architectural flaws that caused deadlocks and poor scalability:

```go
func (f *PooledLDAPClientFactory) acquire(ctx context.Context) (client *LDAPClientPooled, err error) {
    f.mu.Lock()           // Global lock blocks all operations
    defer f.mu.Unlock()
    // ...
    for {
        // ...
        if client, err = f.new(); err != nil {
            time.Sleep(f.sleep)   // Blocks ALL pool operations during retry
            continue
        }
        // ...
    }
}
```

The critical problems included a single global mutex that serialized all pool operations, blocking sleep operations inside locked sections that paralyzed the entire pool during connection retries, poor scalability due to all operations being serialized through the single lock, and tight coupling between pool management and client leasing that created complex interdependencies.

### Architectural Improvements

The new design provides lock-free client operations where leasing and returning clients never blocks pool management activities. Asynchronous pool management runs in the background without interfering with client operations. The system degrades gracefully under resource pressure, continuing to operate even when LDAP connections are difficult to establish. Defensive programming patterns handle edge cases without causing system-wide failures. Performance scales linearly with available CPU cores rather than being limited by lock contention.

### Production Readiness

The new design demonstrates several strengths that indicate production readiness. It uses battle-tested Go primitives including channels and atomic operations that have been proven in high-scale production systems. Comprehensive error handling provides graceful failure modes for all error conditions. Resource management ensures proper cleanup and prevents leaks. Performance characteristics show O(1) client operations with minimal contention between goroutines.

This design represents a significant improvement in both correctness and performance compared to the original implementation. The lock-free architecture eliminates the fundamental scalability and reliability issues present in the mutex-based approach while maintaining full thread safety through well-established Go concurrency primitives.

james-d-elliott:
In understanding there may be ESL issues here I want to provide as much clarity as I can.

I do not currently have time to read through everything you've mentioned and properly digest it, so I have to apologize for that. I will on the weekend within all likelihood, and will spend the time to carefully assess it all.

My current goal is to be able to reproduce this issue for a few reasons which I'm going to outline, and since you've been able to I want your cooperation with that. It's important to me due to interest in the longevity of the project that we can implement tests that prevent significant regressions. 

I don't know if it's possible to add a test preventing regression is possible in this instance but being able to reproduce it will allow me to attach a debugger and assess that specifically. If it is or a design presents itself that is, then implementing a test will prevent future changes from causing a regression which saves us valuable time later, I'm happy to take this part on.

I also want to assess if this is the most suitable way to address this based on reproduction, not that I want to say these or your other previous changes wont also be valuable additions.

I'd also like to chat with you privately at some stage if that's possible, not sure if you frequent Discord/Matrix or are on there at all.

thielj:
Testing non-deterministic code is inherently tricky.

I pushed a small update yesterday picking  changes from my dev branch that should make testing somewhat easier. Initialize() now detects mock testing (Timeout=0), creates exactly Count clients and doesn't start the goroutine. I'm sure there are better ways to do this, but all your existing tests now work with just two small changes and I'm sure this will do for some initial evaluation.

You can test lending operations fairly deterministically, even with multiple coordinated threads or by just testing the outcome after all threads have completed. The unused Metrics routine would have helped with that, but it's gone now 🤷🏼

The goroutine in itself is deterministic, too, and can be tested in various ways, both standalone and together with the lending ops. The heuristics could be factored out and tested separately. You can inject failures, e.g. connections going stale or server outages and test recovery. You can write load tests with pools of various sizes. It's a lot of work though.

Go also has some race detection but I have no further info what it does exactly.

Ultimately, keep in mind that an emulated ARM64 VM can behave significantly different from e.g. a dual socket Xeon server.

I'm mostly using WhatsApp but don't want to post my number here. Feel free to email.

thielj:
PS: I would suggest you use the dev branch for testing. Check the source code for the additional env vars used and how to enable trace logging. Inject failures by setting client lifetime to some seconds.

thielj:
By the way, before you sprinkle more log statements around potential race conditions: think about how the logger serializes multiple threads and how that turns them into some kind of rate limiting turnstiles. It's maybe just a few nanoseconds but it can easily mask "Heisenbugs".

james-d-elliott:
I'm simply looking for a way to reproduce #9936 on master so we can reopen it and/or move this forward in one direction or another.

thielj:
Congratulations. A first step into the right direction! 

Now replace the flawed design altogether because  "This can fairly reliably be replicated via an excessive number of requests" wasn't the issue the other person or I reported. Rather the opposite.

james-d-elliott:
> Now replace the flawed design altogether because "This can fairly reliably be replicated via an excessive number of requests" wasn't the issue the other person or I reported. Rather the opposite.

We would welcome any party who is interested in assisting us in replicating this issue to be involved in the process. If the issue is still present with the fix then I would love to know.

We'll be adding significant trace logging that's going to drastically simplify replication of this issue if it does indeed persist.

> wasn't the issue the other person or I reported. Rather the opposite.

Not quite sure that's the case, maybe me and @Crowley723 are missing something obvious. It seems fairly unreliable to reproduce with small numbers of LDAP requests, both of us have been using identical configurations to yourself (me for months, @Crowley723  for weeks) and failed to replicate it. I suspect there is an additional factor causing the same effect in your case, but it's going to be a bit hard to say for sure without your cooperation.

@thielj 

Please take a moment to read the Authelia Code of Conduct which is detailed [on GitHub](https://github.com/authelia/authelia?tab=coc-ov-file#readme) and [on our Webstite](https://www.authelia.com/policies/code-of-conduct/).

I have tried to give you the benefit of the doubt in regards to language barriers and have tried to extend the olive branch and a means to communicate about this despite private communication by another maintainer suggesting action was needed; but I feel like you're starting to force my hand here.

Currently you're in the correction phase of our process (step 1). I would do this privately but I have no means to do so. If you'd like to provide a means to do this for the future and/or would like me to remove this comment I am happy to cooperate with either of these things.

You are expected to conduct yourself respectfully and professionally with all members of the community. This is part of the code of conduct, which you agree to abide by when interacting with this community publicly or privately.

Your tone comes across incredibly condescending and seems generally disrespectful; I don't know if this is reflected in your emotions or if this is potentially a language barrier but you should be mindful of this.

When asked questions you're often very dismissive rather than receptive and responsive.

I don't believe these kinds of interactions are beneficial to overall positive community engagement, in fact I believe it discourages it which is the primary issue.

I look forward to constructive professional collaboration, and highly hopeful that is the case. I grimace at the idea of having to formally deal with these things.

thielj:
I offered you to email me privately or chat via WhatsApp.

Anyway, I find it very condescending when a maintainer accuses me of withholding log information, followed by a code review focusing on comments and punctuation,  requesting more information that has obviously not been read and requesting changes (that have all been provided) without the intention to ever complete the process.

Total waste of time.

PR #10663, opened 2025-11-04
fix(oidc): jwt profile claims policy hydration

james-d-elliott:
This fixes an issue where the claims policy doesn't correctly apply to the JWT Profile Access Tokens.

### DIFF
--- a/internal/authentication/types.go
+++ b/internal/authentication/types.go
@@ -352,7 +352,7 @@ func (l Level) String() string {
 type Context interface {
 	context.Context
 
-	GetProviderAuthentication() UserProvider
+	GetUserProvider() UserProvider
 	GetLogger() *logrus.Entry
 	GetClock() clock.Provider
 }
--- a/internal/middlewares/authelia_context.go
+++ b/internal/middlewares/authelia_context.go
@@ -628,7 +628,7 @@ func (ctx *AutheliaCtx) GetProviders() (providers Providers) {
 	return ctx.Providers
 }
 
-func (ctx *AutheliaCtx) GetProviderAuthentication() authentication.UserProvider {
+func (ctx *AutheliaCtx) GetUserProvider() authentication.UserProvider {
 	return ctx.Providers.UserProvider
 }
 
--- a/internal/authentication/cached.go
+++ b/internal/authentication/cached.go
@@ -84,7 +84,7 @@ func (c *CredentialCacheHMAC) check(ctx Context, username, password string, sum 
 			return &FlightResult{Cached: true, Valid: true}, nil
 		}
 
-		if valid, err = ctx.GetProviderAuthentication().CheckUserPassword(username, password); err != nil {
+		if valid, err = ctx.GetUserProvider().CheckUserPassword(username, password); err != nil {
 			return &FlightResult{Cached: false, Valid: valid}, err
 		}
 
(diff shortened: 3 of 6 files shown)


internal/authentication/types.go
```diff
@@ -283,7 +283,7 @@ func (l Level) String() string {
 type Context interface {
 	context.Context
 
-	GetUserProvider() UserProvider
+	GetProviderUser() UserProvider
```
Crowley723: I feel like this change is a little confusing, given that it doesn't match the interface name.

internal/handlers/handler_oauth2_token.go
```diff
@@ -29,53 +29,129 @@ func OAuth2TokenPOST(ctx *middlewares.AutheliaCtx, rw http.ResponseWriter, req *
...
+			"client_id":                               client.GetID(),
+			"subject":                                 session.Subject,
+			"access_token_signed_response_kid":        client.GetAccessTokenSignedResponseKeyID(),
+			"access_token_signed_response_alg":        client.GetAccessTokenSignedResponseAlg(),
+			"access_token_encrypted_response_kid":     client.GetAccessTokenEncryptedResponseKeyID(),
+			"access_token_encrypted_response_alg":     client.GetAccessTokenEncryptedResponseAlg(),
+			"access_token_encrypted_response_enc":     client.GetAccessTokenEncryptedResponseEnc(),
+			"enable_jwt_profile_oauth2_access_tokens": client.GetEnableJWTProfileOAuthAccessTokens(),
+		}).
+		Trace("Access Request is being handled by a provider")
```
Crowley723, line 55: Question: "By a provider," isn't it the authelia provider?

james-d-elliott: Yeah really wasn't sure what terminology here would be appropriate. Thinking of multi-issuer and the fact client is really not the correct terminology.

internal/authentication/types.go
james-d-elliott: In general the idea is that we should have functions that can be interfaced easily and convey category then specifics, for example `GetProviders`, `GetProviderStorage`, `GetProviderNotifier`, etc. The uniformity and sorting allow easily implement them locally.

I don't love the naming of it either. The complication is how we named the interface I think, unlike almost all others where we use `Provider` i.e. `storage.Provider`, `notification.Provider`, `ntp.Provider`, `metrics.Provider`, `session.Provider`, `templates.Provider`, `random.Provider`. Whereas this one is `authentication.UserProvider`; maybe it'd make more sense to move to `authentication.Provider` too?

There are obviously going to be notable exceptions to this and already are. But I think we should aim for this kind of uniformity as we grow and make things more complex. Some notable expected examples are the expression package which is likely to have several types of providers potentially.

In general my aim is to make the DX better by naming things predictably.

If we have a specific end-goal I am happy to name it as such with the intention to refactor the underlying interface name later.

Crowley723 (dismissed):

internal/middlewares/authelia_context.go
```diff
@@ -627,10 +628,14 @@ func (ctx *AutheliaCtx) GetProviders() (providers Providers) {
 	return ctx.Providers
 }
 
-func (ctx *AutheliaCtx) GetUserProvider() (provider authentication.UserProvider) {
+func (ctx *AutheliaCtx) GetProviderAuthentication() authentication.UserProvider {
```
Crowley723: Is it worth it to move the rename of the `authentication.UserProvider` to a different pr? Then we could change it to `authentication.Provider` as well?

autheliabot (approved):

PR #10793, opened 2025-11-20
fix(configuration): defaults not applied to server authz

james-d-elliott:
This fixes an edge case issue where the server authz endpoint defaults are not applied.


nightah (approved):
LGTM.

autheliabot (approved):

PR #10827, opened 2025-11-26
Add missing options for Homarr config

jgillman:
consent_mode

OIDC with Homarr wasn't working for me until I added the `consent_mode: 'implicit'` option. [Homarr's OIDC documentation](https://homarr.dev/docs/next/advanced/single-sign-on/) includes this in their documentation for OIDC with Authelia.


james-d-elliott:
Thanks for choosing to contribute but I'm outright rejecting this change. This option only affects the frontend flow on the Authelia side, and does not affect any underlying part of OIDC that Homarr can expect (i.e. no documented part of the spec). 

I'd recommend opening a [discussion](https://github.com/authelia/authelia/discussions) to discuss the actual issue. If solely making this change fixes it it's likely a bug, either with Authelia or more likely Homarr.

In any event we're not planning on documenting consent modes in the client docs.

PR #10896, opened 2025-12-05
docs(oidc): update pangolin

moedtm:
For a working integration with Pangolin role mapping (which is needed for basic functionality), the groups scope is needed by default. so it's best to add this to the example conf.
(To make it compatible with the suggested default role mapping expression in Pangolin: contains(groups, 'admin') && 'Admin' || 'Member)


james-d-elliott (changes requested):
This was previously not needed'; you'll need to update the versions as well. You also have to update the scopes that Pangolin requests.. if it is not requesting the scopes it has no effect. I suspect there is a claim mapping that must occur as well on their end?

moedtm:
I'm no OIDC expert, but I got this working by adding `groups` to the scopes list in the authelia client config.

```yaml
scopes:
          - 'openid'
          - 'profile'
          - 'email'
          - 'groups'
```

Combined with the claims policy config from the Pangolin guide:

```yaml
claims_policies:
  pangolin:
    id_token: ['rat', 'groups', 'email', 'email_verified', 'alt_emails', 'preferred_username', 'name']
```

The catch is that Pangolin's community/free edition doesn't support auto-provisioning for role mapping—you have to set it up manually. The suggested expression is `contains(groups, 'admin') && 'Admin' || 'Member'`, but this only works if `groups` is in your Authelia client's scopes. Otherwise the `contains()` function has nothing to evaluate.

autheliabot (approved):

james-d-elliott (changes requested):
Please update the version for Pangolin that's noted in this guide.

PR #10898, opened 2025-12-08
Fix Jellyfin OpenID Connect configuration options

n0rad:
following https://github.com/9p4/jellyfin-plugin-sso/issues/313#issue-3554624765 and https://github.com/9p4/jellyfin-plugin-sso/issues/264#issuecomment-2870003549

### DIFF
--- a/docs/content/integration/openid-connect/clients/jellyfin/index.md
+++ b/docs/content/integration/openid-connect/clients/jellyfin/index.md
@@ -156,8 +156,10 @@ To configure [Jellyfin] to utilize Authelia as an [OpenID Connect 1.0] Provider,
           <OidScopes>
             <string>groups</string>
           </OidScopes>
+          <SchemeOverride>https</SchemeOverride>
           <CanonicalLinks></CanonicalLinks>
           <DisableHttps>false</DisableHttps>
+          <DisablePushedAuthorization>true</DisablePushedAuthorization>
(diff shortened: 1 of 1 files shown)


autheliabot (approved):

james-d-elliott (changes requested):
This is a workaround for a bug in the plugin which prevents it from supporting the requirements noted in [Section 2.1](https://datatracker.ietf.org/doc/html/rfc9126#section-2.1) of the spec (see quoted text below). This must be noted along with an open issue on the plugin side.

> The request also includes, as appropriate for the given client, any additional parameters necessary for client authentication (e.g., client_secret or client_assertion and client_assertion_type). Such parameters are defined and registered for use at the token endpoint but are applicable only for client authentication. When present in a pushed authorization request, they are relied upon only for client authentication and are not germane to the authorization request itself.

> The client constructs the message body of an HTTP POST request with parameters formatted with x-www-form-urlencoded using a character encoding of UTF-8, as described in [Appendix B](https://www.rfc-editor.org/rfc/rfc6749#appendix-B) of [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)]. If applicable, the client also adds its authentication credentials to the request header or the request body using the same rules as for token endpoint requests.

Specifically the pushed authorization request MUST use the same authentication mechanism as the token endpoint, which it's not doing.

james-d-elliott:
If the requested changes are made this can be reopened.

PR #10976, opened 2025-12-21
wip: Add Kerberos authentication support via SPNEGO

MatthieuCoder:
Hello ! 

This is a PR adding initial spnego (via Kerberos) authentication support,

This is currently very much work in progress - but I'd like to have some feedback regarding my implementation into authelia.


internal/authorization/amr.go
```diff
@@ -38,6 +38,7 @@ type AuthenticationMethodsReferences struct {
 	WebAuthnSoftware             bool
 	WebAuthnUserPresence         bool
 	WebAuthnUserVerified         bool
+	Kerberos                     bool
```
Crowley723, line 43: I am going to point out that we probably want to define Kerberos using the amr `wia` as defined (indirectly) in [RFC8176](https://datatracker.ietf.org/doc/html/rfc8176) (The linked reference can be seen here: [Web archive link](https://web.archive.org/web/20150609232805/http://blogs.msdn.com/b/benjaminperkins/archive/2011/09/14/iis-integrated-windows-authentication-with-negotiate.aspx))

internal/middlewares/authelia_context.go
```diff
@@ -688,6 +693,37 @@ func (ctx *AutheliaCtx) GetWebAuthnProvider() (w *webauthn.WebAuthn, err error)
...
+
+	host, err := types.GetHostAddress(ctx.RemoteAddr().String())
+
+	if err == nil {
+		spnegoService = spnego.SPNEGOService(kt, service.ClientAddress(host))
+		return
+	} else {
+		spnegoService = spnego.SPNEGOService(kt)
+		return
+	}
```
Crowley723: I think we should aim to use explicit returns rather than assignments to the named return values.
```suggestion
	if err != nil {
		return spnego.SPNEGOService(kt), nil
	}
	
	return spnego.SPNEGOService(kt, service.ClientAddress(host)), nil
```

james-d-elliott:
So mostly looks fine except one thing which I think I can solve. Thanks for choosing to contribute. Does this currently function? Have you tested it?

As to the one thing. That dependency worries me. Probably the best illustration of why is the [latest release](https://github.com/jcmturner/gokrb5/releases/tag/v8.4.4).

For starters it's not maintained. Secondly it uses a fork of go's X repositories which have known CVE's (we could sift through and see which ones are vulnerable but for reasons I'll get into this makes no logical sense), and clearly from that release above they took years to resolve the previous ones by updating their fork (which should be a very quick task based on the minimal modifications so it's clear the dev has just given up on the project also figuring out which ones are vulnerable or not is not practical as forking it ourselves and maintaining these fixes would be a better outcome).

I think the solution to this is an annoying one that I've done recently in similar situations where I just fork the dep unless we find something maintained.

MatthieuCoder:
> So mostly looks fine except one thing which I think I can solve. Thanks for choosing to contribute. Does this currently function? Have you tested it?

I haven't tested it yet, I mainly need to integrate it with the web site of things, I'm currently setting up an environment for testing. But as I said earlier this is really early.

> As to the one thing. That dependency worries me. Probably the best illustration of why is the [latest release](https://github.com/jcmturner/gokrb5/releases/tag/v8.4.4).
> 
> For starters it's not maintained. Secondly it uses a fork of go's X repositories which have known CVE's (we could sift through and see which ones are vulnerable but for reasons I'll get into this makes no logical sense), and clearly from that release above they took years to resolve the previous ones by updating their fork (which should be a very quick task based on the minimal modifications so it's clear the dev has just given up on the project also figuring out which ones are vulnerable or not is not practical as forking it ourselves and maintaining these fixes would be a better outcome).

I know about this problem, but given that this dependency is used in other well-known projects, (ex. in [OpenBAO](https://github.com/openbao/openbao/blob/main/builtin/credential/kerberos/path_login.go#L16)) but ultimately I couldn't find an other kerberos/spnego implementation without requiring C libs. 

> I think the solution to this is an annoying one that I've done recently in similar situations where I just fork the dep unless we find something maintained.

Most likely. A fork of this library would benefit the whole ecosystem.

MatthieuCoder:
> Looks good so far, just a couple things I wanted to highlight.
> 
> Thanks for choosing to contribute, I would recommend that before you get too far that you join either Discord or Matrix and discuss next steps. This is so that we can guide development in the right direction.

Sure thing !

internal/authorization/amr.go
MatthieuCoder: Regrding how to treat spnego in RFC8176 - would it be appropriate to classify as a ChannelService (like Duo), because in itself, the kerberos authentication requires the requesting user to have a keytab (which is not handled by the browser)

internal/middlewares/authelia_context.go
MatthieuCoder: should be good now - although I'm not sure about if my "goto" is accepted as a coding standard.

MatthieuCoder:
I tested this using my FreeIPA instance and was successfully able to log-in using my Firefox instance
image

And by visiting manually /authenticated, the correct state is returned by `/state`:

image

Regarding the front-end, would an "automatic" request to /spnego be enough to make it '0-click', or do we want a "Login with Kerberos" kind of button ?

internal/authorization/amr.go
```diff
@@ -47,7 +51,7 @@ func (r AuthenticationMethodsReferences) FactorKnowledge() bool {
 
 // FactorPossession returns true if a "something you have" factor of authentication was used.
 func (r AuthenticationMethodsReferences) FactorPossession() bool {
-	return r.TOTP || r.Duo || r.WebAuthn || r.WebAuthnHardware || r.WebAuthnSoftware
+	return r.TOTP || r.Duo || r.WebAuthn || r.WebAuthnHardware || r.WebAuthnSoftware || r.Kerberos
```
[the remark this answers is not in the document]
MatthieuCoder: I will implement tests at a later stage, this is a bit too early

internal/middlewares/authelia_context.go
```diff
@@ -688,6 +693,45 @@ func (ctx *AutheliaCtx) GetWebAuthnProvider() (w *webauthn.WebAuthn, err error)
...
+
+	return nil, fmt.Errorf("unable to create SPNEGO service: principal '%s' not found in keytab file '%s'", ctx.Configuration.SPNEGO.Principal, keyTabFile)
+
+PrincipalFound:
+
+	host, err := types.GetHostAddress(ctx.RemoteAddr().String())
+
+	if err == nil {
+		return spnego.SPNEGOService(kt, service.DecodePAC(false), service.ClientAddress(host), service.KeytabPrincipal(ctx.Configuration.SPNEGO.Principal)), nil
```
MatthieuCoder: I will add an option to disable PAC decoding

MatthieuCoder:
I added a minimal frontend to make the request to login using the spnego endpoint, it looks ugly if you ask me, but it works!

image

internal/authorization/amr.go
Crowley723: I would say no, it should probably be "Browser". Duo is considered a different channel because the authentication happens out-of-band from the clients requests to the auth service. 
Ie:
- user logs in with password and initiates duo as a second factor
- auth server says "hey duo, send a push to this user's device and tell me if its accepted"
- user accepts the push on their phone
- duo contacts auth server to say its done.

The connection/communication between Duo and Authelia is out-of-band or backchannel (doesn't involve the user's device as a means of communication)

This is fundamentally different than Kerberos.
- Client gets ticket from KDC
- Client sends ticket to auth server via http
- auth server validates the ticket **locally** using keytab configured during **setup**

There is no backchannel communication required to validate the ticket which is why it wouldn't be considered a channel serivce (with some [exceptions](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-pac/c38cc307-f3e6-4ed4-8c81-dc550d96223c)), although I understand the confusion.

internal/middlewares/authelia_context.go
Crowley723: I think we should try to avoid `goto` if possible.
It think it should be something like:
```
found := false
for item in items {
	if item.value = value {
		found = true
		break
	} 
}

if !found {
	return error
}

//other stuff
```

I find this much easier to read.

I tend to do error handling using `if err != nil` rather than `if err == nil`, it leads to a lower cyclomatic complexity (less branches), more readable code, and its much easier to test.

Crowley723:
> I added a minimal frontend to make the request to login using the spnego endpoint, it looks ugly if you ask me, but it works!

I don't think it looks terrible, glad you got it working!


We have had a number of requests to add additional ways to authenticate (or show more information on the login screen).

I like the idea of having a "primary button" that shows the primary method (probably "Sign in with a passkey") then another button below it that has "More Methods" which could expand or open a dialog with more login options. Something like this will have to be done when we do custom authn policies (to allow for multiple first factor methods), but its probably out of scope for this pr. @james-d-elliott any comments on the matter?

web/src/views/LoginPortal/FirstFactor/FirstFactorForm.tsx
```diff
@@ -161,6 +164,49 @@ const FirstFactorForm = function (props: Props) {
...
+        requestMethod,
+        flowID,
+        flow,
+        subflow,
+        userCode,
+        loginChannel,
+        createErrorNotification,
+        translate,
+        focusPassword,
+    ]);
```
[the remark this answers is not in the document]
Crowley723: OI listen up, we don't want user's to think they can login with a password AND kerberos. Unified state is the way to go!

nightah:
I think if SPNEGO is supported and enabled we should provide the means via a toggle to automatically attempt login via said method.

It begs the question as to what sort of qualifying parameters could we couple with the automated login, if any at all.

If my memory and understanding is correct in most cases, maybe even all SPNEGO happens almost silently.

internal/authentication/spnego_keytab_provider.go
```diff
@@ -0,0 +1,56 @@
...
+
+	kt, err := keytab.Load(keyTabFile)
+	if err != nil {
+		return false, err
+	}
+
+	p.kt = kt
+
+	return true, nil
+}
```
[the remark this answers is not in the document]
MatthieuCoder: this is not a problem since SPNEGOKeytabProvider is only created if enabled in the config

MatthieuCoder:
> I think if SPNEGO is supported and enabled we should provide the means via a toggle to automatically attempt login via said method.
> 
> It begs the question as to what sort of qualifying parameters could we couple with the automated login, if any at all.
> 
> If my memory and understanding is correct in most cases, maybe even all SPNEGO happens almost silently.

All of SPNEGO can happen without user interaction - even before showing the UI - I implemented it this way to be able to have the "remember me" parameter

internal/configuration/schema/spnego.go
```diff
@@ -0,0 +1,7 @@
+package schema
+
+type SPNEGO struct {
+	Disable   bool   `koanf:"disable" json:"disable" jsonschema:"default=false,title=Disable" jsonschema_description:"Disables the WebAuthn 2FA functionality."`
+	Keytab    string `koanf:"keytab" json:"keytab" jsonschema:"title=File Path" jsonschema_description:"The filepath to the kerberos keytab."`
+	Principal string `koanf:"principal" json:"principal" jsonschema:"title=Principal" jsonschema_description:"The service principal name."`
```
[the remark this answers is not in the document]
MatthieuCoder: This is now validated

internal/authorization/amr.go
MatthieuCoder: Indeed, Kerberos is now `ChannelBrowser`

nightah:
> All of SPNEGO can happen without user interaction - even before showing the UI - I implemented it this way to be able to have the "remember me" parameter

Yep I understand but it's probably safe to assume that there could be mixed use cases where SPNEGO is enabled. For example there could be access to resources protected by Authelia that aren't necessarily initiated on a domain connected machine via Kerberos.

This is what I meant by if we want to enable a config toggle for automatic SPNEGO authentication we likely would want some qualifying parameters.

Maybe we query the Windows access token to determine if the automatic login should occur?

It's probably worth discussing this on Matrix/Discord so we can agree on some key principals before moving forward.

MatthieuCoder:
> It's probably worth discussing this on Matrix/Discord so we can agree on some key principals before moving forward.

I couldn't get matrix working but I'm @matthieu:puffer.fish, I'm also on the authelia Discord.

internal/authorization/amr.go
```diff
@@ -21,6 +21,9 @@ func NewAuthenticationMethodsReferencesFromClaim(claim []string) (amr Authentica
 			amr.WebAuthnSoftware = true
 		case AMRUserPresence:
 			amr.WebAuthnUserVerified = true
+		case AMRWindowsIntegratedAuthentication:
+			// Kerberos is the only WIA method Authelia supports currently.
```
james-d-elliott:
```suggestion
```

internal/configuration/schema/spnego.go
```diff
@@ -0,0 +1,8 @@
+package schema
+
+type SPNEGO struct {
+	Disable   bool   `koanf:"disable" json:"disable" jsonschema:"default=false,title=Disable" jsonschema_description:"Disables the SPNEGO authentication functionality."`
```
james-d-elliott:
```suggestion
	Enabled   bool   `koanf:"enabled" json:"enabled" jsonschema:"default=false,title=Enabled" jsonschema_description:"Enables the SPNEGO authentication functionality."`
```

Should be explicitly opt-in. Adjust this and the related code/docs.

internal/handlers/handler_firstfactor_spnego.go
```diff
@@ -0,0 +1,267 @@
...
+	spnegoNegTokenRespIncompleteKRB5 = "Negotiate oRQwEqADCgEBoQsGCSqGSIb3EgECAg==" // #nosec
+
+	// HTTPHeaderAuthRequest is the header that will hold authn/z information.
+	HTTPHeaderAuthRequest = "Authorization"
+	// HTTPHeaderAuthResponse is the header that will hold SPNEGO data from the server.
+	HTTPHeaderAuthResponse = "WWW-Authenticate"
+	// HTTPHeaderAuthResponseValueKey is the key in the auth header for SPNEGO.
+	HTTPHeaderAuthResponseValueKey = "Negotiate"
+	ctxCredentials                 = "github.com/jcmturner/gokrb5/v8/ctxCredentials" // #nosec
+)
```
james-d-elliott, line 32: Most of these look like  test constants and constants that already exist i.e. fasthttp.HeaderAuthorization etc.

internal/authentication/spnego_keytab_provider.go
```diff
@@ -0,0 +1,56 @@
...
+	config *schema.SPNEGO
+	kt     *keytab.Keytab
+	mutex  sync.Mutex
+}
+
+func NewSPNEGOKeytabProvider(config *schema.SPNEGO) (*SPNEGOKeytabProvider, error) {
+	keyTabFile := "/etc/krb5.keytab"
+	if config.Keytab != "" {
+		keyTabFile = config.Keytab
+	}
```
james-d-elliott:
```suggestion
```

We wouldn't configure a default keytab. Users will have to explicitly configure it.

internal/authentication/spnego_keytab_provider.go
```diff
@@ -0,0 +1,56 @@
...
+
+	kt, err := keytab.Load(keyTabFile)
```
james-d-elliott:
```suggestion
	kt, err := keytab.Load(config.Keytab)
```

internal/configuration/validator/spnego.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+
+	"github.com/authelia/authelia/v4/internal/configuration/schema"
+)
+
+func ValidateSpnego(config *schema.Configuration, validator *schema.StructValidator) {
+	if config.SPNEGO.Disable == true {
+		return
+	}
+	if config.AuthenticationBackend.LDAP == nil {
```
james-d-elliott:
```suggestion
	if !config.SPNEGO.Enabled {
		return
	}
	
	if len(config.Keytab) == 0 {
	    validator.Push(fmt.Errorf("spnego: option 'keytab' must be provided"))
	}
	
	if config.AuthenticationBackend.LDAP == nil {
```

Example validation adjustments. Essentially if the keytab is not configured we'd fail configuration validation.

internal/handlers/handler_firstfactor_spnego.go
```diff
@@ -0,0 +1,267 @@
...
+		respondUnauthorized(ctx, messageAuthenticationFailed)
+
+		ctx.Logger.WithError(err).Errorf(logFmtErrParseRequestBody, regulation.AuthTypeKerberos)
+
+		return
+	}
+
+	header, err := parseAuthorizationHeader(ctx.Request.Header.Peek(HTTPHeaderAuthRequest))
+	if err != nil {
+		// This NEEDS to be a 401 to trigger the client to send the ticket.
```
james-d-elliott:
```suggestion
```

internal/handlers/handler_firstfactor_spnego.go
```diff
@@ -0,0 +1,267 @@
...
+		ctx.Logger.WithError(fmt.Errorf("missing or malformed %s header", HTTPHeaderAuthRequest)).Errorf(logFmtErrKerberosAuthenticationChallengeGenerate, errStrReqHeaderParse)
+
+		doMarkAuthenticationAttempt(ctx, false, regulation.NewBan(regulation.BanTypeNone, "", nil), regulation.AuthTypeKerberos, nil)
+
+		return
+	}
+
+	token, err := parseSPNEGOToken(header)
+	if err != nil {
+		// This NEEDS to be a 401 to trigger the client to send the ticket.
```
james-d-elliott:
```suggestion
```

internal/middlewares/util.go
```diff
@@ -68,6 +68,12 @@ func NewProviders(config *schema.Configuration, caCertPool *x509.CertPool) (prov
 		providers.Metrics = metrics.NewPrometheus()
 	}
 
+	if !config.SPNEGO.Disable {
+		if providers.SPNEGOProvider, err = authentication.NewSPNEGOKeytabProvider(&config.SPNEGO); err != nil {
+			errs = append(errs, err)
+		}
+	}
```
james-d-elliott:
```suggestion
	if config.SPNEGO.Enabled {
		if providers.SPNEGOProvider, err = authentication.NewSPNEGOKeytabProvider(&config.SPNEGO); err != nil {
			errs = append(errs, err)
		}
	}
```

In addition move the error checking to a startup check. If you're unsure let me know and I'll show you an example.

internal/server/handlers.go
```diff
@@ -332,6 +332,11 @@ func handlerMain(config *schema.Configuration, providers middlewares.Providers)
 		r.DELETE("/api/secondfactor/webauthn/credential/{credentialID}", middlewareElevated1FA(handlers.WebAuthnCredentialDELETE))
 	}
 
+	// SPNEGO KRB5 Authentication related Endpoint.
```
james-d-elliott:
```suggestion
```

james-d-elliott (changes requested):
Few general changes.

james-d-elliott:
I think my main thoughts on the nature of automatic login that I'm thinking about this has been mostly articulated. I can't think of a single instance where a user manually triggered SPNEGO. The reality is most users will not know what that button does, and it will just confuse them generally; more so when it fails.

Doing it silently in the background if available without showing an error to the user is probably the most logical first step. Maybe we can show an error if their machine attempts to perform SPNEGO but there is a failure, but if their machine doesn't perform negociation then they don't need to know about it.

I think if we want to have a manual option that it's an option we add after the fact in a separate PR. To handle the remember me element we could also offer that option after they perform SPNEGO and remember me is enabled.

We also need to consider the authz use case. Hopefully the flow translates fairly well, but I suspect it may be impossible. I suspect since the negociation has to occur using a 401 we may not be able to.. but it's a worthy consideration.. I suspect the main issue will be ensuring the session persists as I suspect the token will not be presented on all subsequent requests, only the single follow up.

internal/configuration/validator/spnego.go
```diff
@@ -0,0 +1,28 @@
+package validator
+
+import (
+	"fmt"
+
+	"github.com/authelia/authelia/v4/internal/configuration/schema"
+)
+
+func ValidateSpnego(config *schema.Configuration, validator *schema.StructValidator) {
+	if config.SPNEGO.Disable == true {
```
autheliabot: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>S1002: should omit comparison to bool constant, can be simplified to config.SPNEGO.Disable (staticcheck)

internal/authentication/spnego_keytab_provider.go
```diff
@@ -0,0 +1,56 @@
...
+	return &SPNEGOKeytabProvider{
+		config: config,
+		kt:     kt,
+	}, nil
+}
+
+func (p *SPNEGOKeytabProvider) GetKeytab() (*keytab.Keytab, error) {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+	return p.kt, nil
```
autheliabot: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>missing whitespace above this line (too many lines above return) (wsl_v5)

internal/configuration/validator/spnego.go
autheliabot: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>missing whitespace above this line (invalid statement above if) (wsl_v5)

internal/middlewares/authelia_context.go
```diff
@@ -688,6 +691,21 @@ func (ctx *AutheliaCtx) GetWebAuthnProvider() (w *webauthn.WebAuthn, err error)
...
 }
 
+func (ctx *AutheliaCtx) GetSPNEGOProvider() (*spnego.SPNEGO, error) {
+	kt, err := ctx.Providers.SPNEGOProvider.GetKeytab()
+	if err != nil {
+		return nil, err
+	}
+
+	host, err := types.GetHostAddress(ctx.RemoteAddr().String())
+
```
autheliabot: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>unnecessary whitespace (err) (wsl_v5)

internal/authentication/spnego_keytab_provider.go
```diff
@@ -0,0 +1,48 @@
...
+	kt, err := keytab.Load(config.Keytab)
+	if err != nil {
+		return nil, err
+	}
+
+	return &SPNEGOKeytabProvider{
+		config: config,
+		kt:     kt,
+	}, nil
+}
```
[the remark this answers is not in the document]
MatthieuCoder: this is ensured by the schema validation - and since `NewSPNEGOKeytabProvider` is only called when `config.SPNEGO.Enabled` is truthy, this is a non-issue

MatthieuCoder:
I drafted a PR on my tree to add LDAP Search for principals: https://github.com/MatthieuCoder/authelia/pull/1

internal/authentication/file_user_provider.go
```diff
@@ -245,7 +245,7 @@ func NewFileCryptoHashFromConfig(config schema.AuthenticationBackendFilePassword
 		hash, err = argon2.New(
 			argon2.WithVariantName(config.Argon2.Variant),
 			argon2.WithT(config.Argon2.Iterations),
-			argon2.WithM(uint32(config.Argon2.Memory)), //nolint:gosec // Validated at runtime.
+			argon2.WithM(uint32(config.Argon2.Memory)),
```
Crowley723: We should make sure this stays in.

internal/commands/storage_run.go
```diff
@@ -1417,7 +1417,7 @@ func runStorageUserTOTPGenerate(ctx context.Context, w io.Writer, store storage.
 
 	totpProvider := totp.NewTimeBasedProvider(config.TOTP)
 
-	if c, err = totpProvider.GenerateCustom(totp.NewContext(ctx, &clock.Real{}, &random.Cryptographical{}), username, config.TOTP.DefaultAlgorithm, secret, uint32(config.TOTP.DefaultDigits), uint(config.TOTP.DefaultPeriod), uint(config.TOTP.SecretSize)); err != nil { //nolint:gosec // Validated at runtime.
```
Crowley723, line 1421: Same thing here, lets not remove the nolint comment

internal/configuration/validator/authentication.go
```diff
@@ -131,7 +131,7 @@ func validateFileAuthenticationBackendPasswordConfigArgon2(config *schema.Authen
 		config.Argon2.Memory = schema.DefaultPasswordConfig.Argon2.Memory
 	case config.Argon2.Memory < argon2.MemoryMin:
 		validator.Push(fmt.Errorf(errFmtFileAuthBackendPasswordOptionTooSmall, hashArgon2, "memory", config.Argon2.Memory, argon2.MemoryMin))
-	case uint64(config.Argon2.Memory) > uint64(argon2.MemoryMax): //nolint:gosec // Validated at runtime.
```
Crowley723: Ditto

internal/configuration/config.template.yml
```diff
@@ -1663,4 +1663,19 @@ session:
               # -----BEGIN CERTIFICATE-----
               # ...
               # -----END CERTIFICATE-----
+
+# SPNEGO:
+  ## Disable SPNEGO authentication.
+  # enabled: true
```
Crowley723: The template should generally reflect the default values for config options. 
```suggestion
  # enabled: false
```

internal/handlers/func_test.go
```diff
@@ -78,5 +78,5 @@ func getStepTOTP(ctx *middlewares.AutheliaCtx, period int) uint64 {
 		period = ctx.Configuration.TOTP.DefaultPeriod
 	}
 
-	return uint64(int(step) / period) //nolint:gosec // This is a testing func.
```
Crowley723, line 81: Same here

internal/handlers/handler_register_totp.go
```diff
@@ -95,7 +95,7 @@ func TOTPRegisterPUT(ctx *middlewares.AutheliaCtx) {
 
 	var config *model.TOTPConfiguration
 
-	if config, err = ctx.Providers.TOTP.GenerateCustom(ctx, userSession.Username, bodyJSON.Algorithm, "", uint32(bodyJSON.Length), uint(bodyJSON.Period), 0); err != nil { //nolint:gosec // Validated at runtime.
```
Crowley723, line 98: Same here

internal/handlers/handler_register_totp_test.go
```diff
@@ -145,8 +145,8 @@ func TestTOTPRegisterPUT(t *testing.T) {
 						Return(*totp.NewTOTPOptionsFromSchema(mock.Ctx.Configuration.TOTP)),
 					mock.TOTPMock.
 						EXPECT().
-						GenerateCustom(mock.Ctx, testUsername, mock.Ctx.Configuration.TOTP.DefaultAlgorithm, "", uint32(mock.Ctx.Configuration.TOTP.DefaultDigits), uint(mock.Ctx.Configuration.TOTP.DefaultPeriod), uint(0)).                                      //nolint:gosec
-						Return(&model.TOTPConfiguration{Username: testUsername, Algorithm: mock.Ctx.Configuration.TOTP.DefaultAlgorithm, Digits: uint32(mock.Ctx.Configuration.TOTP.DefaultDigits), Period: uint(mock.Ctx.Configuration.TOTP.DefaultPeriod)}, nil), //nolint:gosec
```
Crowley723, line 149: Same here

internal/handlers/handler_register_totp_test.go
```diff
@@ -251,7 +251,7 @@ func TestTOTPRegisterPUT(t *testing.T) {
 						Return(*totp.NewTOTPOptionsFromSchema(mock.Ctx.Configuration.TOTP)),
 					mock.TOTPMock.
 						EXPECT().
-						GenerateCustom(mock.Ctx, testUsername, mock.Ctx.Configuration.TOTP.DefaultAlgorithm, "", uint32(mock.Ctx.Configuration.TOTP.DefaultDigits), uint(mock.Ctx.Configuration.TOTP.DefaultPeriod), uint(0)). //nolint:gosec
```
Crowley723, line 254: Same here

internal/server/handlers.go
```diff
@@ -78,7 +78,7 @@ func handleError(cpath string) func(ctx *fasthttp.RequestCtx, err error) {
 			if matches := reTLSRequestOnPlainTextSocketErr.FindStringSubmatch(err.Error()); len(matches) == 3 {
 				if version, verr := utils.TLSVersionFromBytesString(matches[1] + matches[2]); verr == nil && version != -1 {
 					statusCode = fasthttp.StatusBadRequest
-					message = fmt.Sprintf(errFmtMessageServerTLSVersion, tls.VersionName(uint16(version))) //nolint:gosec // This conversion is safe as the only versions potentially returned are from the crypto/tls pkg.
```
Crowley723, line 81: Same here

internal/suites/action_totp.go
```diff
@@ -139,7 +139,7 @@ func (rs *RodSession) doRegisterTOTPAdvanced(t *testing.T, page *rod.Page, inval
 	}
 
 	credential.ValidationOptions = totp.ValidateOpts{
-		Period:    uint(uperiod), //nolint:gosec // This is a test function.
```
Crowley723, line 142: Same here

internal/totp/totp.go
```diff
@@ -17,9 +17,9 @@ func NewTimeBasedProvider(config schema.TOTP) (provider *TimeBased) {
 		opts:      NewTOTPOptionsFromSchema(config),
 		issuer:    config.Issuer,
 		algorithm: config.DefaultAlgorithm,
-		digits:    uint32(config.DefaultDigits), //nolint:gosec // Validated at runtime.
-		period:    uint(config.DefaultPeriod),   //nolint:gosec // Validated at runtime.
-		size:      uint(config.SecretSize),      //nolint:gosec // Validated at runtime.
```
Crowley723, line 22: Same here

internal/totp/totp_test.go
```diff
@@ -123,7 +123,7 @@ func TestTOTPGenerateCustom(t *testing.T) {
 				assert.Equal(t, tc.digits, c.Digits)
 				assert.Equal(t, tc.algorithm, c.Algorithm)
 
-				expectedSecretLen := int(tc.secretSize) //nolint:gosec
```
Crowley723, line 126: Same here

internal/utils/time.go
```diff
@@ -149,7 +149,7 @@ func matchParseTimeStringWithLayouts(input string, layouts []string) (index int,
 // UnixNanoTimeToMicrosoftNTEpoch converts a unix timestamp in nanosecond format to win32 epoch format.
 func UnixNanoTimeToMicrosoftNTEpoch(nano int64) (t uint64) {
 	if nano >= 0 {
-		return uint64(nano/100) + timeUnixEpochAsMicrosoftNTEpoch //nolint:gosec // This is a gated condition and is checked,
```
Crowley723: Same here

web/src/views/LoginPortal/FirstFactor/FirstFactorForm.tsx
```diff
@@ -383,6 +430,31 @@ const FirstFactorForm = function (props: Props) {
...
+                            <Button
+                                id="spnego-login-button"
+                                variant="contained"
+                                color="primary"
+                                fullWidth={true}
+                                endIcon={loading ? <CircularProgress size={20} /> : null}
+                                disabled={disabled}
+                                onClick={spnegoLogin}
+                            >
+                                {translate("Login with SPNEGO")}
```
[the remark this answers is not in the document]
Crowley723: Maybe "Login with Kerberos"?

Crowley723 (changes requested):
A couple nitpicks, mostly issues with removing `nolint` directives.

internal/authentication/file_user_provider.go
MatthieuCoder: I don't really understand *what* removed or *why* those got removed, I'll fix it tomorrow :/

Crowley723: No worries, it happens.

internal/middlewares/authelia_context.go
```diff
@@ -688,6 +691,21 @@ func (ctx *AutheliaCtx) GetWebAuthnProvider() (w *webauthn.WebAuthn, err error)
...
+	if err != nil {
+		return spnego.SPNEGOService(kt, service.KeytabPrincipal(ctx.Configuration.SPNEGO.Principal)), nil
+	}
```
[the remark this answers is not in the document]
MatthieuCoder: such an error would be a parsing error between fasthttp and gokrb5, which is very unlikely and shouldn't theoretically happen

internal/configuration/validator/authentication.go
```diff
@@ -437,6 +437,10 @@ func setDefaultImplementationLDAPAuthenticationBackendProfileAttributes(config *
 		config.UsersFilter = implementation.UsersFilter
 	}
 
+	if ldapImplementationShouldSetStr(config.PrincipalsFilter, implementation.PrincipalsFilter) {
+		config.PrincipalsFilter = implementation.PrincipalsFilter
+	}
+
```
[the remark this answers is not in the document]
MatthieuCoder: I moved the check to the spnego validator - not ideal but should do the trick given that the spnego config isn't available in the "validateLDAPRequiredParameters" parameters

MatthieuCoder:
in my todo: 

- [ ] Check the realm matches the one in the config

internal/authentication/ldap_user_provider_lifecycle.go
```diff
@@ -152,6 +152,98 @@ func (p *LDAPUserProvider) parseDynamicUsersConfiguration() {
 		ldapPlaceholderInput, p.usersFilterReplacementInput)
 }
 
+func (p *LDAPUserProvider) parseDynamicPrincipalsConfiguration() {
```
autheliabot, line 155: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>cyclomatic complexity 19 of func `(*LDAPUserProvider).parseDynamicPrincipalsConfiguration` is high (> 15) (gocyclo)

internal/configuration/validator/spnego.go
```diff
@@ -0,0 +1,48 @@
...
+		validator.Push(fmt.Errorf("spnego: option 'realm' must be provided"))
+	}
+
+	if config.AuthenticationBackend.LDAP == nil {
+		validator.Push(fmt.Errorf("spnego: kerberos authentication requires LDAP backend to be configured"))
+	}
+
+	if config.AuthenticationBackend.LDAP.PrincipalsFilter == "" {
+		validator.Push(fmt.Errorf(errFmtLDAPAuthBackendMissingOption, "principals_filter"))
+	} else {
```
autheliabot, line 33: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>unnecessary leading newline (whitespace)

internal/configuration/validator/spnego.go
```diff
@@ -0,0 +1,48 @@
...
+
```
autheliabot, line 34: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>unnecessary whitespace (leading-whitespace) (wsl_v5)

internal/handlers/handler_firstfactor_spnego.go
```diff
@@ -0,0 +1,263 @@
...
+
+		return
+	}
+
+	var (
+		userDetails  *authentication.UserDetails
+		kerberosUser = context.Value(ctxCredentials).(*credentials.Credentials)
+		username     = kerberosUser.UserName()
+	)
+	userProvider := ctx.Providers.UserProvider.(*authentication.LDAPUserProvider)
```
autheliabot, line 154: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>missing whitespace above this line (invalid statement above assign) (wsl_v5)

internal/middlewares/authelia_context.go
autheliabot, line 701: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>unnecessary whitespace (err) (wsl_v5)

internal/configuration/config.template.yml
```diff
@@ -538,6 +538,16 @@ identity_validation:
     ## (&(|({username_attribute}={input})({mail_attribute}={input}))(objectClass=person))
     # users_filter: '(&({username_attribute}={input})(objectClass=person))'
 
+    # additional_principals_dn: 'ou=users'
+
+    ## The principal filter used in search queries to find the user profile based on input filled in login form.
+    ##
+    ## Recommended setting is as follows:
```
autheliabot, line 545: **[yamllint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>[warning] missing starting space in comment (comments)

internal/configuration/config.template.yml
```diff
@@ -538,6 +538,16 @@ identity_validation:
...
+    ## - FreeIPA: (&(krbPrincipalName={input})(objectClass=krbPrincipalAux))
```
autheliabot, line 546: **[yamllint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>[warning] missing starting space in comment (comments)

internal/configuration/config.template.yml
```diff
@@ -580,6 +590,10 @@ identity_validation:
       ## (sAMAccountName and uid) to follow https://datatracker.ietf.org/doc/html/rfc2307.
       # username: 'uid'
 
+      ## The attribute holding the principal name of the user. Only used if principals_filter is set.
+      ## For FreeIPA this is usually 'krbPrincipalName' and for Microsoft Active Directory it's usually 'userPrincipalName'.
```
autheliabot, line 594: **[yamllint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>[warning] missing starting space in comment (comments)

autheliabot, line 594: **[yamllint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>[error] line too long (124 > 120 characters) (line-length)

PR #11160, opened 2026-01-29
ci: remove cve ignore for cve-2025-60876

Crowley723:
https://github.com/wolfi-dev/os/pull/77355


Crowley723:
Just trying to see if grype is still showing the cve.

nightah:
It will still show the CVE because we're patching our own variant and not installing their package in a recognised environment.

PR #11163, opened 2026-01-29
feat(session): use existing sql storage

mairas:
I am creating a Raspberry Pi OS variant that focuses on browser-based management and access, and applications distributed as Debian packages and deployed as Docker containers. One particular feature is that I am implementing SSO for the web apps using Authelia. I want the Authelia sessions to persist over reboots, but for a single-device and single-user use case, having a full-fledged Redis/Valkey container feels like overkill.

Since sessions already require a storage backend for other Authelia data, this PR adds SQL-backed session storage by extending the existing `storage.Provider` interface. When a SQL storage backend (SQLite, MySQL, or PostgreSQL) is configured and Redis is not, sessions are automatically stored in the database — no additional configuration needed beyond setting `session.secret`.

## Changes

- **Migration V0024**: `sessions` table for all three backends (SQLite, MySQL, PostgreSQL)
- **`storage.SessionProvider` interface**: `SaveSession`, `LoadSession`, `DeleteSession`, `DeleteExpiredSessions`, `CountSessions` — embedded in `storage.Provider`
- **SQL implementation**: upsert/select/delete/GC queries with correct dialect handling per backend (SQLite `ON CONFLICT`, MySQL `ON DUPLICATE KEY UPDATE`, PostgreSQL positional params)
- **`SQLSessionProvider` adapter**: implements `fasthttp/session/v2.Provider`, delegates to storage interface
- **Wiring**: `session.NewProvider` accepts `storage.SessionProvider`; `middlewares.NewProviders` passes the storage provider through
- **Validation**: `session.secret` required when SQL storage is active and Redis is not configured
- **Cleanup**: removed file-based session provider and related config/validation/docs
- **Tests**: 24 unit tests covering happy paths and all error paths

## Motivation

For lightweight deployments (IoT, embedded, single-user self-hosted), requiring Redis just for session persistence is unnecessary overhead. The SQL storage backend is already required for other Authelia data, so reusing it for sessions is a natural fit with zero additional infrastructure.

## Design decisions

- **No new table columns for username** — username-based session lookup is future Authelia team work, not in scope here
- **`NeedGC() = true`** — delegates GC scheduling to the fasthttp session framework rather than managing a background goroutine
- **Non-atomic `Regenerate`** — Load → Save → Delete with logged (not propagated) delete errors, consistent with how other providers handle this
- **`context.Background()` for all storage calls** — the `fasthttp/session/v2.Provider` interface doesn't accept contexts; this is an inherent limitation of the upstream interface


internal/session/file_provider_test.go
```diff
@@ -0,0 +1,836 @@
...
+
+	defer provider.Close()
+
+	err = provider.Regenerate([]byte("old"), []byte("new"), 0)
+	assert.Error(t, err)
+
+	err = provider.Regenerate([]byte("old"), []byte("new"), -time.Second)
+	assert.Error(t, err)
+}
+
```
autheliabot: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>File is not properly formatted (gofmt)

james-d-elliott:
I would consider the sessions element of authelia to be frozen at the moment due to a significant rework which will require all session storage to allow lookup of sessions by username.

I would also suggest that the appropriate storage location would be the existing storage interface (SQL) rather than adding additional files which users would not be managing in any way (encryption is a firm requirement of any session storage we implement).

mairas:
> I would consider the sessions element of authelia to be frozen at the moment due to a significant rework which will require all session storage to allow lookup of sessions by username.

OK, got it. Is there a tracking issue I can use to follow this rework? I would be happy to resubmit the PR at a later time.

> I would also suggest that the appropriate storage location would be the existing storage interface (SQL) rather than adding additional files which users would not be managing in any way (encryption is a firm requirement of any session storage we implement).

Now that you said it, this is quite obvious! I just was thinking about a file-based persistence solution and didn't really research the alternatives. For completeness and future reference, I'll change the implementation to use the SQL storage provider instead.

Do you have an estimate how far away the sessions rework is? Weeks? Months? If it's far in the future, it might be best to close this PR and I'll reopen a new one later on.

PR #11173, opened 2026-01-31
fix(configuration): max retries default

james-d-elliott:
This fixes an issue where the regulation max_retries value does not match the documented default. Effectively the documentation is wrong as this has never been the default however it makes more sense to set it to the documented value.


autheliabot (approved):

james-d-elliott:
@all-contributors please add @kevincox for bug

PR #11415, opened 2026-03-10
refactor: set cache duration

james-d-elliott:
This sets the cookie cache duration to 1 year for i18next language detector. In addition this hardens the default Content Security Policy with explicit rules.

### DIFF
--- a/cmd/authelia-gen/const.go
+++ b/cmd/authelia-gen/const.go
@@ -140,8 +140,10 @@ const (
 )
 
 const (
+	codeCSPDirectiveDefaultSrc   = "default-src"
 	codeCSPSelf                  = "'self'"
 	codeCSPDevelopmentDefaultSrc = "'self' 'unsafe-eval'"
+	codeCSPNone                  = "'none'"
 	codeCSPNonce                 = "${NONCE}"
 )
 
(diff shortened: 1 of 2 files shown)


autheliabot (dismissed):

cmd/authelia-gen/const.go
```diff
@@ -155,15 +155,20 @@ const (
 
 var (
 	codeCSPValuesCommon = []CSPValue{
-		{Name: "default-src", Value: ""},
+		{Name: "default-src", Value: codeCSPSelf},
 		{Name: "frame-src", Value: "'none'"},
 		{Name: "object-src", Value: "'none'"},
```
nightah: Should these two also be made a const to be re-used?

cmd/authelia-gen/const.go
```diff
@@ -155,15 +155,20 @@ const (
...
 
-	codeCSPValuesProduction = []CSPValue{}
+	codeCSPValuesProduction = []CSPValue{
+		{Name: "connect-src", Value: codeCSPSelf},
+		{Name: "script-src", Value: codeCSPSelf},
+	}
 
-	codeCSPValuesDevelopment = []CSPValue{}
+	codeCSPValuesDevelopment = []CSPValue{
+		{Name: "default-src", Value: codeCSPDevelopmentDefaultSrc},
```
nightah: This should also likely be considered for a const since it's re-used several times.

PR #11479, opened 2026-03-16
refactor(web): replace mui with tailwindcss + shadcn/base ui

nightah:
Replace Material UI with Tailwind CSS and shadcn/ui components built on Base UI primitives for the entire web frontend. This removes the dependency on @mui/material, @mui/icons-material, @emotion/react, @emotion/styled, @emotion/cache, and react-spinners in favour of a lighter utility-first CSS approach with accessible headless components.

Retire the remaining component libraries along with it. The three icons still using FontAwesome move to lucide-react, dropping four @fortawesome packages, and react18-input-otp is replaced by Base UI's OTPField. Notifications move to Base UI's Toast, which is styled from the theme instead of injecting its own stylesheet, so the Content Security Policy is unchanged.

Add a colour prop to the Button component, mirroring MUI's colour behaviour, allowing variant and colour to be controlled independently across default, ghost, and outline variants with primary, destructive, secondary, and success colour options.

Add an error prop to the Input component, which automatically sets aria-invalid for accessibility, matching the behaviour previously provided by MUI TextField.

Express the theme palette in oklch, the format Tailwind CSS and Base UI use for their own tokens, and correct the values that had drifted from the colours they were ported from.

Improve accessibility beyond the MUI baseline: dialogs and the settings drawer now have accessible names, icon-only controls carry labels, and the one-time password field is named and marks its invalid state. Translate the sign out tooltips and the password method heading, which rendered in English whatever the selected language.

Fix defects the migration surfaced that predate it. Tooltips sit above the page and took pointer events, so one could cover the control it describes and swallow the click meant for it. The two factor panels reported nothing registered while they were still loading, briefly telling a registered user they had registered nothing. The copy button left its timeouts running after unmount. The WebAuthn and Duo methods guarded the start of their ceremony with a ref that does not survive an effect being run twice, so in development the ceremony was never started and the method waited for an approval it had not asked for.

Fix integration test compatibility, including notification element class names, aria-invalid attributes on form inputs, element ids on the interactive control rather than the hidden input Base UI assigns them to, and a useRef guard in ResetPasswordStep2 to prevent double token consumption during React StrictMode.

Make the integration suites tolerate a page that animates. An element part way through a transition has no visible shape, sits outside the viewport, or is disabled and takes no pointer events, and rod reports each of those as a failure rather than waiting for it to pass, so interactions retry within a deadline of their own and report the state the element was in when they gave up. Notification waits account for several being displayed at once, and the fixed intervals that stood in for waiting on the page are replaced by waits on what the assertion reads.

Update the Vite dev server readiness check in the integration test environment to wait for the server restart after .env.development changes, ensuring tests do not start before the frontend is stable.


web/package.json
```diff
@@ -8,29 +8,29 @@
...
+    "lucide-react": "1.31.0",
     "qrcode.react": "4.2.0",
+    "radix-ui": "1.6.7",
     "react": "19.2.8",
     "react-dom": "19.2.8",
     "react-i18next": "17.0.11",
     "react-router-dom": "7.18.2",
     "react-spinners": "0.17.0",
     "react18-input-otp": "1.1.4",
+    "sonner": "2.0.8",
```
james-d-elliott: You will need to add a CSP hash for the inline style that sonner adds. For 2.0.8 it's `sha256-StEaX+se6YS7pqjzrzMIA0KaX9zF/8zAhvQXZAe5epY=`.

web/src/views/Settings/TwoFactorAuthentication/WebAuthnCredentialsPanel.tsx
```diff
@@ -263,36 +265,41 @@ const WebAuthnCredentialsPanel = function (props: Props) {
...
-                            </Button>
-                        </Tooltip>
-                    </Grid>
-                    <Grid size={{ xs: 12 }}>
+            <Card>
+                <CardContent className="grid grid-cols-12 gap-4 p-4">
+                    <div className="col-span-12">
+                        <h5 className="text-xl font-semibold">{translate("WebAuthn Credentials")}</h5>
+                    </div>
+                    <div className="col-span-12 md:col-span-2 xs:col-span-4">
```
james-d-elliott, line 273: The xs breakpoint doesn't actually exist, we probably need to add it and mirror MUI?

web/src/views/LoginPortal/FirstFactor/FirstFactorForm.tsx
```diff
@@ -175,7 +166,7 @@ const FirstFactorForm = function (props: Props) {
     };
 
     const handleUsernameKeyDown = useCallback(
-        (event: KeyboardEvent<HTMLDivElement>) => {
+        (event: KeyboardEvent<HTMLInputElement>) => {
```
james-d-elliott, line 174: We will need to add event.preventDefault() to this handler in the Enter path since there is an actual form now which has caused the event to change types, I'm fairly sure this will cause issues when you press Enter otherwise.

web/src/views/LoginPortal/FirstFactor/FirstFactorForm.tsx
```diff
@@ -245,128 +236,105 @@ const FirstFactorForm = function (props: Props) {
...
                     ) : null}
-                    <Grid size={{ xs: 12 }}>
+                    <div className="w-full">
                         <Button
                             id="sign-in-button"
-                            variant="contained"
-                            color="primary"
-                            fullWidth={true}
-                            endIcon={loading ? <CircularProgress size={20} /> : null}
+                            variant="default"
```
james-d-elliott, line 336:
```suggestion
                            type="submit"
                            variant="default"
```

web/src/layouts/SettingsLayout.tsx
```diff
@@ -55,74 +45,45 @@ const SettingsLayout = function (props: Props) {
...
-        </Box>
+                        <Menu className="size-6" />
+                    </Button>
+                    <h6 className={cn("grow text-lg font-medium", drawerOpen && "hidden")}>{translate("Settings")}</h6>
+                </div>
+            </nav>
+            <Sheet open={drawerOpen} onOpenChange={setDrawerOpen}>
+                <SheetContent side="left" className="p-0" style={{ width: drawerWidth }}>
+                    <div className="text-center">
+                        <h6 className="my-4 text-lg font-medium">{translate("Settings")}</h6>
```
james-d-elliott: Wondering if this should be this instead?


```suggestion
                        <SheetTitle className="my-4 text-lg font-medium">{translate("Settings")}</SheetTitle>
```

james-d-elliott (dismissed):

web/package.json
james-d-elliott: Ahh.. never mind it entirely does dynamic inline styling. That hash only works for a fully default notification.

james-d-elliott: This one might just need some extra checking, looks like vite may bundle it correctly as well. I suspect it'll still error on the console but may be worth just ignoring.

web/src/views/LoginPortal/SecondFactor/OTPDial.tsx
```diff
@@ -27,46 +25,35 @@ export enum State {
...
+                    disabled={disabled}
+                    validationType="numeric"
+                    inputMode="numeric"
                     autoComplete="one-time-code"
-                />
-            </StyledOtpInputContainer>
+                >
+                    {Array.from({ length: props.digits }, (_item, index) => (
+                        <OTPField.Input key={index} autoFocus={index === 0} />
+                    ))}
```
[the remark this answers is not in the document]
nightah: Current code already carries aria-invalid per input plus a field-level label.

james-d-elliott (dismissed):

james-d-elliott (dismissed):

autheliabot (dismissed):

autheliabot (dismissed):

james-d-elliott (dismissed):

james-d-elliott (approved):

autheliabot (approved):

PR #11525, opened 2026-03-20
refactor(web): move password visibility to reusable hook

Crowley723:
Moves password visibility state to a common hook, also changes the behavior to be click-to-toggle rather than hold-to-reveal which should be more accessible and easy to use.


james-d-elliott (dismissed):

james-d-elliott (approved):

nightah (approved):
LGTM.

PR #11641, opened 2026-04-03
feat(storage): improved encryption kdf

james-d-elliott:
This improves the Key Derivation Function used to build the AES-256-GCM key for storage encryption. Previously it was just a SHA256 sum of the users chosen input. This is perfectly fine if the value is truly random, however switching the HKDF ensures even really low entropy keys are strong against brute-force.


Crowley723 (approved):
Looks good.

PR #11659, opened 2026-04-05
ci: add required minumum release age for js dependencies

Crowley723:
Adds a 2 day lag time for javascript dependency updates. This aims to prevent merging malicious packages soon after initial release.

Since we don't generally cut releases exclusively for dependency updates (except for critical security updates), this shouldn't effect things too much.

### DIFF
--- a/.renovaterc
+++ b/.renovaterc
@@ -81,7 +81,9 @@
       ],
       "addLabels": [
         "javascript"
-      ]
+      ],
+      "minimumReleaseAge": "2 days",
+      "minimumReleaseAgeBehaviour": "timestamp-required"
     }
   ],
   "postUpdateOptions": [
--- a/.renovaterc
+++ b/.renovaterc
@@ -83,7 +83,7 @@
         "javascript"
       ],
       "minimumReleaseAge": "2 days",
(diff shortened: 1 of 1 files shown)


james-d-elliott (approved):

james-d-elliott:
I don't mind doing this, I suspect it's in relation to axios, my comments are in line with that assumption. Some things we may want to consider:

1. Using pnpm protects us from 99% of all supply chain attacks
2. The DNS black holing on the CI nodes is fairly robust, in this instance if we were stupid and using npm the hosts for the stage 2 and stage 3 payloads would fail to resolve
3. This may affect our ability to merge any dependency due to missing vulnerable dependency bumps and npm dependencies represent over 75% of our direct and transient vulnerable dependencies I think
4. We can make improvements in other ways that would highlight a supply chain attack like this, for example:
    1.  We could highlight dependencies that do not have release notes
    2. We could implement a whitelist only firewall or DNS blackhole on the CI nodes
    3. We could implement the delay ourselves and let anything non-vulnerable breathe for longer, it's not like we auto-merge these bumps like many do which is what the proposed policy is intended to prevent if I recall correctly

PR #11700, opened 2026-04-08
docs: fix incorrect command name and link reference

CarrotManMatt:
Uses `crypto hash generate` for command name link rather than incorrect `crypt hash generate`


nightah (approved):
LGTM.

james-d-elliott (approved):

PR #11748, opened 2026-04-11
test(suites): add external browser suites for docs and templates

nightah:
Replaces the `pkgvalidate.sh` build-only check with a new `suites external` namespace that spins up a project-local dev server (Hugo for docs, react-email for templates) and drives a headless rod browser against it to assert the rendered output is correct.

Before, CI only confirmed that `pnpm -C docs build` and `pnpm -C internal/templates/src render` exited cleanly — a regression in a shortcode, a broken image path, a missing search initialiser, a Tailwind class, or a template PreviewProps drift would pass silently.

Suites are gated behind a `//go:build externalsuites` tag and registered in their own `ExternalGlobalRegistry`, so `go test ./internal/suites` and `authelia-scripts suites list` remain unaffected. A new `authelia-scripts suites external [list|test]` sub-namespace runs them via a shared `DevServer` lifecycle helper that spawns `pnpm dev` in its own process group, polls for HTTP readiness, and tears the whole tree down on exit.

In Buildkite, `.buildkite/steps/pkgvalidate.sh` is deleted, and the two old package-validation steps are replaced with a single `External Tests` step that runs `e2epackages.sh | buildkite-agent pipeline upload`, mirroring the `e2etests.sh` dynamic pattern so new external suites register without pipeline edits.


james-d-elliott (approved):
LGTM

autheliabot (approved):

PR #11803, opened 2026-04-16
fix(configuration): preserve dots in map key names during koanf remap

nightah:
Koanf uses '.' as its path delimiter. When `koanfRemapKeys` calls `ko.All()` it produces a flat map where dots from the path hierarchy and dots from user-defined map key names (e.g. OIDC scope names, claim names, policy names) are indistinguishable. The subsequent confmap.`Provider(keys, ".")` calls maps.Unflatten which naively splits on '.', mangling any map key containing dots into incorrect nested levels.

This introduces `koanfUnflattenWithKeyMap` which uses Koanf's `KeyMap` to determine the correct key parts for each flattened key, preserving dots within map key names. For keys not found in the KeyMap (e.g. keys introduced by deprecation remapping), it falls back to `strings.Split` which is safe because all deprecation target keys are well-known paths without dots in segments.

Additionally the key validator regex in `NewKeyMapPattern` is updated to allow dots, colons, and tildes in map key segments so that dotted scope names, URI-style claim names, and
similar identifiers are not rejected during key validation.


james-d-elliott (approved):
LGTM

autheliabot (approved):

PR #11811, opened 2026-04-17
refactor(web): adopt abort controller and drop mounted-ref pattern

nightah:
eslint-plugin-react-hooks v7.1.0 bump introduces preserve-manual-memoization, set-state-in-effect, immutability, refs, and exhaustive-deps rules that surface a broader set of stale patterns across the login and settings views. Rather than suppress the new rules, the affected components move to modern React idioms.


james-d-elliott (approved):
LGTM

autheliabot (approved):

PR #11859, opened 2026-04-22
feat(web): add custom_css configuration option

flexponsive:
This PR introduces a new root-level configuration option, `custom_css`, allowing administrators to customize the Authelia portal's appearance without requiring a custom build. This supports both local file paths and external URLs.

This feature enables common branding requests, such as adding company names to the login page or setting custom background images, directly addressing two of the project's top 5 most upvoted discussions: [#7469](https://github.com/authelia/authelia/discussions/7469) and [#5169](https://github.com/authelia/authelia/discussions/5169)

## Key Changes

  * **Configuration:** Added `custom_css` to the root configuration schema.
  * **Security & CSP:** If an external URL is provided, the domain is appended to the `style-src` Content Security Policy directive. This ensures the custom stylesheet is loaded without requiring manual CSP overrides.
    * *Other resources*,  like images referenced in the CSS must be hosted on the same origin as Authelia (e.g., by mounting an assets folder into the Docker container and using a relative path), due to the existing CSP policy.
  * **Frontend:** Renders the `<link>` tag server-side via Go templates at the end of the `<head>` section.

## Example Customization

authelia customized login page with custom_css

PR #11862, opened 2026-04-22
feat(web) adding custom css asset override

flexponsive:
This PR introduces a new asset override capability for `custom.css`, allowing administrators to customize the Authelia portal's appearance without requiring a custom build.

By leveraging the existing `server.asset_path` configuration, administrators can now simply place a `custom.css` file in their assets directory to have it automatically injected into the portal. This addresses highly requested branding and customization needs, such as setting custom backgrounds or adjusting specific UI elements, as seen in popular community discussions like [#7469](https://github.com/authelia/authelia/discussions/7469) and [#5169](https://github.com/authelia/authelia/discussions/5169).

## Key Changes

* **Asset Override Support:** Integrated `custom.css` into the existing asset override system.
* **Automatic Injection:** Authelia now detects the presence of `custom.css` in the configured `asset_path` and automatically injects it into the portal's HTML.
* **Frontend Integration:** Updated the server-side Go templates for both the React index and the JSON index to handle the injection. The `<link>` tag is inserted at the end of the `<head>` section to ensure custom styles take precedence.
* **CSP Compatibility:** The custom stylesheet is served via the internal asset middleware from the Authelia origin (`/static/media/custom.css`), ensuring it works seamlessly with the default Content Security Policy.
* **Documentation:** Updated the Server Asset Overrides reference guide to include the new `custom.css` asset.

## Example Customization

To apply custom styles, set the `asset_path` in your configuration and place a `custom.css` file in that directory:

```yaml
server:
  asset_path: '/config/assets/'
```

Example of what customization is possible:

authelia customized login page with custom_css

Note - I earlier tried an alternative implementation of the same feature in #11859, but after addressing the AI feedback It became a pretty PR. Looking forward to seeing how this one will go!


Brezinben:
Hello any news about this pr ?

james-d-elliott:
It has not passed any design stage and we're still thinking on the final design. 

I believe the shape will be materially different allowing the end-user to control the light/dark mode and allowing the admin to configure which theme/css is available in light/dark mode. Then realistically the admin should be allowed to permit the users to select their own theme pairs from the frontend.

PR #11899, opened 2026-04-29
docs: update authelia nginx site config sample

arylatt:
The Authelia site config sample suggested overriding the `/api/verify` and `/api/authz/` paths to not include the `proxy.conf` snippet.

This is not really correct as:

* The `/api/verify` endpoint is deprecated in favour of the `/api/authz/...` endpoints.
* The `/api/authz/` endpoints should have the headers and configuration applied from the `proxy.conf` snippet.

    The NGINX guide calls out that it is covering the scenario that Authelia and the Protected Applications are running on the same host, and therefore NGINX can contact Authelia directly (e.g. `http://authelia:9091/api/authz/...`).

    Whilst not applying the `proxy.conf` snippet to `/api/authz/...` can be useful in cases where the Protected Applications are not running on the same host, that configuration is currently outside the scope of the guide.


james-d-elliott:
Thanks for the PR. I'm in two minds about this.

While the intent at the time of adding it to nginx was to help users, I believe this may also not be understood correctly by users.

I think in general what we should do is:

1. Remove `/api/verify` entirely as it's deprecated.
2. Remove the paths from the standard example.
3. Add a section in the nginx guide detailing what the section is meant to achieve.
4. Detail associated risks and mitigations.
5. Link to it int he NPM guide.

arylatt:
Thanks for the feedback @james-d-elliott.

Good point, in my case the reason that:

```
location /api/authz/ {
  proxy_pass $forward_scheme://$server:$port;
}
```

works better without the `include /snippets/proxy.conf;`, is because in my setup the Authelia server and the protected applications are hosted on separate machines, and therefore the `$upstream_authelia` value is another NPM instance which is what causes the headers to be overridden ultimately causing the issue I was encountering.

In the NGINX integration guide, it does call out that the expectation is that all the protected applications exist on the same host as Authelia and as such NGINX can reach Authelia directly (therefore avoiding that issue).

I'll look at updating the `site-confs/auth.conf` snippet to only include `location /` which is what the guide should say based on the assumptions stated, etc... And I'll align the NPM and NGINX guides.

arylatt:
@james-d-elliott I've updated the PR to reflect a change to the NGINX documentation instead, removing the references to `/api/verify` and `/api/authz/` overrides.

As mentioned, my issue was related to wiring up `$upstream_authelia` on a separate machine to use the public-facing Authelia endpoint and `proxy.conf` overriding things. Updating my configuration to access Authelia directly over a VPN tunnel fixes that issue more securely.

Since the NGINX and NPM guides mention that they are focused on single-host deployments, I'm not sure where that kind of documentation would be best to reside - it might be too specific for these guides.

Let me know your thoughts :)

james-d-elliott (approved):

autheliabot (approved):

PR #11903, opened 2026-04-30
feat(notification): support fifo destinations for filesystem notifier

### LINKED ISSUE #11386: Allow notifier filesystem to be a FIFO pipe
### Description

I'd like to use a FIFO pipe for the filesystem notifier so I can pipe the notification flow using an external notifier. This is currently impossible as Authelia panics and aborts when it can't sync the file.

### Use Case

Custom filesystem backed notifiers, such as fifo-piping to ntfy or alike

### Details

To reproduce:
`mkfifo /notifier.fifo`
_point authelia notifier.filesystem.filename to /notifier.fifo_
Authelia will write to the pipe, then abort what its doing because the `sync` syscall is not available.

### Documentation

_No response_

### Generative AI

No

### Pre-Submission Checklist

nightah:
Pointing the filesystem notifier at a fifo previously panicked: `StartupCheck` truncated the path with `os.WriteFile`, which on a fifo opens `O_WRONLY|O_CREATE|O_TRUNC` and blocks until a reader connects, and `Send` always called `f.Sync()`, which returns `EINVAL` on a pipe.

`FileNotifier` now stat-checks the destination via `os.ModeNamedPipe`. When the path is a fifo, `StartupCheck` skips the empty-write step and `Send` opens with `O_WRONLY` only and skips the trailing `f.Sync()`. Detection is per-call so it works whether or not `disable_startup_check` is set, and it survives the operator replacing the file between startup and runtime. Behaviour for regular files is byte-identical to before, including the `O_TRUNC` / `O_APPEND` open flags and the `Sync` call. The bytes written to the destination (`Date:` / `Recipient:` / `Subject:` header, rendered template body, double newline) are identical regardless of whether the destination is a regular file or a fifo, so existing parsers continue to work. The fifo must be pre-created by the operator with `mkfifo` and a reader must be attached on the other end; if no reader is connected when a notification is generated, `Send` blocks until one is, matching POSIX fifo semantics.

A new `authelia debug notification` subcommand is added alongside `authelia debug tls` / `expression` / `oidc`. It loads the configuration, runs the notifier startup check, and dispatches a single test notification using the configured notifier (SMTP or filesystem). It is useful as an ops tool for verifying that the SMTP server, filesystem path, or fifo consumer is reachable, and is the deterministic, browser-free trigger used by the new CLI suite tests.

Coverage for the fifo branches is added in `internal/notification/file_notifier_test.go` (`TestIsFIFO`, fifo `StartupCheck`, fifo `Send` end-to-end via `syscall.Mkfifo` plus a reader goroutine), and the CLI integration suite gains two end-to-end tests: `TestShouldDeliverSequentialNotificationsToFIFOViaDebugCommand` runs three notifications back-to-back through fresh `O_RDONLY` readers to validate pipe re-use, and `TestShouldDeliverConcurrentNotificationsToFIFOViaDebugCommand` holds the fifo open `O_RDWR` while two `authelia debug notification` invocations write concurrently and asserts both subjects land intact.

Closes #11386.

### DIFF
--- a/docs/content/configuration/notifications/file.md
+++ b/docs/content/configuration/notifications/file.md
@@ -41,3 +41,15 @@ This section describes the individual configuration options.
 {{< confkey type="string" required="yes" >}}
 
 The file to add email text to. If it doesn't exist it will be created.
+
+If `filename` points to an existing FIFO (named pipe), Authelia detects this from the file mode and writes notifications
+to the pipe instead of a regular file. The startup check leaves the FIFO untouched and the per-notification write skips
+the disk sync that is unsupported on pipes. The byte stream written to the pipe is identical to the regular-file format,
+so any tooling that already parses notifications continues to work.
+
+The FIFO must be created by the operator with `mkfifo` before Authelia starts; Authelia will not create one. A reader
+must be attached on the other end of the pipe. If no reader is connected when a notification is generated, the write
+blocks until one is.
+
+Use the `authelia debug notification` subcommand to verify the configured notifier (file, FIFO, or SMTP) without
+exercising the full identity verification flow.
--- a/docs/content/reference/cli/authelia/authelia_debug.md
+++ b/docs/content/reference/cli/authelia/authelia_debug.md
@@ -2,7 +2,7 @@
 title: "authelia debug"
 description: "Reference for the authelia debug command."
 lead: ""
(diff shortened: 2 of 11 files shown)


james-d-elliott (approved):
LGTM

autheliabot (approved):

PR #11975, opened 2026-05-08
Custom Redis key prefix

brwyatt:
Simple change to add a `key_prefix` key to the Redis config to allow changing the key prefix for deployments that might need it (defaults to the current value `authelia-session`)

Reasons this change is desired:
- Multiple installs that may use the same backend Redis server(s) (Gamma/staging environments/deployments, multi-tenant environments, etc)
- Enabling using a Redis/Valkey cluster behind a cluster proxy (for example, Shotover) to force consistent hashing to support some key operations used by Authelia.

This change also does not change the default settings, so those who don't need it, are not impacted.

I personally needed (and use) this in my own environment in order to make use of my Valkey cluster (with Shotover transparent proxy), as some of the key operations Authelia performs that do not work across partitions.

### DIFF
--- a/internal/configuration/schema/keys.go
+++ b/internal/configuration/schema/keys.go
@@ -405,6 +405,7 @@ var Keys = []string{
 	"session.redis.high_availability.sentinel_password",
 	"session.redis.high_availability.sentinel_username",
 	"session.redis.host",
+	"session.redis.key_prefix",
 	"session.redis.max_retries",
 	"session.redis.maximum_active_connections",
 	"session.redis.minimum_idle_connections",
--- a/internal/configuration/schema/session.go
+++ b/internal/configuration/schema/session.go
@@ -53,6 +53,7 @@ type SessionRedis struct {
 	MaximumActiveConnections int           `koanf:"maximum_active_connections" yaml:"maximum_active_connections" toml:"maximum_active_connections" json:"maximum_active_connections" jsonschema:"default=8,title=Maximum Active Connections" jsonschema_description:"The maximum connections that can be made to redis at one time."`
 	MinimumIdleConnections   int           `koanf:"minimum_idle_connections" yaml:"minimum_idle_connections" toml:"minimum_idle_connections" json:"minimum_idle_connections" jsonschema:"title=Minimum Idle Connections" jsonschema_description:"The minimum idle connections that should be open to redis."`
 	TLS                      *TLS          `koanf:"tls" yaml:"tls,omitempty" toml:"tls,omitempty" json:"tls,omitempty" jsonschema:"title=TLS" jsonschema_description:"The TLS configuration for the redis server."`
+	KeyPrefix                string        `koanf:"key_prefix" yaml:"key_prefix,omitempty" toml:"key_prefix,omitempty" json:"key_prefix,omitempty" jsonschema:"default=authelia-session,title=Key Prefix" jsonschema_description:"The prefix for the keys stored in redis."`
 
 	HighAvailability *SessionRedisHighAvailability `koanf:"high_availability" yaml:"high_availability,omitempty" toml:"high_availability,omitempty" json:"high_availability,omitempty" jsonschema:"title=High Availability" jsonschema_description:"The redis high availability configuration."`
 }
@@ -91,6 +92,7 @@ var DefaultRedisConfiguration = SessionRedis{
 	Timeout:                  time.Second * 5,
 	MaxRetries:               0,
 	MaximumActiveConnections: 8,
+	KeyPrefix:                "authelia-session",
 	TLS: &TLS{
 		MinimumVersion: TLSVersion{Value: tls.VersionTLS12},
 	},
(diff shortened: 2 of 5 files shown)


james-d-elliott:
Is there a reason you'd not use a db index? In basically all iterations you can increase this to 256 with a theoretical maximum far exceeding this.

It should be noted that sessions are in a change freeze at the moment, and the chances are we're not going to support this going forward as we'll be using it for pubsub etc.

I think with absence of a good reason to not use DB indexes we would probably not accept this.

brwyatt:
Unfortunately, in my case, DB indexes are not a valid solution, as DB Indexes are unavailable and do not work when Redis/Valkey are operating in clustered mode. Likewise, some of the operations Authelia needs doesn't natively work across cluster partitions, thus the need to be able to change the prefix to (in my case) leverage hash keys to enforce stable hashing of  Authelia keys on the cluster nodes.

Without this change, it would require spinning up separate, dedicated Sentinel-based Redis/Valkey, rather than leveraging the existing shared cluster and Shotover proxy infrastructure. This change effectively makes no logical changes to the Redis code, and simply turns a static string into a configurable value, so should (though reality is rarely so convenient) be low impact to planned rework/changes and low risk. The only logic added (excluding tests) is a single branch in the config validator (outside the Redis code) to ensure the default value is set if it isn't provided.

THAT SAID... I do recognize the need for change freezes, and I do understand there may be other priorities or conflicting plans. I also recognize that, if/once added, it does mean it becomes a "feature" that needs to be maintained that would have to be included in any future refactors, and (even if minimal), that is still a non-zero amount of increased effort.

I do hope, though, that given the simple nature of this change (and that the majority is in the config validator) and preservation of default behavior, that accepting/including this change is able to be merged as a "quick win" that has benefit to those who can use it, and neutral impact to anyone else. (at least that's my hope, though I understand/respect if not)

Crowley723:
I want to mention that we don't support Redis running in clustered mode. Just sentinel and standalone. See https://github.com/authelia/authelia/discussions/5474

brwyatt:
@Crowley723 

Correct. In my case, this is bypassed with Shotover acting as a transparent protocol proxy for cluster-unaware clients (like Authelia), which handles the connection part, but it doesn't magically make commands work cross-partitions that can't. This would allow users to leverage hashkeys to "fix" all the keys to a cluster partition, without having to officially support clustering, but still benefiting from the transparent handling of partition primary failovers, so while it doesn't provide all the benefits of clustering (performance-wise), it does still allow for some of the operational benefits on the admin side (or, at least I feel more comfortable with the failover patterns and semantics of clustering behind a proxy vs the Sentinel pattern -- but I'll admit that it may be a failing/flaw on my part... but it is also similarly why I run MariaDB/Galera rather than tracking primaries/replicas).

But, that said... removing a static, hard-coded string for a prefix to be a configurable value... seems like the kind of thing that opens up a lot of end-user options beyond that could adapt to any number of "weird" environments.

But I can also recognize that there may be no reasonable/practical use beyond pseudo-support (or, I guess "pseudo-compatibility") for clusters. So I'll "push" for it, within reason, but also happy to step down here if it doesn't make sense or fit with what y'all are willing to accept/support. (And I'm working on a backup to use transform hooks in Shotover, though that just _feels_ wrong (and more computationally complex), especially compared to making a static string configurable...)

james-d-elliott:
The change freeze exists due to a ground up rewrite of the session lib, we're dumping it (we will be adding cluster support). We can look at adding support for this specifically with cluster.

brwyatt:
Sounds good to me! And it will probably be more useful to (officially) have a configurable prefix there due to clusters not supporting DB indexes.

I'll switch tracks to leverage the Shotover proxy transform in the meantime, then, and look forward to official cluster support in the future! (I mean, realistically, I'll probably run into something else with similar limitations that'll need, too)

james-d-elliott:
Looking at how I've done the session rework the keys are bound to the domain of the session cookie. i.e. `authelia:<domain>:<purpose>:<lookup>`. I believe this solves your use case entirely?

brwyatt:
I don't think so, at least not without knowing more about the commands being used with the keys. The issue (with clustering in this context) is any action that would cause a `CROSSSLOT` error. So it kinda depends more on what operations are being called on things. I haven't really dug into the cause of the error currently, but my assumption would be renames/moves. So in this case (assuming similar operations are in use here that would cause `CROSSSLOT` errors due to to keys being hashed to different partitions/slots), it would need to be `{authelia:<domain>:<purpose>}:<lookup>` to force keys into the same slot to avoid the risk of those operations having errors. Unless the "lookup" part is more stable in this implementation than it currently is.

But I'll stress that this is based on a lot of assumptions and not having dug too deeply into implementation details. But if native cluster support is being baked-in on this newer iteration... I'd kinda assume that there's more work under the hood that makes the key (and `CROSSSLOT` errors) less of an issue, so it probably doesn't matter too much in that context.

jack9603301:
I'm in favor of this feature, which means that key conflicts can be 
avoided with reasonable settings.Message ID: 
***@***.***>

james-d-elliott:
Will be taken over locally due to the changes, proper research has been conducted into how clustering works and it will be implemented per https://redis.io/docs/latest/operate/oss_and_stack/reference/cluster-spec//

brwyatt:
Thanks! Looking forward to official (and direct) cluster support!

james-d-elliott:
You can play around with it in the feat-session-rewrite docker tag. sessions between the two are not compatible.

PR #12094, opened 2026-05-18
fix(authentication): incorrect bind mode

james-d-elliott:
This fixes an issue where theoretically anonymous bind could be used in situations where it would not work causing an error or could inadvertently introduce future regressions.


nightah (approved):
LGTM.

autheliabot (approved):

PR #12095, opened 2026-05-18
fix(storage): harden one-time code consumption

james-d-elliott:
This hardens the consumption of one-time codes drastically increasing defense in depth by both including the client IP in the signature bounding the signature to the client IP, and ensuring the update actually changes a row. This would prevent the theoretical scenario of the same user using the same code twice on two different sessions with the same username at the exact same nanosecond.


nightah (approved):
LGTM.

autheliabot (approved):

PR #12120, opened 2026-05-21
Update Roundcube redirect URI for OAuth callback

jmp1985:
Based on the documentation here: https://github.com/roundcube/roundcubemail/wiki/Configuration:-OAuth2


james-d-elliott:
Please include an additional update to the version you have tested this with.

james-d-elliott (changes requested):
Please also update the version you've tested this with. The version is at the top of the file.

jmp1985:
The pull request is a draft at the moment. I will update it with additional information when I have some spare time and will change it from draft to open when I am ready.

PR #12146, opened 2026-05-23
fix(utils): domain canonical case handling

james-d-elliott:
This fixes an issue where domains requested with uppercase letters do not match a cookie domain leading to a failure to find a session config. The result of this is incorrect 40x or 50x status codes.


internal/utils/url_test.go
```diff
@@ -67,8 +67,34 @@ func TestHasURIDomainSuffix(t *testing.T) {
...
-	assert.False(t, HasDomainSuffix("abc", ""))
-	assert.False(t, HasDomainSuffix("", ""))
+	testCases := []struct {
+		Name         string
+		Domain       string
+		DomainSuffix string
+		Expected     bool
+	}{
+		{"ShouldNotMatchEmptySuffix", "abc", "", false},
+		{"ShouldNotMatchEmptyDomainAndEmptySuffix", "", "", false},
```
Crowley723, line 77: Nit: These technically test the same thing.
"" + "" tests the same as "abc" + "" (since empty strings are technically still strings)

Crowley723 (dismissed):
One nit, take it or leave it.
Looks good otherwise

internal/utils/url_test.go
james-d-elliott: Yeah makes sense. Just added to ensure regressions don't occur if we change it in any meaningful way.

Thanks, will undraft and merge once the respective dependent commit is merged.

Crowley723 (dismissed):

nightah (dismissed):
LGTM.

autheliabot (approved):

Crowley723 (approved):

PR #12154, opened 2026-05-24
test: avoid timezone and fixed-port assumptions

immanuwell:
## Summary
This cleans up 2 env-bound tests. One hardcodes a local time, the other grabs port 3389 and hopes for the best.

## Repro
On `master`:
- `TZ=Asia/Yerevan go test ./internal/model -run TestNewOAuth2DeviceCodeSessionFromRequest -count=1`
- `go test ./internal/authentication -run TestNewLDAPClientDialerStandard -count=1`

Second one is very real btw, if `127.0.0.1:3389` is already taken it fails right away. On this box `ss -ltn "( sport = :3389 )"` shows that port is in use.

## Fix
- derive `CheckedAt` from `time.Unix(100000000, 0)`
- use an ephemeral listener port in the LDAP dialer test


nightah:
The port change is a welcome addition, but I don't understand the rationale for the other OIDC test changes.

Did you experience an issue with these?

immanuwell:
@nightah I reproduced it on master with:

```
env TZ=Asia/Yerevan go test ./internal/model -run TestNewOAuth2DeviceCodeSessionFromRequest -count=1
```

the test hardcodes `CheckedAt` as `time.Date(..., time.Local)`, but the actual value comes from `LastChecked: time.Unix(100000000, 0)`
under `TZ=Asia/Yerevan` those are different local times for different instants, so the assertion fails
this change makes the expectation use the same instant as the input instead of a timezone-dependent local-time literal

james-d-elliott:
We don't do any of our checks on systems which use UTC time, as long as your timeszone matches the actual time it will work. The test is failing due to a misconfiguration of the timezone, which is the same as it would in production in a bad config.

PR #12160, opened 2026-05-25
fix(handlers): hoist issuer checks

james-d-elliott:
This moves the issuer check up higher at every endpoint where applicable. While this probably doesn't make for a meaningful fix it will ensure the most relevant and meaningful error is shown first. In addition it changes the way these messages are communicated. It also add some additional simple hardening.


autheliabot (dismissed):

autheliabot (approved):

nightah (approved):
LGTM.

PR #12169, opened 2026-05-26
fix(authorization): case-insensitive domain matching [security]

james-d-elliott:
This fixes edge cases where domain matching was case-sensitive. This specifically applies when using wildcard records or user/group segment domains.

Fixes: GHSA-j748-h363-wqj8


Crowley723 (approved):

nightah (approved):
LGTM.

PR #12254, opened 2026-06-07
Polish translation fixes

fortellerq:
Hey!
Polish is my native language and I noticed an error in the translation. Originally "Sign in with a passkey" in polish locale would translate back to "Sign in with a password". "Hasło" means "password" and "passkey" is usually translated into "klucz uwierzytelniania", but can be shortened to just "klucz" (which, in turn, means literally a key). In this context it can be used without losing the meaning (using full "klucz uwierzytelniania" would make the label rather long).

Then I figured to check other labels as the list isn't very long and found few that were either wrong, or gramatically incorrect. I feel pretty confident about these changes, but can invite someone from Poland to perform the review.

Best regards!
Michał


fortellerq:
closing as this clearly doesn't follow the PR rules.

PR #12255, opened 2026-06-07
fix(web): fix grammar in polish locale

fortellerq:
Hey!
Polish is my native language and I noticed an error in the translation. Originally "Sign in with a passkey" in polish locale would translate back to "Sign in with a password". "Hasło" means "password" and "passkey" is usually translated into "klucz uwierzytelniania", but can be shortened to just "klucz" (which, in turn, means literally a key). In this context it can be used without losing the meaning (using full "klucz uwierzytelniania" would make the label rather long).

Then I figured to check other labels as the list isn't very long and found few that were either wrong, or gramatically incorrect. I feel pretty confident about these changes, but can invite someone from Poland to perform the review.

Best regards!
Michał


james-d-elliott (changes requested):
Adjustments to translations are to occur [here](https://translate.authelia.com/) as per our contribution guideliines.

Crowley723:
Hey @fortellerq, thanks for reviewing the translations. We make use of crowdin to manage translations. I would ask that you make an account over at https://authelia.crowdin.com/ and submit your changes there.

fortellerq:
hello @james-d-elliott, @Crowley723 
Thank you for directing me to the right place and sorry for not following the rules.

I just made the changes to labels on Crowdin. I am not familiar with this platform and making changes was almost too easy. How could I make sure that changes there will be picked up? Is just making changes enough, or some kind of review is required?

PR #12258, opened 2026-06-11
ci: broaden lefthook grype ignore to all embedded packages

nightah:
The grype ignore rule for the lefthook prebuilt binaries was scoped to `stdlib` go-module findings only, so each new transitive dependency statically linked into the binary (e.g. `golang.org/x/sys` flagged by `GO-2026-5024`) surfaced as a fresh failure and had to be added by hand.

Drop the `name` and `type` constraints and match purely on the `**/lefthook*/**` source path. Any package embedded in the lefthook binary (stdlib, `golang.org/x/*` modules, and future transitive deps) is now suppressed, while the path scoping still ensures the same packages remain visible when they appear in Authelia's own runtime binaries. Lefthook is a developer/CI git hook runner only and is never shipped in a runtime artifact.

### DIFF
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -25,11 +25,10 @@ ignore:
       Explicitly patched in authelia/base. The reported version string is unchanged from upstream busybox 1.37.0
       so grype still flags it, but the vulnerable code paths have been backported.
   - package:
-      name: stdlib
-      type: go-module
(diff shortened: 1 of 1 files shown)


autheliabot (approved):

james-d-elliott (approved):
LGTM

PR #12340, opened 2026-06-20
fix(authorization): deprecated wildcard consistency and did

james-d-elliott:
This aligns all of the wildcards to use the modern method and makes the deprecation more explicit for the deprecated ones. This also adds defense-in-depth measures which will be expanded on in a future version.


autheliabot (dismissed):

internal/authorization/access_control_domain.go
```diff
@@ -64,20 +63,25 @@ func (m AccessControlDomainMatcher) IsMatch(domain string, subject Subject) (mat
...
+	case m.UserWildcard, m.GroupWildcard:
+		if !utils.StringHasSuffixFold(domain, m.Name) {
+			return false
 		}
 
-		return domain == fmt.Sprintf("%s%s", subject.Username, m.Name)
-	case m.GroupWildcard:
-		if subject.IsAnonymous() && utils.StringHasSuffixFold(domain, m.Name) {
+		if subject.IsAnonymous() {
 			return len(domain) > len(m.Name)
```
clems4ever: Let's consider the case of user wildcard {user}.example.com with an anonymous user. The following url would match: https://john.evil.example.com/. I think this is not expected.
More generally, do we really want any subdomain to match if the user is anonymous? I do not think so. What do you think?
This pb was already there before your fix as far as I see but this PR might be an opportunity to fix it. Otherwise, tell me and I can open another PR to fix.

internal/authorization/authorizer_test.go
```diff
@@ -175,6 +175,41 @@ func (s *AuthorizerSuite) TestShouldCheckDynamicDomainRules() {
...
+	tester.CheckAuthorizations(s.T(), UserWithGroups, "https://dev.EXAMPLE.com/", fasthttp.MethodGet, OneFactor)
+	tester.CheckAuthorizations(s.T(), UserWithGroups, "https://example.com/", fasthttp.MethodGet, Denied)
+	tester.CheckAuthorizations(s.T(), UserWithGroups, "https://john.example.co/", fasthttp.MethodGet, Denied)
+	tester.CheckAuthorizations(s.T(), UserWithoutGroups, "https://bob.example.com/", fasthttp.MethodGet, OneFactor)
+	tester.CheckAuthorizations(s.T(), UserWithoutGroups, "https://BOB.example.com/", fasthttp.MethodGet, OneFactor)
+	tester.CheckAuthorizations(s.T(), UserWithoutGroups, "https://john.example.com/", fasthttp.MethodGet, Denied)
+	tester.CheckAuthorizations(s.T(), UserWithoutGroups, "https://dev.example.com/", fasthttp.MethodGet, Denied)
+	tester.CheckAuthorizations(s.T(), UserWithoutGroups, "https://admins.example.com/", fasthttp.MethodGet, Denied)
+	tester.CheckAuthorizations(s.T(), AnonymousUser, "https://john.example.com/", fasthttp.MethodGet, OneFactor)
+	tester.CheckAuthorizations(s.T(), AnonymousUser, "https://dev.example.com/", fasthttp.MethodGet, OneFactor)
```
clems4ever, line 198: Can we consider adding a testcase with a username and a group that contain a dot?

internal/authorization/access_control_domain.go
```diff
@@ -64,20 +63,25 @@ func (m AccessControlDomainMatcher) IsMatch(domain string, subject Subject) (mat
...
 		}
 
 		i := strings.Index(domain, ".")
```
clems4ever: Another pb I see. what if the matcher is {user}-abc.example.com and we are in the authenicated case? This code will break because the variable `i` will reference the cutpoint between john-abc and .example.com, however john-abc obviously does not match john.

clems4ever (commented):
few problems. Some existed before your fix but are worth being mentioned in this review I guess. Maybe we can target a fix for all of them at once. Let me know what you think.

internal/authorization/access_control_domain.go
james-d-elliott: Because the user is not known we have to treat an anonymous user as a match here. Otherwise we're unsure if it matches.

This is similar to any rule that has subject matching, they must be 1FA or higher and must authenticate the user if they are soft-hit.

Once the user is authenticated we can then determine if it really matches.

internal/authorization/access_control_domain.go
james-d-elliott: This is intentional, the matcher was never intended to match part subdomains.

We deprecated it for regex group matches.

internal/authorization/authorizer_test.go
james-d-elliott: Added.

autheliabot (dismissed):

internal/authorization/authorizer_test.go
```diff
@@ -175,6 +175,92 @@ func (s *AuthorizerSuite) TestShouldCheckDynamicDomainRules() {
...
+			Domains: []string{"{user}-example.example.com"},
+			Policy:  oneFactor,
+		}).
+		WithRule(schema.AccessControlRule{
+			Domains: []string{"{group}-example.example.com"},
+			Policy:  oneFactor,
+		}).
+		Build()
+
+	tester.CheckAuthorizations(s.T(), UserWithGroups, "https://john-example.example.com/", fasthttp.MethodGet, Denied)
```
clems4ever: Isn't this one supposed to produce one factor instead of denied?

clems4ever: same for the groups.

clems4ever: I mean if it is not, that's really counter intuitive for the users of this feature.

internal/authorization/authorizer_test.go
```diff
@@ -175,6 +175,92 @@ func (s *AuthorizerSuite) TestShouldCheckDynamicDomainRules() {
...
+		}).
+		Build()
+
+	UserHyphenated := Subject{Username: "john-doe", Groups: []string{"dev-team"}, IP: net.ParseIP("10.0.0.8")}
+	UserUnderscored := Subject{Username: "john_doe", Groups: []string{"dev_team"}, IP: net.ParseIP("10.0.0.8")}
+
+	tester.CheckAuthorizations(s.T(), UserHyphenated, "https://john-doe.example.com/", fasthttp.MethodGet, OneFactor)
+	tester.CheckAuthorizations(s.T(), UserHyphenated, "https://dev-team.example.com/", fasthttp.MethodGet, OneFactor)
+	tester.CheckAuthorizations(s.T(), UserHyphenated, "https://JOHN-DOE.example.com/", fasthttp.MethodGet, OneFactor)
+	tester.CheckAuthorizations(s.T(), UserUnderscored, "https://john_doe.example.com/", fasthttp.MethodGet, Denied)
```
clems4ever, line 278: We should probably return a warning in the logs of the server, which would explicitly tell that the user with an _ will likely not behave as expected by the configuration. It cannot be deduced easily by looking simply at the configuration, we can therefore expect some people to open issues about that. Personally I would have dropped those matches (since it's an already deprecated feature) in order to simplify the code and avoid the edge cases.

james-d-elliott: I don't think this is necessary. The request will never make it to Authelia since the domain is entirely invalid.

internal/authorization/authorizer_test.go
james-d-elliott: I don't believe so, it is analogous to how the standard wildcard works. I'll double check the existing behavior to ensure it hasn't changed but I'm relatively certain it had this restriction too.

james-d-elliott: See the adjustments.

autheliabot (dismissed):

docs/content/configuration/security/access-control.md
```diff
@@ -121,15 +121,18 @@ Rules may start with a few different wildcards:
...
-- The user wildcard is `{user}.`, which when in front of a domain dynamically matches the username of the user. For
-  example `{user}.{{< sitevar name="domain" nojs="example.com" >}}` would match `fred.{{< sitevar name="domain" nojs="example.com" >}}` if the user logged in was named
-  `fred`. _**Warning:** this is officially deprecated as the [domain_regex] criteria completely replaces the
-  functionality in a much more useful way. It is strongly recommended you do not use this as it will be removed in a
-  future version, most likely v5.0.0._
-- The group wildcard is `{group}.`, which when in front of a domain dynamically matches if the logged in user has the
-  group in that location. For example `{group}.{{< sitevar name="domain" nojs="example.com" >}}` would match `admins.{{< sitevar name="domain" nojs="example.com" >}}` if the user logged in was
-  in the following groups `admins,users,people` because `admins` is in the list.
+  When using a wildcard like this the string __must__ be quoted like `'*.{{< sitevar name="domain" nojs="example.com" >}}'`.
```
autheliabot: 🚫 **[eslint]** <[prettier/prettier](https://github.com/prettier/eslint-plugin-prettier#options)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Replace `__must__` with `**must**`
```suggestion
  When using a wildcard like this the string **must** be quoted like `'*.{{< sitevar name="domain" nojs="example.com" >}}'`.
```

docs/content/configuration/security/access-control.md
```diff
@@ -577,7 +586,122 @@ identical rules, and one of them has a subject based reliant criteria, and the o
...
+### Case Sensitivity
+
+By default Authelia uses case-insensitive regular expressions in appropriate locations. This is different from the
+default behavior of the Go standard library which is case-sensitive. This is achieved by prepending a `(?i)` to the
+regular expression before parsing it. You can negate this behavior by adding `(?-i)` to the beginning of any regular
+expression which explicitly disables case insensitivity.
+
+The main area affected by this is currently:
+
+* The [domain_regex](#domain_regex) criteria
```
autheliabot: 🚫 **[eslint]** <[prettier/prettier](https://github.com/prettier/eslint-plugin-prettier#options)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Replace `*` with `-`
```suggestion
- The [domain_regex](#domain_regex) criteria
```

docs/content/configuration/security/access-control.md
```diff
@@ -577,7 +586,122 @@ identical rules, and one of them has a subject based reliant criteria, and the o
...
+reflected in the [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.1) which mirrors this sentiment
+more softly but still explicitly states that the path component is case-sensitive.
+
+All regular expressions in Authelia can be modified to be case-insensitive by adding the `(?i)` to the beginning of the
+regular expression and can be modified to be case-sensitive by adding the `(?-i)` to the beginning of the regular
+expression. You can also adjust certain parts of the regular expression to match case-insensitively or case-sensitively
+depending on your needs. Below is a table of examples of how to modify the case-sensitivity of a regular expression.
+
+
+|       Description        |       Case-Sensitive       |     Case-Insensitive      |
```
autheliabot: 🚫 **[eslint]** <[prettier/prettier](https://github.com/prettier/eslint-plugin-prettier#options)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Delete `⏎`
```suggestion
|       Description        |       Case-Sensitive       |     Case-Insensitive      |
```

docs/content/configuration/security/access-control.md
```diff
@@ -577,7 +586,122 @@ identical rules, and one of them has a subject based reliant criteria, and the o
...
+|:------------------------:|:--------------------------:|:-------------------------:|
```
autheliabot: 🚫 **[eslint]** <[prettier/prettier](https://github.com/prettier/eslint-plugin-prettier#options)> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Replace `:------------------------:|:--------------------------:|:-------------------------:` with `·:----------------------:·|·:------------------------:·|·:-----------------------:·`
```suggestion
| :----------------------: | :------------------------: | :-----------------------: |
```

internal/authorization/const.go
```diff
@@ -50,9 +50,31 @@ const (
...
+	// DomainTokenGroup is the token in a domain criteria which is translated into a pattern matching a group.
+	DomainTokenGroup = "{group}"
+)
+
 var (
 	// IdentitySubexpNames is a list of valid regex subexp names.
 	IdentitySubexpNames = []string{subexpNameUser, subexpNameGroup}
+
+	// IdentityDomainTokens is a list of the domain criteria tokens which are translated into patterns matching the
+	// identity of the subject.
```
nightah:
```suggestion
```

internal/authorization/const.go
```diff
@@ -50,9 +50,31 @@ const (
...
+const (
+	// DomainTokenUser is the token in a domain criteria which is translated into a pattern matching the username.
+	DomainTokenUser = "{user}"
+
+	// DomainTokenGroup is the token in a domain criteria which is translated into a pattern matching a group.
+	DomainTokenGroup = "{group}"
+)
+
 var (
 	// IdentitySubexpNames is a list of valid regex subexp names.
```
nightah:
```suggestion
```

nightah (approved):
LGTM.

autheliabot (approved):

PR #12354, opened 2026-06-22
docs(oidc): update headscale

fredrikekre:
The OIDC implementation in Headscale have been fixed since Headscale
version 0.27 (see <https://github.com/juanfont/headscale/pull/2663>) so
the escape hatch configuration is not needed anymore.

### DIFF
--- a/docs/content/integration/openid-connect/clients/headscale/index.md
+++ b/docs/content/integration/openid-connect/clients/headscale/index.md
@@ -25,9 +25,7 @@ seo:
 - [Authelia]
   - [v4.39.20](https://github.com/authelia/authelia/releases/tag/v4.39.20)
 - [Headscale]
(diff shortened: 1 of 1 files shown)


james-d-elliott (approved):

autheliabot (approved):

james-d-elliott:
Thanks for this. How does headscale associate users?

fredrikekre:
In the database you mean? `${iss}/${sub}` by default (https://github.com/juanfont/headscale/blob/33a65052c9aa0bd492b937408b0cf4e618a1bdc8/hscontrol/types/users.go#L306-L345). There doesn't seem to be a way to link an existing user to an OIDC user (except manually updating the DB).

james-d-elliott:
Nice, thanks. That's the correct way to do it (though usually there's a flow to link accounts).

PR #12403, opened 2026-06-28
docs(oidc): add note for wallos v4.9.6

LucasDooms:
A new security setting in wallos was added that may break the OIDC setup process and may also break existing setups.

This adds a note to the docs on how to fix that issue.

Tested on authelia v4.39.20 and wallos v4.9.6.

### DIFF
--- a/docs/content/integration/openid-connect/clients/wallos/index.md
+++ b/docs/content/integration/openid-connect/clients/wallos/index.md
@@ -78,16 +78,15 @@ To configure [Wallos] there is one method, using the [Web GUI](#web-gui).
 
 #### Web GUI
 
-**Important Note:**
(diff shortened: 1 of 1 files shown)


james-d-elliott (changes requested):
Thanks for contributing.

Please remove the note and just update the relevant item in the guide, i.e. in the configure the following options section add the relevant option they must configure.

Crowley723 (commented):
"Enter auth.example.com into the form" isn't super clear to me where in the form it should be put. Is it clear in the ui which field the value should go into? Does the proper field have a label? Maybe we can indicate what field to enter the value into.

LucasDooms:
Here is a screenshot of the settings:
image

There is only one box there, so it should be clear hopefully.

OIDC for reference:
image

james-d-elliott (approved):

autheliabot (approved):

PR #12484, opened 2026-07-11
refactor(configuration): log offending value for totp allowed_digits

TowyTowy:
Disclosure: this change was prepared with the assistance of an AI coding tool (Claude). The tool was used to locate the defect in the TOTP config validator, apply the one-line fix, and write the accompanying unit-test case; I have reviewed and fully understand every line.

**What**

The `allowed_digits` validation loop in `validateTOTPValueSetDigits` builds its error with `config.TOTP.DefaultDigits` instead of the offending value from the `AllowedDigits` slice. A misconfiguration such as `allowed_digits: [20]` therefore reports the value as `'6'` (the default) rather than `'20'`, which is misleading — especially since `6` is itself a valid value.

**Why**

The sibling `allowed_algorithms` and `allowed_periods` loops already report the actual offending value; this aligns `allowed_digits` with that established behavior. The bug has been present since `allowed_digits` was introduced in v4.38.

**Tests**

Corrected `ShouldRaiseErrorWhenInvalidTOTPAllowedValues` (which had asserted the incorrect `'6'`) and added `ShouldReportOffendingAllowedDigitsValueNotDefault` to lock in the offending value independent of the default. `go test ./internal/configuration/validator/` passes; `go vet` and `gofmt` clean.


james-d-elliott (approved):
LGTM

autheliabot (approved):

PR #11432, opened 2026-03-12
feat(webauthn): related origins

james-d-elliott:
This adds support for WebAuthn Related Origins which permits credentials to be used in shared cross-domain scenarios. Origins are grouped into a relying party which declares them in the Related Origin Requests document served at the '/.well-known/webauthn' endpoint, and native application, browser extension, and local document facets are permitted via opaque origins which are matched directly rather than declared in that document.

Configuration validation rejects anything the WebAuthn provider would only reject at request time, including opaque origins no client conveys, origins which cannot be declared in the document, and origin sets which exceed the registrable domain label limit clients process.

Closes #12495, Closes #13146

PR #12496, opened 2026-07-12
feat(webauthn): configurable additional origins

### LINKED ISSUE #12495: Support configurable additional WebAuthn origins (native app facet origins, e.g. android:apk-key-hash)
### Description

Authelia builds its list of allowed Web Authentication (WebAuthn) origins as exactly one
entry, the portal origin derived from the forwarded request:

```go
// internal/middlewares/authelia_context.go, GetWebAuthnProvider()
config := &webauthn.Config{
    RPID:      origin.Hostname(),
    RPOrigins: []string{origin.String()},
    ...
}
```

Passkey assertions from native apps carry an app facet origin in their signed
`clientDataJSON` rather than an `https://` one, so they always fail validation:

```
error="Error validating origin (verification_error):
Expected Values: [https://auth.example.com],
Received: android:apk-key-hash:<base64url-sha256-of-apk-signing-cert>"
```

A config option for extra allowed origins, matched verbatim, would fix this:

```yaml
webauthn:
  additional_origins:
    - 'android:apk-key-hash:-sYXRdwJA3hvue3mKpYrOZ9zSPC7b4mbgzJmdZEDO5w'
```

### Use Case

Any Android app that runs a login flow against an Authelia portal, whether it embeds the
portal in a WebView (`WEB_AUTHENTICATION_SUPPORT_FOR_APP`) or calls the Credential
Manager application programming interface (API) natively.

The operator authorises the app by hosting a Digital Asset Links (DAL) file on the
Relying Party (RP) domain, which for Authelia is the portal's own domain. The ceremony
then works end to end, except that the phone's credential provider stamps the assertion
with the app facet origin instead of the web origin. That is documented Android
behaviour, not a provider quirk: Google's reference implementation derives a non-browser
caller's origin as `android:apk-key-hash:<sha256-of-signing-cert>`, and Google's
server-side passkey guide tells relying parties to "accept all the origins as valid on
the app server". The origin sits inside the signed `clientDataJSON`, so no reverse proxy
can rewrite it; the server's allowed-origins list is the only place it can be accepted.
iOS is unaffected, since the operating system there verifies the app-to-domain
association itself and stamps the normal https origin.

Accepting the key hash of your own app's signing certificate is the same trust statement
the operator already makes by hosting the DAL file.

### Details

This is a recurring requirement wherever native Android apps meet a WebAuthn server,
and there is plenty of prior art:

- go-webauthn, the library Authelia uses, merged purpose-built native app origin
  matching in go-webauthn/webauthn#468 (following go-webauthn/webauthn#92): `RPOrigins`
  entries that are not web origins are compared by exact string match. Nothing in the
  library needs to change; only the ability to add entries to the slice is missing.
- Keycloak added a configurable "custom FIDO2 origins" WebAuthn policy field for exactly
  this (keycloak/keycloak#23155, merged in keycloak/keycloak#23156).
- web-auth/webauthn-framework shipped support in v5.2.0
  (web-auth/webauthn-framework#393); Hanko documents `android:apk-key-hash:` entries in
  `webauthn.relying_party.origins`; SimpleWebAuthn covers it via its `expectedOrigin`
  array. The mismatch has been reported since 2020 against Google's own sample code
  (android/security-samples#71).
- The Related Origin Requests (ROR) work in #11432 does not cover this: its validator
  requires each origin to resolve to a hostname matching a
  `session.cookies[].authelia_url`, which an opaque facet string can never do.

### Documentation

- Google's server-side instruction to allow these origins (section "Verify and save the
  public key on the app server"):
  https://developer.android.com/identity/passkeys/create-passkeys
- go-webauthn native app origin support:
  https://github.com/go-webauthn/webauthn/pull/468
- Keycloak's equivalent feature:
  https://github.com/keycloak/keycloak/issues/23155
- WebAuthn in Android WebViews for non-browser apps:
  https://developer.android.com/identity/sign-in/credential-manager-webview
- Credential provider reference implementation deriving the facet origin
  (`appInfoToOrigin()`):
  https://developer.android.com/identity/sign-in/credential-provider
- Earliest report of the mismatch, against Google's own sample code:
  https://github.com/android/security-samples/issues/71

### Generative AI

Yes

### Pre-Submission Checklist

sherrmann:
This PR was mostly created using Claude Fable, but I have nonetheless put a lot of work
into iterating on, testing, and validating it.

Adds a `webauthn.additional_origins` configuration option. The listed origins are added
to the go-webauthn `RPOrigins` slice after the portal origin that Authelia works out
from the request. Each entry is compared as an exact string, which makes it possible to
allow origins from native apps such as
`android:apk-key-hash:<base64url-sha256-of-apk-signing-cert>`.

Until now the allowed-origins list only ever contained the portal origin. When an
Android app runs a login flow, the phone writes the app's origin into the signed
`clientDataJSON` instead of the portal's `https://` origin, so the ceremony always
failed with an origin verification error. Because that value is part of the signed
data, no proxy can rewrite it; the server's allow list is the only place it can be
accepted. go-webauthn already compares such entries as exact strings
(go-webauthn/webauthn#468), so only the ability to add entries was missing.

Validation rejects empty and duplicate entries, reusing the existing `validateList`
helper and error message style. When the option is not set, behaviour is unchanged and
the middleware test now checks this explicitly. The documentation includes a security
note that every listed origin is fully trusted, so operators should only add origins
they control, such as the one derived from their own app's signing certificate. The
configuration templates, keys, and JSON schemas were regenerated with authelia-gen.

Closes #12495


james-d-elliott:
see #11432

PR #12515, opened 2026-07-14
docs: update forwarded headers

LilPhyziX:
This is to update the X-Forwarded-For headers documentation for Cloudflare, after Cloudflare has updated their dashboard and fundamentally changed a few settings, separating removing the rule entirely and still allowing users to make a separate transform rule to allow specific IPs to modify the header.


james-d-elliott (approved):
LGTM

autheliabot (approved):

PR #12521, opened 2026-07-15
fix(authentication): recursive read lock deadlock in file user db save

RealFakeAccount:
## Description

`FileUserDatabase.Save` takes a read lock and, while holding it, calls `ToDatabaseModel`, which acquires the **same** `*sync.RWMutex` a second time:

```go
func (m *FileUserDatabase) Save() (err error) {
	m.RLock()
	defer m.RUnlock()
	if err = m.ToDatabaseModel().Write(m.Path); err != nil { // ToDatabaseModel() calls m.RLock() again
		return err
	}
	return nil
}
```

Go's `sync.RWMutex` is not reentrant. Per the documentation, *"if a goroutine holds a RWMutex for reading and another goroutine might call Lock, no goroutine should expect to be able to acquire a read lock until the initial read lock is released"* — a blocked writer parks subsequent `RLock` calls. So if a concurrent writer calls `Lock` (via `SetUserDetails`, used by password changes at `file_user_provider.go`, or via `Load` from the on-disk file `Reload`) in the window between `Save`'s outer `RLock` and the inner `RLock` in `ToDatabaseModel`, the inner `RLock` blocks behind the pending writer while the writer waits for `Save`'s outer read lock — a permanent deadlock. Because every read/write shares this one mutex, that wedges all file-backend authentication (`CheckUserPassword`, `GetDetails`, subsequent `Save`/`Load`).

## Fix

`ToDatabaseModel` already snapshots `Users` under its own read lock, and `Write` then serializes that private copy — no shared state is touched outside a lock. So `Save` does not need to hold the mutex at all. Removing the redundant outer lock means the mutex is acquired exactly once, and it is no longer held across YAML marshaling and file I/O (a nice side benefit for write latency under concurrency).

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## How has this been tested?

`go test ./internal/authentication/` passes. `Save` is exercised by the existing `ChangePassword`/`UpdatePassword` suites, which remain green. A deterministic regression test for the deadlock isn't included because triggering it requires a writer to land in a narrow scheduling window between the two `RLock` calls; such a test would be inherently flaky and, on regression, hang rather than fail. The fix itself is a straightforward removal of a provably-redundant recursive read lock.

### DIFF
--- a/internal/authentication/file_user_provider_database.go
+++ b/internal/authentication/file_user_provider_database.go
@@ -56,10 +56,6 @@ type FileUserDatabase struct {
 
 // Save the database to disk.
 func (m *FileUserDatabase) Save() (err error) {
-	m.RLock()
-
-	defer m.RUnlock()
-
 	if err = m.ToDatabaseModel().Write(m.Path); err != nil {
 		return err
 	}
--- a/internal/authentication/file_user_provider_database_test.go
+++ b/internal/authentication/file_user_provider_database_test.go
@@ -5,7 +5,9 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/go-crypt/crypt"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


james-d-elliott (changes requested):
Since this is an alleged bug that has not been reported please write a test that will cause the described deadlock prior to this fix and that is resolved by this fix i.e. it should fail before this commit, and pass after this commit without alteration.

PR #12550, opened 2026-07-18
fix(handler): treat empty basic parameter as malformed

james-d-elliott:
This hardens the basic auth handling against potential future weaknesses where an empty username or password could result in an incorrect successful authentication. There is no known way for this to occur at this stage and this has been carefully interrogated. In addition it aligns it with the improvements and fixes made in 3adae9081c41032986c7dfabb74b3202a6359927 and factorizes the usage of the handleGetBasic validator.


internal/handlers/handler_authz_authn.go
```diff
@@ -423,9 +411,19 @@ func handleGetBasic(ctx AuthzContext, delayer middlewares.Delayer, authn *Authn,
...
+	username, password := authn.Header.Authorization.Basic()
+
+	if len(username) == 0 || len(password) == 0 {
+		return nil, authentication.NotAuthenticated, fmt.Errorf("failed to validate parsed credentials of %s header: the username or password was empty", header)
+	}
+
+	if details, err = ctx.GetUserProvider().GetDetails(username); err != nil || details == nil {
+		if err == nil {
+			err = authentication.ErrUserNotFound
+		}
```
Crowley723: I think we should split the two cases here. If the error isn't empty, we should determine what the error is (is it a connection failure or an actual not found error). If the details are nil, then the error should be set to not found.

Currently we are ignoring what the error is and always logging it as a user not found error.

Crowley723 (commented):
Looks good, one potential change.

Crowley723 (approved):

autheliabot (approved):

nightah (approved):
LGTM.

PR #12618, opened 2026-07-29
docs(oidc): update nextcloud

mrrudy:
Two issues fixed in the Nextcloud OpenID Connect Login App integration guide:

1. **Boolean vs string type mismatch:** The expression `"nextcloud-admins" in groups` returns a boolean (true/false), but Nextcloud's oidc_login app expects `?string` for the `is_admin` attribute, causing a TypeError crash with HTTP 500 on login. Fixed by using a ternary expression that returns string values instead: `"nextcloud-admins" in groups ? "yes" : "no"`

2. **Empty custom_claims dict not resolving definitions-based attributes:** Using `is_nextcloud_admin: {}` fails config validation in Authelia v4 with 'attribute name is not a known attribute'. Fixed by explicitly specifying the `attribute` field to reference the user_attribute definition.


autheliabot (dismissed):

james-d-elliott (approved):

docs/content/integration/openid-connect/clients/nextcloud/index.md
```diff
@@ -83,15 +84,16 @@ which will operate with the application example:
 definitions:
   user_attributes:
     is_nextcloud_admin:
-      ## Expression to evaluate admin privilege for Nextcloud.
-      expression: '"nextcloud-admins" in groups'
+      ## Expression to evaluate admin privilege for Nextcloud. Must return a string, not a boolean.
```
james-d-elliott:
```suggestion
      ## Expression to evaluate admin privilege for Nextcloud.
```

james-d-elliott (changes requested):

james-d-elliott (approved):

autheliabot (approved):

PR #12704, opened 2026-08-10
fix: data structure issues

james-d-elliott:
This fixes general data structure and mapping issues across the repository. This includes several marshaling issues where data was marshaled when it shouldn't be, marshaled with an incorrect value like the disabled state of file provider user accounts, etc.


nightah (approved):
LGTM.

autheliabot (approved):

PR #12705, opened 2026-08-10
fix(configuration): add authelia.pam to scopes in schema

hendrik1120:
This adds the authelia.pam scope as an allowed value in the json schema


james-d-elliott (approved):

nightah (approved):
LGTM.

PR #12724, opened 2026-08-12
test(suites): improve reliability and failure diagnostics

nightah:
Suite failures were absorbed by a blanket automatic retry, which meant flakiness had no measurement and no record of what failed. Pulling 300 builds from the Buildkite API showed the scale of it: 3,666 suite jobs carried 3,237 automatic retries, only 52% passed on the first attempt, and the retry rate scaled with host contention from 10.5 per build when nothing else was running to 67 when a dozen builds overlapped.

Emit `go test -json` and upload it so the flake rate can be measured, bound the retries, and capture a screenshot, the page markup, the resource timing and every container's logs whenever a test fails, echoing the container logs through the test so they reach the failure body rather than a separate process the log cannot be attributed from. Fix the races this surfaced: an element helper that never waited, a one-time password secret read before it was populated, a notification asserted after the redirect that destroys it, and a navigation with no load gate.

Address the contention the measurements pointed at. The containers declared no limits, so every containerised Go runtime sized itself for the whole host, and the six second default read timeout turned a scheduling stall into a 408 and a blank login page. Gate the environment on healthchecks under CI and declare the dependencies the backend needs, since a failed startup check is fatal and a backend started before its directory or database is ready crash loops until that dependency appears.

Scope the remaining log based waits to entries produced after the action being waited on, because `docker compose logs` is cumulative and a wait could otherwise be satisfied by an entry from a previous test. Doing so exposed a high availability test that had been passing for the wrong reason, which now resolves the redis role from the sentinel rather than assuming the order tests happen to run in.

The last three builds completed with no retries at all across all 28 suite jobs.

Closes #12720


internal/suites/external_main_test.go
```diff
@@ -31,7 +31,11 @@ func TestMain(m *testing.M) {
 		os.Exit(130)
 	}()
 
-	os.Exit(m.Run())
+	code := m.Run()
+
+	closeSharedBrowsers()
+
+	os.Exit(code)
```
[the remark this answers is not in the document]
nightah: Both external suites create their session via RodSessionWithoutDevtools(), which sets disableDevtools = true. That takes the dedicated branch in NewRodSession, returning a browser that is never registered in the sharedBrowsers map.

internal/suites/suite_duo_push_test.go
```diff
@@ -289,8 +288,12 @@ func (s *DuoPushWebDriverSuite) TestShouldFailSelectionBecauseOfSelectionBypasse
 
 	s.doLoginOneFactor(s.T(), s.Context(ctx), "john", "password", false, BaseDomain, "")
 	s.doChangeMethod(s.T(), s.Context(ctx), "push-notification")
-	s.doClickButton(s.T(), s.Context(ctx), "selection-link")
-	s.verifyNotificationDisplayed(s.T(), s.Context(ctx), "Device selection was bypassed by Duo policy")
+
+	// An allowed pre-auth marks the second factor successful as it raises the notification, so the
+	// redirect that follows removes the notification before it can be looked for.
```
james-d-elliott:
```suggestion
```

internal/suites/action_visit.go
```diff
@@ -1,25 +1,39 @@
...
 func (rs *RodSession) doCreateTab(t *testing.T, url string) *rod.Page {
-	p, err := rs.WebDriver.MustIncognito().Page(proto.TargetCreateTarget{URL: url})
-	assert.NoError(t, err)
+	incognito, err := rs.WebDriver.Incognito()
+	require.NoError(t, err)
+
+	rs.contexts = append(rs.contexts, incognito)
+
+	p, err := incognito.Page(proto.TargetCreateTarget{URL: url})
+	require.NoError(t, err)
```
james-d-elliott:
```suggestion
	type tab struct {
		page *rod.Page
		err  error
	}

	created := make(chan tab, 1)

	go func() {
		browser, err := rs.WebDriver.Incognito()
		if err != nil {
			created <- tab{err: err}

			return
		}

		page, err := browser.Page(proto.TargetCreateTarget{URL: url})

		created <- tab{page: page, err: err}
	}()

	select {
	case result := <-created:
		require.NoError(t, result.err)
		require.NotNil(t, result.page)

		return result.page
	case <-time.After(time.Second * 10):
		require.FailNowf(t, "timeout creating tab", "the tab at '%s' was not created within 10 seconds", url)

		return nil
	}
```

james-d-elliott (dismissed):
More of a WDYT.

autheliabot (dismissed):

james-d-elliott (dismissed):

autheliabot (approved):

james-d-elliott (approved):

PR #12726, opened 2026-08-13
test(suites): support concurrent runs on a shared docker daemon

nightah:
Each Buildkite agent runs its own Docker daemon, so `.buildkite/hooks/pre-exit` prunes the entire daemon after every job and each of the ~50 agents on a host re-unpacks the same images from the pull-through registry. Docker does not support two daemons sharing a `data-root`, so storing each image once per host means one daemon per host, and that only works once concurrent jobs stop colliding on the resources they currently assume they own alone.

Every suite used the compose project `authelia` and the subnet `192.168.240.0/24`. Both are now driven by `COMPOSE_PROJECT_NAME` and `SUITE_SUBNET`, which `bootstrap.sh` derives from a per-agent `SUITE_SLOT`, and the Traefik Docker provider is constrained to its own project so it stops discovering other jobs' containers. That constraint is given as command arguments and `static.yml` is retired, because Traefik treats its three static configuration sources as mutually exclusive and a `--configFile` makes it ignore every other argument. The files compose cannot interpolate follow the same variable: the `hostEntries` table, the redis `sentinel.conf` and `slave.conf` templates via `entrypoint.sh`, and `NetworkACL/configuration.yml` through Authelia's own `template` configuration filter. The matching in `prepareHostsFile` also moves from a substring test to a field comparison, since `mail.example.com` otherwise matched the `mx1.mail.example.com` line and left the real entry pointing at the previous subnet.

Chrome and the test process both run inside the agent container, and on a shared daemon the suite network belongs to the host namespace rather than the agent's, so `Up()` attaches the agent to `${COMPOSE_PROJECT_NAME}_authelianet` and `Down()` detaches it before teardown. `SUITE_TMP` is mounted so that the host, the agent and the suite containers all resolve one path, which is what the CLI suite relies on when a command writes a file in one and reads it from the other. `SUITE_IMAGE` replaces the single mutable `authelia:dist` tag with a per slot one whose lifetime matches its owner, the phpldapadmin and envoy admin ports are offset per slot, and the crossbuild container loses its fixed name and takes its toolchain from `/buildkite/.tools` rather than `/usr`, because both the name and those mount sources belong to the host now rather than to the agent that requested them.

Agents that opt in with `CLEANBUILD=false` no longer run `docker system prune -af --volumes` or remove every container on the daemon; they remove only what their own compose project created, including the images it built. Because the daemon is no longer wiped between jobs, the integration images gain a `pull_policy` so a mutable tag cannot freeze at whatever a host first pulled.

Alongside the migration this fixes several failures it exposed. A suite that dies during startup now has its container logs collected before teardown, rather than leaving nothing but an exit code. Clicks locate their element again when the document replaced it, which was failing the change password scenario with a detached node. Incognito browser contexts are tracked so they are disposed rather than accumulating for the life of the shared browser. The reset password view guards against initiating twice, and the development frontend pins pnpm and is given enough memory to build. Every variable falls back to its current value, so local development and the existing agents are unchanged and the migration itself becomes a configuration change on the agent side.


james-d-elliott (dismissed):
LGTM

autheliabot (dismissed):

james-d-elliott (dismissed):

james-d-elliott (dismissed):

autheliabot (dismissed):

james-d-elliott (approved):

autheliabot (approved):

PR #12729, opened 2026-08-13
test(model): prevent drift for user model

james-d-elliott:


Crowley723 (commented):
What about embedding the UserDetails model in the UserDetailsExtended then using UserDetailsExtended everywhere?

james-d-elliott:
The main reason I decided not to do that last year was import cycle chances. I can revisit it to double check if this is a real possibility, in any event the test is probably worthwhile to add as we'd be aliasing it in that case.

Crowley723 (approved):

autheliabot (approved):

PR #12730, opened 2026-08-14
ci(buildkite): localised node artifacts

nightah:
The coverage build image tarball is the largest artifact in the pipeline and every `:selenium:` suite step has to download it before it can load the image, so with the number of suites we run the fetch from the shared Buildkite artifact store dominates the setup time of each of those steps.

A `post-artifact` hook is added so that once the `:docker: Build Image [coverage]` step has uploaded normally on a `tabris` agent, a `-tabris` suffixed copy of `authelia-image-coverage.tar.zst` is also uploaded to a node-local S3 bucket via the `BUILDKITE_ARTIFACT_UPLOAD_DESTINATION` and `BUILDKITE_S3_*` overrides, placing the artifact alongside the agents that will consume it.

The `pre-command` hook then prefers that localised copy on `tabris` agents, renaming it back to the canonical name so the remainder of the hook (the Kubernetes path, the slot aware `docker load`, and the integration container rewrites) is untouched. If the localised download fails, because the image was built on another node or the bucket is unavailable, it falls back to the default artifact store, and agents outside the pool keep the original single download path.


james-d-elliott (dismissed):
LGTM

Crowley723 (dismissed):

james-d-elliott (approved):

autheliabot (approved):

PR #12755, opened 2026-08-16
fix(authentication): stale user database aliases

james-d-elliott:
The alias based lookups for the user database were never reset on reload. This could result in a lookup not returning a helpful error.


autheliabot (approved):

nightah (approved):
LGTM.

PR #12756, opened 2026-08-16
fix(configuration): jwk type direct not usable

james-d-elliott:
This fixes an issue where the JWK direct algorithm cannot be used. This is caused by an inverse check of the length which incorrectly validates that it is less than 32 bytes whereas it should be at least 32 bytes.


autheliabot (approved):

nightah (approved):
LGTM.

PR #12761, opened 2026-08-16
fix(logging): incorrect multi writer order

james-d-elliott:
This fixes an issue where the io.MultiWriter wrote to files before os.Stdout. This ensures os.Stdout is always written to regardless of the file status.


autheliabot (approved):

nightah (approved):
LGTM.

PR #12765, opened 2026-08-17
fix(random): nominal entropy reduction

james-d-elliott:
The original implementation of the random bytes nominally reduced effective entropy by a fraction of a bit which is negligible. This fixes that issue by introducing a helper which performs character selection via rejection sampling instead.


autheliabot (dismissed):

nightah (approved):
LGTM.

autheliabot (approved):

PR #12772, opened 2026-08-17
feat(systemd): systemd notifier

### LINKED ISSUE #12348: Support `sd_notify` to signal when Authelia is ready to serve
### Description

Podman's Quadlets accept the `Notify=true` option, that makes them wait for a process to send a readiness signal before considering containers as running. This is especially useful for the dependency chain, where other containers are not started before Authelia is actually ready to serve. Without any `Notify=` option, Podman considers a container as started as soon as the process starts, but this doesn't take into account the setup time.

To be honest, Podman also supports `Notify=healty`, that waits for a healthcheck command to return a success. This is not as optimal as using [`sd_notify`](https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html) because it's a synchronous operation (must be run each X seconds), and likely more stressful on the system.

 As a reference, [Caddy supports `sd_notify`](https://github.com/caddyserver/caddy/blob/51a4bde17640dfc1f25342f74c507cbdec41472b/notify/notify_linux.go#L29), without depending on external packages, but there is a convenient [package](https://pkg.go.dev/github.com/coreos/go-systemd/v22@v22.7.0/daemon) ready-to-use if you need it.

### Use Case

In Podman Quadlets, `sd_notify` would allow Authelia to signal when it's ready to serve in an asynchronous fashion, so that other Quadlets that depend on it can be fired at the right time.

### Details

_No response_

### Documentation

- [Podman options](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#notify-defaults-to-false)
- [`sd_notify`](https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html)
- The [go-systemd](https://pkg.go.dev/github.com/coreos/go-systemd/v22@v22.7.0/daemon) package

### Generative AI

No

### Pre-Submission Checklist

james-d-elliott:
This implements the systemd notifier tooling.

Closes #12348

### DIFF
--- a/internal/systemd/monotonic_linux.go
+++ b/internal/systemd/monotonic_linux.go
@@ -1,0 +1,28 @@
+// SPDX-FileCopyrightText: 2026 Authelia
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build linux
+
+package systemd
+
+import (
+	"syscall"
+	"time"
+	"unsafe"
+)
+
+// clockMonotonic is the CLOCK_MONOTONIC clock identifier. Lives here since this is the only file that uses it and it
+// is build gated.
+const clockMonotonic = 1
+
+// monotonic returns the current CLOCK_MONOTONIC time in microseconds.
+func monotonic() (usec int64, ok bool) {
+	var ts syscall.Timespec
+
+	if _, _, errno := syscall.Syscall(syscall.SYS_CLOCK_GETTIME, clockMonotonic, uintptr(unsafe.Pointer(&ts)), 0); errno != 0 {
+		return 0, false
+	}
+
+	return int64(time.Duration(ts.Nano()) / time.Microsecond), true
+}
--- a/authelia.service
+++ b/authelia.service
@@ -8,11 +8,17 @@ Documentation=https://www.authelia.com
 After=multi-user.target
 
 [Service]
+Type=notify
+NotifyAccess=main
 User=authelia
 Group=authelia
 UMask=027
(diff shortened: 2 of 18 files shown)


internal/systemd/monotonic_linux.go
```diff
@@ -0,0 +1,24 @@
...
+
+// clockMonotonic is the CLOCK_MONOTONIC clock identifier. Lives here since this is the only file that uses it and it
+// is build gated.
+const clockMonotonic = 1
+
+// monotonic returns the current CLOCK_MONOTONIC time in microseconds.
+func monotonic() (usec int64, ok bool) {
+	var ts syscall.Timespec
+
+	if _, _, errno := syscall.Syscall(syscall.SYS_CLOCK_GETTIME, clockMonotonic, uintptr(unsafe.Pointer(&ts)), 0); errno != 0 {
```
livingsilver94, line 19: You do not need to introduce new code for the monotonic clock. [Documentation](https://pkg.go.dev/time#hdr-Monotonic_Clocks) says the monotonic clock is embedded into `time.Time` and used when computing Durations, so I think you just need to store the initial time.Time when the process starts and compute differences when required.

james-d-elliott: You actually do. 

The monotonic itself is not exposed by the API and it's not the value expected by systemd even if it was as it stores the process monotonic. The expected value is the current system monotonic (as identified by `CLOCK_MONOTONIC`)  in microseconds. Using the process start will cause errorless hangs. We could calculate it from `/proc/uptime`, but that's hacky.

PR #12774, opened 2026-08-17
docs(oidc): update immich

fredrikekre:
Immich support PKCE (since v1.132.0) and `client_secret_basic` (since
v1.133.0). The `.well-known` suffix is optional and added automatically
during discovery, looks cleaner without it IMO.


james-d-elliott (approved):
LGTM

autheliabot (approved):

james-d-elliott:
Also discovery is only meant to occur using the issuer URL, which does not include the well known, so good change.

PR #12796, opened 2026-08-19
test(suites): support concurrent runs across git worktrees

nightah:
Sharing a Docker daemon between suite runs was made safe by #12726, but only for CI, where each job runs the tests and Chrome inside its own agent container. That container is what supplies the two resources the suites still assume they own alone: its own `/etc/hosts` and its own `/tmp`. A development machine has no agent container, so several worktrees on it collide on both and only one can run a suite at a time.

The `hostEntries` table moves from `cmd/authelia-scripts/cmd/bootstrap.go` to `internal/suites/hosts.go`, so `/etc/hosts`, Chrome and the Go clients all describe the network from one source. Chrome resolves the suite domains from `--host-resolver-rules` and the Go clients resolve them in `DialContext`, which `NewHTTPTransport` and `dialSSHContext` now dial through. Resolving at the dial layer rather than by name leaves the SNI name, the `Host` header and the cookie domain as the name the caller asked for, and it means a slotted shell no longer has to write `/etc/hosts` at all. Leaving that file on the default subnet is what keeps an unslotted worktree browsable while slotted ones run beside it.

`SUITE_TMP` moved the host side of the `/tmp` mounts but no Go code read it, so the test process still named the containers' path. `SUITE_TMP_PATH` now names where that process finds the shared directory while the containers keep seeing it at `/tmp`, and the host side of every exchange goes through `SuiteTmpPath`. `t.TempDir()` was the one place a single path crossed the boundary, being written by a container and then read back here, and it also honors `TMPDIR`, which the containers do not share; `SuiteTempDir` and `SuiteTmpContainerPath` replace it so each side is named explicitly.

CI supplies `SUITE_SLOT` per agent and nothing supplied it locally. `authelia-scripts suites slot` hands each worktree the lowest free number under an `flock` on a registry keyed by worktree path, and drops entries whose directory is gone so a deleted worktree frees its slot without anyone remembering to. `bootstrap.sh` asks for one when a shell is unslotted outside CI, and derives `SUITE_TMP` and `SUITE_TMP_PATH` from it there, since only a local shell has to move the path it reads through as well.

Every variable still falls back to the value a single local run has always used, so leaving them unset is byte for byte the existing behaviour.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -30,24 +30,43 @@ else
   export SUITE_PULL_POLICY=always
 fi
 
-# Agents that share a Docker daemon are each given a slot so their suites cannot collide on the compose project, the
-# network subnet or the debug ports. Everything falls back to the historical values when the slot is unset.
-if [[ -n "${SUITE_SLOT}" ]]; then
-  export COMPOSE_PROJECT_NAME="authelia-${SUITE_SLOT}"
-  export SUITE_SUBNET="10.240.${SUITE_SLOT}"
-  export LDAP_ADMIN_PORT="$((9090 + SUITE_SLOT))"
-  export ENVOY_ADMIN_PORT="$((9901 + SUITE_SLOT))"
-fi
-
-if [[ -n "${SUITE_TMP}" ]]; then
-  mkdir -p "${SUITE_TMP}"
-fi
-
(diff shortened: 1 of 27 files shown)


james-d-elliott (approved):
LGTM

autheliabot (approved):

PR #12804, opened 2026-08-20
fix: correct defects across the web frontend and the suites

nightah:
A review of the frontend refactor turned up a set of defects that share no single cause but do share a shape: behaviour that is wrong only at the edges, where nothing exercised it. `CopyButton` never handled a rejected clipboard write, so a denied permission or an insecure context left the button spinning with no way back, and the timers it scheduled after the write outlived the component because they were created once the unmount cleanup had already run. `Progress` divided by a maximum of zero, let a value outside the range escape the track, ignored a custom minimum entirely, and reported a maximum to assistive technology that the bar itself had already normalised away. `LinearProgressBar` treated an explicit height of zero as absent. `Authenticated` reused the identifier `MinimalLayout` already renders, so the page carried the same id twice. The One-Time Password panel let registration start before the configuration it depends on had loaded, and its registration dialog kept the spinner and the error icon inside the branch that requires a secret, so the initial request and every failure showed an empty box, and the blur that softens the code behind them covered them too.

The accessibility work is of the same kind. Four password visibility toggles had drifted apart: each was missing a different one of a translated label, Enter activation and a pressed state, and since they are press and hold controls with no click handler, Enter reached none of them. They now match `PasswordForm`, which was the closest to correct. `PasskeyIcon` is decorative next to the button's own label and is hidden from assistive technology accordingly.

Where a defect had siblings the review did not name, those are fixed too, because a fix applied to one of several identical call sites reads as a deliberate distinction rather than an oversight. That covers the `Spinner` sizing, which sets the wrapper through `className` while the icon keeps its own default and overflows, and the hardcoded border colour that does not follow the theme.

`TwoFactorAuthenticationOptionsMethodsRadioGroup` built an object with a single `target.value` field and cast it to a `ChangeEvent`, which the radio group never produces. The contract is now the value itself and both panel handlers take it directly.

The suites gain their own `node_modules` for the dev frontend. The container is alpine and a development machine is usually glibc, so bind mounting `web` into it left the two installs writing different native binaries over each other in one tree, and their `.modules.yaml` disagreed on both the store path and the pnpm version, which is why the entrypoint deletes that file. An anonymous volume separates the two trees, and the mount point has to be created and owned in the image because the volume is seeded from it while the layer that creates it necessarily runs as root. The shared store is still reached and still hits: an install reuses every package and downloads none. Alongside that `doEnterOTP` counts and types only the editable slots, which is what the predicates that read the field back already did. The field also renders a hidden input carrying the whole value, and counting it let a field one slot short satisfy the assertion meant to catch exactly that, while typing a short code failed later on the value check and reported the symptom rather than the cause.

Every fix that was not purely visual is covered by a test, and each of those tests was confirmed to fail without its fix.


james-d-elliott (approved):
LGTM

autheliabot (approved):

PR #12815, opened 2026-08-21
test(suites): fetch a page again when it loses a response

nightah:
A `PathPrefix` run failed with the portal never rendering. The document was served, 10 of the 27 chunks the entry module preloads returned `200`, and the other 17 recorded a zero `responseStatus` against a zero `encodedBodySize`, which is what a request that never received a response looks like. The module graph therefore never resolved: `#root` stayed empty, `#first-factor-stage` never appeared, and the test spent its whole budget waiting for an element that could not arrive. No wait recovers that page, because the document has loaded and nothing further is ever fetched, so `doVisit` now fetches it once more when it finds an empty root alongside a response that was lost. A page that is merely slow, or one that was answered and still did not render, is left alone so that no visit is made worse by this.

The proxy the suites run behind was starved at the time. It logged a docker event stamped `22:43:54` at `22:44:11`, 17 seconds late, and the configuration rebuild it was working through enumerated the 199 containers the shared daemon was holding for every other suite on the machine. `--providers.providersthrottleduration` coalesces those events so that work stays off the two cores the container has to proxy the suite with.

Whether any of the 17 requests reached the proxy at all is a question only its access log answers, and that log was not collectable. It shared stdout with the debug log, where one rebuild emits a line per container the daemon holds, so the 200 lines a failing test collected held no access log line at all. Measured against a quiet daemon the same tail also held none of the `Remote error` lines the `forwardAuth` middleware logs and none of the routing configuration, so raising the level to `INFO` would have traded one missing record for another. The access log moves to a file in the suite directory instead, which a failing test collects whole and `Up` discards from the run before, and the tail of stdout carries the debug log alone at a depth one rebuild cannot exhaust.

The suites are swept of the comments that restated the code they sat above. The GoDoc of every unexported function goes, since the name says what the header did, along with the narration in the scenarios, where `// Logout.` sat above `doLogout` and `// Login one factor.` above `doLoginOneFactor`. What a reader cannot get from the code stays, moved to the line it explains: why the previous run's access log is removed, why a missing one is not an error, why `rod` is asked for the text as rendered, and how many failures it takes for the browser to give up on a chunk.

### DIFF
--- a/.buildkite/steps/e2etests.sh
+++ b/.buildkite/steps/e2etests.sh
@@ -19,12 +19,13 @@ cat << EOF
   - label: ":selenium: ${SUITE_NAME} Suite"
     command: "authelia-scripts --log-level debug suites test ${SUITE_NAME} --failfast --headless"
     artifact_paths:
+      - "screenshots/**/*.access.log"
+      - "screenshots/**/*.containers.log"
+      - "screenshots/**/*.html"
+      - "screenshots/**/*.png"
+      - "screenshots/**/*.resources.json"
       - "test-results-*.json"
       - "test-results-*.xml"
-      - "screenshots/**/*.png"
-      - "screenshots/**/*.html"
-      - "screenshots/**/*.resources.json"
-      - "screenshots/**/*.containers.log"
     timeout_in_minutes: ${TIMEOUT}
     retry:
       automatic:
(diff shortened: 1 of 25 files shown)


james-d-elliott (approved):
LGTM

autheliabot (approved):

PR #12816, opened 2026-08-21
perf(server): serve pre-compressed embedded assets

nightah:
The portal is served entirely uncompressed. A first paint issues 28 requests for the entry script and its `modulepreload` graph which currently total 503,058 bytes on the wire, and the largest chunk alone is 821,230 bytes. Bursting that over a single connection is slow everywhere and is enough to stall a loaded reverse proxy mid-flight, leaving the module graph unresolved and the portal unmounted.

Every compressible file in the embedded `public_html` filesystem is now compressed once during handler construction with brotli and gzip, and the resulting representation is selected per request from `Accept-Encoding`. Pre-compressing avoids paying the compression cost on each request, which matters most for exactly the large chunks that cause the problem. Neither level is the maximum as the returns diminish sharply: brotli level 11 takes roughly 32 times as long as level 6 while only producing output approximately 7% smaller. Compressing the whole embedded filesystem at level 6 costs around 69ms of startup time.

The first paint drops from 503,058 to 168,337 bytes with `Accept-Encoding: br`, and the whole static tree from 1,750,748 to 695,593 bytes. Formats which are already compressed such as images and fonts are excluded by extension, files smaller than 1KiB are skipped, and a representation is discarded if it isn't actually smaller than the original. Each representation carries its own `ETag` and responses which have compressible representations set `Vary: Accept-Encoding` so caches and conditional requests can't cross representations.

The templated assets are excluded. `public_html/index.html`, `public_html/api/index.html` and `public_html/api/openapi.yml` are rendered per request to carry the CSP nonce and the configuration of the running instance, so the embedded file is never the body of a response and a representation compressed ahead of time could never be sent. Only the extension decided this before, which left the two documents compressed at startup for nothing and left the specification uncompressed by the accident of `.yml` not being a compressible extension rather than by intent. None is large enough to be worth compressing on the fly.

`internal/templates` now names those paths as `AssetPathsTemplated` and loads them through it, and the server excludes exactly that list rather than a copy of it, so an asset added to the loader later cannot be pre-compressed by omission. The loader keeps the template names and error strings it had, which the existing tests assert verbatim.

### DIFF
--- a/internal/server/asset.go
+++ b/internal/server/asset.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/authelia/authelia/v4/internal/handlers"
 	"github.com/authelia/authelia/v4/internal/middlewares"
+	"github.com/authelia/authelia/v4/internal/templates"
 	"github.com/authelia/authelia/v4/internal/utils"
 )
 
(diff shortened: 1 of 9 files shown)


james-d-elliott (dismissed):

james-d-elliott (dismissed):
LGTM

autheliabot (dismissed):

james-d-elliott (approved):

autheliabot (approved):

PR #12858, opened 2026-08-24
fix(handlers): use regeneration more frequently

james-d-elliott:
This makes the use of the session regeneration feature more frequent using it at endpoints that in part benefit from this kind of hardening. This also removes the uses of ctx.Error() which was shadowing the RequestCtx version of it.


nightah (dismissed):
LGTM.

autheliabot (dismissed):

autheliabot (approved):

nightah (approved):
LGTM.

PR #12992, opened 2026-09-05
fix(oidc): use resource strategy to match granted resource

### LINKED ISSUE #12984: Bearer authz audience matching became exact-string in v4.39.21, breaking every path-based audience
### Version

v4.39.21, v4.39.22

### Deployment Method

Docker

### Reverse Proxy

Traefik

### Reverse Proxy Version

v3.7.12

### Description

**Disclosure:** this report was drafted with the assistance of generative AI, in accordance with the [Artificial Intelligence Policy](https://www.authelia.com/policies/artificial-intelligence/). It was used to diff `authelia.com/provider/oauth2` between `v0.2.31` and `v0.3.0`, to locate the call sites cited below, and to draft the text of this report. The failure itself was observed on a production instance before any of that analysis, every claim below was verified against the running instance and against the source at the tagged releases, and I have reviewed and understand the content.

---

Since `v4.39.21`, an `authelia.bearer.authz` token is denied for any request whose URL sits below a whitelisted audience. Only a request whose full URL, path and query string included, is byte-identical to an entry in the client's `audience` list is now authorized.

This contradicts the documented contract, which has been in place since the feature shipped in `v4.38.0` ([oauth-2.0-bearer-token-usage.md](https://www.authelia.com/integration/openid-connect/oauth-2.0-bearer-token-usage/#general-protections)):

> The audience of the token is also considered and if the token does not have an audience which is **an exact match or the prefix of the URL being requested**, the authorization will automatically be denied.

and, further down the same page:

> These rules effectively give both administrators and end-users fine-grained control over which endpoints can utilize this authorization scheme as administrators will be required to allow **each individual URL prefix** which can be requested […]

Token issuance is unaffected: `POST /api/oidc/token` still returns `200` and grants the full audience list. Only the authorization step fails, so the symptom presents as a blanket `401` from the proxy with nothing logged above `debug`.

#### Root cause

`v4.39.21` moved `authelia.com/provider/oauth2` from `v0.2.31` to `v0.3.0` (#12922, commit `240bfed81d35`). That bump renamed the audience matcher and changed its semantics:

| | `v0.2.31` `DefaultAudienceMatchingStrategy` (`audience_strategy.go:19`) | `v0.3.0` `DefaultAudienceStrategy` (`strategy_audience.go:44`) |
|---|---|---|
| implementation | parse both sides as URLs; require equal scheme + host and that the requested path is the whitelisted path or a subpath of it | `slices.Contains(haystack, n)` |
| effect | exact match **or prefix** | exact string equality only |

The URL-aware logic was not deleted, it was relocated to `DefaultResourceStrategy` / `IsMatchingResourceIndicator` (`strategy_resource.go:73`), which is reached only for RFC 8707 `resource` indicators. The bearer authz path never calls it.

`internal/handlers/handler_authz_authn.go:638` still compares the whole request URL against the audience list using the *audience* strategy:

```go
audience := []string{object.URL.String()}      // e.g. https://app.example.com/ping/x
strategy := provider.GetAudienceStrategy(ctx)  // resolves to DefaultAudienceStrategy

if err = strategy(requester.GetGrantedAudience(), audience); err != nil {   // haystack is the granted entries
    return "", "", false, authentication.NotAuthenticated, fmt.Errorf("token does not contain a valid audience for the url '%s' with the error: %w", audience[0], err)
}
...
if err = strategy(client.GetAudience(), audience); err != nil {             // haystack is the registered entries
```

`object.URL` is built from `X-Forwarded-Proto` + `X-Forwarded-Host` + `X-Forwarded-Uri`, so `object.URL.String()` carries the path and query string. A client registered with `audience: ['https://app.example.com']` can therefore never satisfy either check for any request other than the bare origin.

#### Why this reads as accidental rather than intended

1. The documented behaviour above is unchanged on `master` and dates to `v4.38.0`.
2. In `v0.3.0`, `DefaultAudienceStrategy` and `ExactAudienceStrategy` are now functionally identical (both `slices.Contains`), while `ExactAudienceStrategy` retains a doc comment drawing a distinction that no longer exists: *"does not assume that audiences are URIs, but compares strings as-is […] Use this strategy when your audience values are not URIs"*.
3. `ResourceStrategy` in the same release carries a stale doc comment describing the behaviour that was removed: *"Defaults to DefaultAudienceStrategy (URL-based matching)."*

#### Scope and workaround

There is no configuration escape hatch:

- `internal/oidc/config.go:738` hardcodes `oauthelia2.DefaultAudienceStrategy`; no configuration key exposes it.
- `v0.3.0` added a per-client override via `AudienceStrategyProvider`, but `RegisteredClient` does not implement it and `handler_authz_authn.go` calls the config-level `provider.GetAudienceStrategy(ctx)` directly, so that path is unreachable.
- The check cannot be skipped. `needle` is always exactly one element, so the `len(needle) == 0` early return never fires; an empty `audience` list makes both checks fail rather than pass.

The only workarounds are to enumerate every requested URL verbatim in `audience`, query strings included, or to drop bearer authz for those endpoints via a `bypass` access control rule.

I do not believe #12973 addresses this. It changes the same line to `JoinGrantedAudienceAndResource(requester.GetGrantedAudience(), requester.GetGrantedResource())`, but `strategy` is still the exact-match audience strategy, so prefix matching does not return. `master` of `authelia/oauth2-provider` still has the `slices.Contains` implementation and none of its 7 open PRs touch it.

This is distinct from #12970 (RFC 8707 `resource` on the `authorization_code` token exchange) and #12974 (introspection endpoint client auth), though all three originate in the same `v0.3.0` bump.

### Reproduction

The client used below is registered for the `client_credentials` grant with `requested_audience_mode: 'implicit'`, so the token is granted the client's full audience list and the token request itself never exercises the audience strategy. Full configuration is in the Configuration section.

**Case A, the original failure.** Register `audience: ['https://app.example.com']` and request any path under that origin:

```bash
curl -i -H "Authorization: Bearer authelia_at_..." https://app.example.com/api/health
```

Result on `v4.39.20`: `200`. Result on `v4.39.21` and `v4.39.22`: `401`. `https://app.example.com/` also fails, because `object.URL.String()` normalises to a trailing slash and no longer matches the entry without one.

**Case B, isolating it to subpath matching.** This is the run captured in the logs below. Register the audience entry with a path, then request a subpath of it, which the `v0.2.31` matcher accepted as a prefix match:

```yaml
audience:
  - 'https://app.example.com/ping'
```

```bash
TOKEN=$(curl -s -u 'monitoring:<secret>' \
  -d 'grant_type=client_credentials&scope=authelia.bearer.authz' \
  https://auth.example.com/api/oidc/token | jq -r .access_token)

curl -s -o /dev/null -w '%{http_code}\n' -H "Authorization: Bearer $TOKEN" https://app.example.com/ping
# 200  -> exact match is accepted

curl -s -o /dev/null -w '%{http_code}\n' -H "Authorization: Bearer $TOKEN" https://app.example.com/ping/x
# 401  -> subpath is rejected
```

`/ping/x` is a subpath of the whitelisted `/ping`, so `DefaultAudienceMatchingStrategy` in `v0.2.31` would have authorized it and the backend would have answered (`404` in my case, since that route does not exist upstream). Under `v0.3.0` Authelia rejects it before the backend is reached.

### Expectations

`https://app.example.com` in a client's `audience` should authorize any request under that origin, as documented. The bearer authz path should keep URL prefix semantics, either by using `DefaultResourceStrategy` / `IsMatchingResourceIndicator` at `handler_authz_authn.go:638`, or by restoring the URL-aware implementation to `DefaultAudienceStrategy`.

Failing that, if exact matching is the intended new behaviour, the two documentation passages quoted above and the stale `ResourceStrategy` and `ExactAudienceStrategy` doc comments should be corrected, and the change called out as breaking in the release notes.

### Configuration (Authelia)

```yaml
log:
  level: 'debug'

identity_providers:
  oidc:
    clients:
      - client_id: 'monitoring'
        client_name: 'Monitoring'
        client_secret: '$pbkdf2-sha512$310000$...'
        scopes:
          - 'authelia.bearer.authz'
        grant_types:
          - 'client_credentials'
        requested_audience_mode: 'implicit'
        token_endpoint_auth_method: 'client_secret_basic'
        audience:
          - 'https://app.example.com/ping'

server:
  endpoints:
    authz:
      forward-auth-internal:
        implementation: 'ForwardAuth'
        authn_strategies:
          - name: 'HeaderAuthorization'
            schemes:
              - 'Bearer'
          - name: 'CookieSession'

access_control:
  default_policy: 'deny'
  rules:
    - domain:
        - 'app.example.com'
      subject: 'oauth2:client:monitoring'
      policy: 'one_factor'
      networks:
        - 'internal'

definitions:
  network:
    internal: '10.0.0.0/24'
```

### Build Information

```shell
Last Tag: v4.39.22
State: tagged clean
Branch: v4.39.22
Commit: 8626f138f8a39350457b105de0e97fdf9ebf4953
Build Number: 58087
Build OS: linux
Build Arch: arm64
Build Compiler: gc
Build Date: Thu, 03 Sep 2026 18:24:12 +1000
Development: false
Extra:

Go:
    Version: go1.27.1
    Module Path: github.com/authelia/authelia/v4
    Executable Path: github.com/authelia/authelia/v4/cmd/authelia
```

### Logs (Authelia)

Case B above, at `log.level: debug`, from startup through to the denial. Hostnames, client id and client IP are sanitized; nothing else is altered or removed.

Note that the failure is only visible at `debug`. Because the matched access control rule carries a subject (`oauth2:client:monitoring`), `ruleHasSubject` is true, so the branch at `internal/handlers/handler_authz.go:229` swallows the authentication error and logs it as *"the matched rule was a bypass rule"*, which is misleading here since the matched rule is a `one_factor` rule. At `info` the request produces no log line at all, only a `401` at the proxy.

```shell
time="2026-09-04T21:23:39+00:00" level=info msg="Authelia v4.39.22 is starting"
time="2026-09-04T21:23:39+00:00" level=info msg="Log severity set to debug"
time="2026-09-04T21:23:39+00:00" level=info msg="Storage schema is being checked for updates"
time="2026-09-04T21:23:39+00:00" level=info msg="Storage schema is already up to date"
time="2026-09-04T21:23:40+00:00" level=debug msg="webauthn-metadata provider: startup check skipped as it is disabled"
time="2026-09-04T21:23:41+00:00" level=info msg="Startup complete"
time="2026-09-04T21:23:41+00:00" level=info msg="Listening for non-TLS connections on '[::]:9091' path '/'" server=main service=server
time="2026-09-04T21:23:45+00:00" level=debug msg="Check authorization of subject username= groups= ip=10.0.0.10 and object https://auth.example.com/api/oidc/token (method POST)."
time="2026-09-04T21:23:46+00:00" level=debug msg="Access Request with id '9f3c1a04-6b2e-4d77-9c15-2a8be0d41f37' is being processed" method=POST path=/api/oidc/token remote_ip=10.0.0.10
time="2026-09-04T21:23:46+00:00" level=debug msg="Access Request with id '9f3c1a04-6b2e-4d77-9c15-2a8be0d41f37' on client with id 'monitoring' is being processed" method=POST path=/api/oidc/token remote_ip=10.0.0.10
time="2026-09-04T21:23:46+00:00" level=debug msg="Access Request with id '9f3c1a04-6b2e-4d77-9c15-2a8be0d41f37' on client with id 'monitoring' has successfully been processed" method=POST path=/api/oidc/token remote_ip=10.0.0.10
time="2026-09-04T21:23:46+00:00" level=debug msg="Check authorization of subject username= groups= ip=10.0.0.10 and object https://app.example.com/ping (method GET)."
time="2026-09-04T21:23:47+00:00" level=debug msg="Check authorization of subject username= groups= ip=10.0.0.10 and object https://app.example.com/ping/x (method GET)."
time="2026-09-04T21:23:47+00:00" level=debug msg="Error occurred while attempting to authenticate a request but the matched rule was a bypass rule" error="failed to validate Authorization header with bearer scheme: token does not contain a valid audience for the url 'https://app.example.com/ping/x' with the error: invalid_target" method=GET path=/api/authz/forward-auth-internal remote_ip=10.0.0.10
time="2026-09-04T21:23:47+00:00" level=debug msg="Responding 401 WWW-Authenticate" method=GET path=/api/authz/forward-auth-internal remote_ip=10.0.0.10
```

The token request at `21:23:46` succeeds and grants the audience. The whitelisted `/ping` at `21:23:46` produces an authorization check with no error and returns `200`. The subpath `/ping/x` at `21:23:47` produces the audience failure and a `401`.

### Logs (Proxy / Application)

Traefik access log for the same three requests, same sanitization, fields trimmed to the relevant ones:

```shell
{"ClientHost":"10.0.0.10","RequestHost":"auth.example.com","RequestPath":"/api/oidc/token","RequestMethod":"POST","DownstreamStatus":200,"RouterName":"secure@file"}
{"ClientHost":"10.0.0.10","RequestHost":"app.example.com","RequestPath":"/ping","RequestMethod":"GET","DownstreamStatus":200,"RouterName":"backend@docker"}
{"ClientHost":"10.0.0.10","RequestHost":"app.example.com","RequestPath":"/ping","RequestMethod":"GET","DownstreamStatus":200,"RouterName":"secure@file"}
{"ClientHost":"10.0.0.10","RequestHost":"app.example.com","RequestPath":"/ping/x","RequestMethod":"GET","DownstreamStatus":401,"RouterName":"secure@file"}
```

The `/ping` request appears twice because it reaches the backend router as well as the entrypoint router (specific to my setup). The `/ping/x` request has no backend entry: it is rejected at the forward-auth middleware and never routed upstream.

### Documentation

- Documented contract, unchanged since `v4.38.0`: https://www.authelia.com/integration/openid-connect/oauth-2.0-bearer-token-usage/#general-protections
- `v0.2.31` implementation (prefix matching): https://github.com/authelia/oauth2-provider/blob/v0.2.31/audience_strategy.go#L19
- `v0.3.0` implementation (exact matching): https://github.com/authelia/oauth2-provider/blob/v0.3.0/strategy_audience.go#L44
- Where the URL logic went: https://github.com/authelia/oauth2-provider/blob/v0.3.0/strategy_resource.go#L73
- Affected call site: https://github.com/authelia/authelia/blob/v4.39.22/internal/handlers/handler_authz_authn.go#L638
- Hardcoded strategy: https://github.com/authelia/authelia/blob/v4.39.22/internal/oidc/config.go#L738
- Misleading log branch: https://github.com/authelia/authelia/blob/v4.39.22/internal/handlers/handler_authz.go#L229
- RFC 6750: https://datatracker.ietf.org/doc/html/rfc6750
- RFC 8707: https://datatracker.ietf.org/doc/html/rfc8707

### Generative AI

Yes

### Pre-Submission Checklist

- [x] This is a bug report and not a support request

- [x] I have read the security policy and this bug report is not a security issue or security related issue

- [x] I have either included the complete configuration file or I am sure it's unrelated to the configuration

- [x] I have either included the complete debug / trace logs or the output of the build-info command if the logs are not relevant

- [x] I have provided all of the required information in full with the only alteration being reasonable sanitization in accordance with the [Troubleshooting Sanitization](https://www.authelia.com/r/sanitize) reference guide

- [x] I have checked for related proxy or application logs and included them if available

james-d-elliott:
This fixes an issue where the granted resource values were just compared against the audience strategy. Granted resources now are matched using the specific strategy specific to them. It's noteworthy that the behavior has changed as allowed under our version policy and that implicit audiences granted are no longer sufficient for prefix matching requested resources, and that now only explicitly requested resources using the `resource` parameter are compared using prefix match rules. 

This is to better align ourselves with the relevant specifications, where a requested audience is an exact match and a requested resource is permitted to be matched by server policy. Note that the relevant specifications for the `resource` parameter and `audience` parameter are RFC8707 and RFC8693 respectively.

Fixes #12984


nightah (dismissed):
LGTM.

autheliabot (approved):

nightah (approved):
LGTM.

PR #13093, opened 2026-09-10
feat(authentication): add registration custom url

### LINKED ISSUE #8949: Add customisation option for a registration URL
### Description

I'd simply like to be able to show an option for users to register an account, redirecting to an external service.

### Use Case

I've got a community that uses Authelia as it's SSO solution. New members do not automatically have an account created and needs to set it up through a web service, so I'd like to be able to easily direct them to that whenever they try to auth. This would be similar to the custom url option for password resets https://www.authelia.com/configuration/first-factor/introduction/#custom_url

### Details

_No response_

### Documentation

_No response_

### Generative AI

No

### Pre-Submission Checklist

nightah:
Deployments which provision accounts through a separate service had no way to point users at it. A user who reached the login page without an account had nothing to click, and operators were reduced to documenting the registration address out of band or customising the frontend.

This adds `authentication_backend.registration.custom_url`, mirroring the existing `authentication_backend.password_reset.custom_url` in both shape and validation: the value must carry an `http` or `https` scheme, and is threaded to the frontend through `TemplatedFileOptions` and the embedded `data-registrationurl` attribute rather than an API call, since it is static for the lifetime of the process.

When set, a `Register` link renders on the login page beside the reset password link and opens the configured URL in a new tab. When empty, nothing is rendered and the page is unchanged. Authelia gains no internal registration functionality from this; the option only surfaces a link to a service the operator runs separately, and accounts must still exist in the authentication backend before a user can sign in.

image

Closes #8949.


docs/static/schemas/latest/json-schema/configuration.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://www.authelia.com/schemas/v4.39/json-schema/configuration.json",
+  "$id": "https://www.authelia.com/schemas/v4.40/json-schema/configuration.json",
```
james-d-elliott, line 3: We should generate this file and other schema changes just before release to prevent breaking people.

james-d-elliott (changes requested):
LGTM other than the schema inclusion. I think it'd be easier to just leave all schema generation to after everything merges to avoid the need to resolve additional conflicts which we're going to have lots of to begin with.

james-d-elliott (approved):

autheliabot (approved):

PR #13094, opened 2026-09-10
feat(handlers): notify users when a password reset fails

### LINKED ISSUE #2926: Email user when get Error when attempting to reset password
## Feature Request

So current behavior when attempt to reset the new password you get the below popup which is expected.

Imgur

I understand that we don't want to give information to the "potential" attacker and that's fine.

But the popup msg of course does not help the user as to what they can do differently. 
So an idea was to send them an email stating what the log msg says.

### Description

Email the user with the error msg when they get a popup error msg when attempting to reset the password.

### Use Case

At least with this one case we had was that the user was using an old ldap password but the msg was not helpful.

```
  "level": "error",
  "method": "POST",
  "msg": "unable to update password. Cause: LDAP Result Code 19 \"Constraint Violation\": Password is not being changed from existing value",
  "path": "/api/reset-password",
```

An email to the user stating they are trying to use an old password would probably save them some headaches, as well as the admin :)

Hope I did this FR right.

Thank you for the awesome support!

-Dave

nightah:
A user whose password reset was rejected saw only `Unable to reset your password.`, which is deliberately vague so it discloses nothing to an attacker. The consequence was that a user who simply reused their existing password, or picked one the backend rejected, had no way to learn what to do differently, and the operator had to read the logs to tell them.

The response to the browser is unchanged. Instead the user is notified at the address they already proved control of in order to reach this step, so the notification can name the actual cause without disclosing anything to someone who does not hold the mailbox.

Only causes the user can act on are disclosed. `passwordResetFailureReason` maps the backend error to wording for a password which fails the password policy, fails the backend's own requirements, matches a password already set on the account, or was changed too recently to change again, and returns nothing for everything else, so a connection failure or a misconfiguration produces no notification at all.

The reuse and minimum age rejections are matched by new `ldapPasswordReuseErrors` and `ldapPasswordTooYoungErrors` slices rather than being folded into `ldapPasswordComplexityErrors`, because each is a distinct cause which warrants distinct wording, and because reusing the existing slice would also change the message shown in the browser. `Password is too young to change` already sat in the complexity slice, which would have told the user to choose a better password when the remedy is to wait.

The strings are taken from the servers themselves: `servers/slapd/overlays/ppolicy.c` in OpenLDAP, and `ldap/servers/slapd/pw.c` in 389 Directory Server and therefore FreeIPA, which spells every one of these differently and so matched nothing at all before. Active Directory reports length, complexity and history alike as `0000052D`, so a reuse there cannot be told apart and falls back to the backend wording.

The notifications reference gains the event, since it is the seventh notification Authelia can send and that page claims to list all of them.

Fixes #2926.


james-d-elliott (approved):
LGTM

PR #13095, opened 2026-09-10
feat(totp): make the suggested authenticator app configurable

### LINKED ISSUE #1981: Configurable OTP App Links
## Feature Request
Configurable OTP App Links

### Description

Instead of only providing links to google authenticator in the register one time password view, make the urls a configuration setting. Also, provide the ability to add a link to fdroid (or should that be a separate feature request?).

### Use Case
I'd like to suggest different apps to my users.

-----

I'd be happy to submit a PR for this but I have a few questions on implementation details before I do.
1. Regarding the API route to retrieve these settings, which of the following would be preferred?
    * New route. Something like `/api/configuration/authenticator_urls`
    * Add auth urls to the body of the existing response to `/api/configuration`
    * Add auth urls to the body of the existing response to `/api/secondfactor/totp/identity/finish`
2. When providing the optional link to fdroid, the current width of 128 is a few pixels too wide to have three badges on one line.
    * Should the width be decreased when showing three badges in order to fit all on one line?
      width126    

    * Should the badges be allowed to overflow to multiple lines?
      width128

    * Should the container be made larger to allow for three badges at 128?
3. I've never written Go before so I'm not sure what the standards are. For an optional fdroid url configuration setting, should its type be a `*string` and then check for `nil` or use a `string` and check for `""` (or is there perhaps another mechanism to represent an Option/Maybe)?

nightah:
The registration dialog hardcoded the Google Authenticator store listings in `web/src/constants/constants.ts`, so an operator who standardises on a different authenticator application had no way to point users at it and the dialog kept advertising an application they did not want used.

This adds `totp.apps.apple_store` and `totp.apps.google_play`, each with a `url` and a `disable` option. Every `url` is optional and defaults to the Google Authenticator listing for that store, so a deployment which configures nothing behaves exactly as it did before, and each configured value must be an absolute `https` URL. The values reach the frontend through the embedded `data-totpapp*` attributes rather than an API call, as they are static for the lifetime of the process, which is the same route already used by `resetpasswordcustomurl`.

A store which sets `disable` sends an empty link to the frontend and its badge is not rendered, which also means a disabled store's `url` is neither defaulted nor validated. Disabling every store removes the prompt above the badges as well, so an operator who does not want to suggest an application at all can turn the whole section off without the prompt being left behind pointing at nothing.

The prompt above the badges was `Need Google Authenticator?`, which would be wrong for any other configured application, so it is now `Need an authenticator app?`. The store badges themselves are unchanged: they identify the store rather than the application, so they remain correct whichever listing the link points at.

This does not add an F-Droid badge. That needs a third store badge shipped as a trademarked asset with its own `REUSE.toml` annotation, which is a licensing decision rather than a code one, and the issue itself raises it as possibly separate.

Closes #1981


james-d-elliott (changes requested):
LGTM just the schema properly should be held off on.

james-d-elliott (dismissed):

autheliabot (dismissed):

james-d-elliott (approved):

autheliabot (approved):

PR #13096, opened 2026-09-10
feat: add two-factor recovery codes

### LINKED ISSUE #1319: Add recovery code
Hi all, 

I think generate recovery code in case of second factor lost can be great...

Like Google, Microsoft and more.

nightah:
A user who loses their second factor device has no way back into their account without an administrator intervening. Recovery codes give them a fallback they can set up in advance and store somewhere safe.

Generating a batch issues ten single-use codes and revokes any the user already held, and requires an elevated session. Only the HMAC of the normalised code is persisted, under a key separate from the `otc` and `otp` keys so it can be rotated on its own with `authelia storage encryption rotate hmac rc`. Redeeming a code raises the session to two-factor but deliberately does not elevate it, so re-enrolling a primary factor still goes through identity verification. Both generating and redeeming notify the user by email.

Codes live in a new `recovery_codes` table at schema version 30. Operators get an `authelia storage user recovery-codes` command group, Prometheus counters and gauges covering adoption and depletion, and a `second_factor_recovery_code` rate limit rather than the endpoint borrowing the TOTP bucket.

Also sorts the `en` settings locale keys, which had drifted out of order.

Closes #1319.


james-d-elliott:
Small initial note, we should probably use 32 chars for the codes and group them by 4. This is the standard most password managers use for example. Since they don't need to be remembered and ad-hoc are used this should both be easy to type and impossibly secure.

PR #13110, opened 2026-09-11
fix(oidc): allow conformant alg none jarm

james-d-elliott:
This fixes an issue where a conformant JWT Secured Authorization Request i.e. OpenID Connect 1.0 Request Object was rejected. JARM's that omitted the typ header but included alg none were incorrectly rejected as the typ header can be absent when alg is explicitly none, this has minimal impact but still needs to be fixed.


nightah (approved):
LGTM.

autheliabot (approved):

PR #13128, opened 2026-09-12
feat(storage): index the oauth2 cascade and add a clean command

### LINKED ISSUE #13038: Provide additional cli tools to cleanup stale oauth2 records
### Description

The `oauth2_consent_session` table, despite only having 5 active rows (320k total), was taking up 208 MB of space. `oauth2_access_token_session` 147 active rows (40k total) at 71 MB. 

It should be possible to cleanup these stale records without having to manually execute SQL queries.

### Use Case

Reclaim storage space used by stale/unneeded database records.

### Details

Allow a command such as `authelia storage clean` (possibly with optional sub-commands to clean only specific tables/record types).

Additionally, concurrently add indexes on any table that references `oauth2_consent_session.challenge_id`

Currently, none of the tables that reference `oauth2_consent_session.challenge_id` have indexes on challenge_id, which means that when deleting from `oauth2_consent_session`, each delete causes a full table scan in order to complete the (on-delete) cascade.

### Documentation

_No response_

### Generative AI

No

### Pre-Submission Checklist

nightah:
Every table which references `oauth2_consent_session (challenge_id)` does so `ON DELETE CASCADE`, and each of them indexes `request_id`, `client_id` and `(client_id, subject)` but never `challenge_id`. The only index on that column anywhere is the unique one on the parent, so deleting a single consent session scans `oauth2_access_token_session`, `oauth2_authorization_code_session`, `oauth2_device_code_session`, `oauth2_openid_connect_session`, `oauth2_pkce_request_session` and `oauth2_refresh_token_session` end to end. That is a plausible explanation for the 320k rows and 208MB reported behind five live consent sessions, and it gets worse the longer an installation runs.

Schema version 30 adds the missing index to each of those tables. It is a no-op for MySQL: InnoDB requires an index on a foreign key column and creates one alongside the constraint, so the indexes are already there and an explicit one would only duplicate them. The MySQL migration is empty in the same way `V0009`, `V0025` and `V0026` already are.

It also adds `authelia storage clean oauth2`, which removes consent sessions that have both expired and been responded to, letting the cascade take the sessions which reference them. A session which has never been responded to is never removed as a flow may still be in progress against it. `--before` restricts this to rows older than a duration in the common syntax, and `--dry-run` reports the count without deleting anything.

Fixes #13038.


internal/storage/const.go
```diff
@@ -39,6 +39,19 @@ const (
...
+const (
+	queryFmtCountStaleOAuth2ConsentSessions = `
+		SELECT COUNT(*)
+		FROM %s
+		WHERE expires_at IS NOT NULL AND expires_at < ? AND responded_at IS NOT NULL AND responded_at < ?;`
+
+	queryFmtDeleteStaleOAuth2ConsentSessions = `
+		DELETE FROM %s
+		WHERE expires_at IS NOT NULL AND expires_at < ? AND responded_at IS NOT NULL AND responded_at < ?;`
+)
```
james-d-elliott: This is in the wrong file, should be in the _queries.go file.

PR #13132, opened 2026-09-12
docs: publish webhook payload json schemas

james-d-elliott:
Publishes the v1 JSON Schema documents for the webhook payloads: one per event type, one for the CloudEvents envelope, and one for the batched array. Receivers dereference these from the dataschema attribute of every event, and the OpenAPI webhooks section references them, so they must resolve before the subsystem ships.

The documents are generated from the Go payload types. This branch carries the output alone, so regenerate them from the branch which owns the generator rather than editing them here.


autheliabot (approved):

nightah (approved):
LGTM.

PR #13147, opened 2026-09-13
refactor(cmd): add a major version bump to the json schema generator

nightah:
The `next` meta version was hardcoded to a minor bump, so `--versions latest,next` could only produce `v4.40` from `v4.39`. There was no way to generate the schema for a major release.

`next` becomes `minor` and `major` is added. `--versions latest,minor` writes `latest` and `v4.40`, `--versions latest,major` writes `latest` and `v5.0`, and the `$id` of `latest` points at whichever the bump produces. `major`, `minor` and `current` are mutually exclusive.

An explicit version wrote the wrong `$id`. `--versions 4.38.0` created `v4.38` but named the `web/package.json` version in the identifier, because the directory came from the parsed version and the identifier from the current one.

`release` and `contributors` move under `misc`. Neither belongs to a plain `authelia-gen` run and both were skipped at the root, which `misc` already is. The contributors workflow runs `misc contributors` now.

`--latest` and `--next` are dropped. They landed alongside `--versions` and nothing ever read either one.

Command identity comes from `Name()` rather than `Use`. `Use` carries the argument hints, so `resolveCmdName` returned `misc.release {major|minor|patch}` as the name `--exclude` matches, and `rootCmdGetArgs` put that same string back on the argument list it re-executes with.

The certificate pages in the CLI reference pick up a `--bundles` rename from `privkey-chain` to `priv-chain` that was never regenerated.


autheliabot (approved):

james-d-elliott (approved):
LGTM

PR #13151, opened 2026-09-14
feat(middlewares): ipv6 rate limit mask

james-d-elliott:
Rate limits now group IPv6 addresses by a configurable prefix length using the ipv6_mask option on each endpoint, defaulting to 64 and accepting values from 48 to 128.


james-d-elliott:
@coderabbitai full review

PR #13155, opened 2026-09-14
fix(handlers): use correct status code for consent redirects

james-d-elliott:
This forces a 303 See Other rather than a 302 Found status when redirecting users for authorization consent which matches the requirements of FAPI 2.0.


nightah (approved):
LGTM.

autheliabot (approved):

PR #13158, opened 2026-09-14
feat: show an access denied page for forbidden requests

### LINKED ISSUE #2319: Create a deny page

nightah:
Denied users now see a portal page instead of the proxy's 403.

`ForwardAuth` and `ExtAuthz` redirect browsers with a session cookie to `/error?ec=forbidden`. Header credentials, XHR and non-HTML requests keep the 403. `AuthRequest` keeps the 403 and adds `Location`. `Legacy` is unchanged.

The page offers Switch User and Logout. It shows only the host of `rd`.

The NGINX suites follow the redirect with `error_page 403 =302 $redirection_url`. `AccessDeniedScenario` covers Traefik, Caddy, Envoy, HAProxy and Standalone.

Fixes #2319 and Closes #7251.


james-d-elliott (changes requested):
LGTM mostly just one nit.

Users who currently have their own handling of these status codes and are happy with them are unable to restore this functionality unless I'm misunderstanding something.

So if that's the case should we not add a opt-in or opt-out functionality?

PR #13167, opened 2026-09-15
fix(middlewares): include query in envoy authz path

james-d-elliott:
The envoy ext_authz path now includes the query string so the authorization object matches the other implementations.


autheliabot (approved):

nightah (approved):
LGTM.

PR #13192, opened 2026-09-17
fix(locales): update polish translation for password and passkey

kylo32:
There are some minor but needed changes in the Polish language


james-d-elliott:
Please use https://translate.authelia.com, these translations will automatically be replaced by Crowdin at the next sync if not.

kylo32:
> Please use https://translate.authelia.com, these translations will automatically be replaced by Crowdin at the next sync if not.

Sorry, I was not aware of this dedicated page. I added my suggestions on Crowdin.

james-d-elliott:
Not a problem at all.

